PR #23, opened 2026-04-16
Fix: prevent nested tokio runtime panic on Windows daemon

### LINKED ISSUE #22: tokensave daemon panics on Windows: "Cannot start a runtime from within a runtime"
### tokensave version

4.0.2

### Operating system

Windows

### Installation method

Scoop

### What happened?

Running tokensave daemon on Windows panics immediately with a tokio nested runtime error.

### Steps to reproduce

1. Build tokensave on Windows
2. Run tokensave daemon
3. Process panics

### Error output

```shell
thread 'main' panicked at tokio-1.51.1/src/runtime/scheduler/multi_thread/mod.rs:91:9:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

### Component

Daemon / file watcher

### Additional context

Root cause: daemon-kit on Windows does not fork — it runs the daemon closure inline in the same thread. The
closure creates a new tokio::runtime::Runtime via Runtime::new().block_on(), but #[tokio::main] already has an active runtime, causing the panic. On Unix/macOS the fork produces a clean child
process without a runtime, so the same code works fine.

Fix: use tokio::task::block_in_place + Handle::current().block_on() on Windows, keep Runtime::new() on Unix.

LucioPg:
## Summary

- tokensave daemon panicked on Windows because daemon-kit runs the closure inline (no fork), so Runtime::new().block_on() nested inside the existing #[tokio::main] runtime
- Use tokio::task::block_in_place + Handle::current().block_on() on Windows; keep Runtime::new() on Unix where the forked child genuinely has no runtime

-

## Test plan

- [x] `cargo build --no-default-features --features lite` compiles on Windows                                                                                                                                                                                                                                 
- [x] `tokensave daemon` starts without panic on Windows (previously panicked with nested runtime error)
- [ ] `cargo test` — blocked by pre-existing linker issue on this machine (stale `tokensave-large-treesitters` rlib with unresolved `tree_sitter_dockerfile`), unrelated to this change
- [ ] Verify `tokensave daemon` still works on Unix/macOS (CI or maintainer)

## Checklist

Closes #22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **Windows daemon nested runtime panic** — `tokensave daemon` panicked on Windows because `daemon-kit` runs the closure inline (no fork), creating a nested tokio runtime. Now uses `block_in_place` + `Handle::current()` on Windows while keeping `Runtime::new()` on Unix where the forked child genuinely has no runtime.
+
 ## [4.0.2] - 2026-04-14
 
 ### Added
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Thank you @LucioPg 
Releasing 4.0.3 with the fix.

PR #26, opened 2026-04-16
Fix/symbolic ref default branch

### LINKED ISSUE #25: `"HEAD"` as default branch in `branch-meta.json` leading to an incorrect MCP server lookup
### tokensave version

4.0.3

### Operating system

Windows

### Installation method

Scoop

### What happened?

When `refs/remotes/origin/HEAD` exists as a symbolic reference (e.g. pointing to `refs/remotes/origin/master`), `detect_default_branch()` writes the literal string `"HEAD"` as the `default_branch` value in `.tokensave/branch-meta.json`.
`reference.name()` returns the reference's own name (`refs/remotes/origin/HEAD`), not the symbolic ref's target. Stripping `refs/remotes/origin/` yields `"HEAD"` instead of the actual branch name.

### Steps to reproduce

1. Clone a repo where `refs/remotes/origin/HEAD` is a symbolic ref
2. Delete `.tokensave/` (or use a fresh clone)
3. Run `tokensave init` or any command that bootstraps branch metadata
4. Check `.tokensave/branch-meta.json` — `default_branch` is `"HEAD"`

### Error output

```shell

```

### Component

MCP server

### Additional context

I'm not sure if there's something obvious I might be missing in my understanding of the codebase, but I ran into this issue where the MCP server couldn't resolve the correct database because it was looking for a branch named `"HEAD"` instead of the actual branch name. This caused `branch-meta.json` to be written with an invalid `default_branch`, and all subsequent tool calls failed to find the right DB. Manually editing the file to the correct branch name fixed it, which confirmed the root cause. Note: this did not reproduce in CLI mode.

LucioPg:
## Summary

- Fix `detect_default_branch()` resolving `refs/remotes/origin/HEAD` symbolic ref to the literal string `"HEAD"` instead of the actual branch name (e.g. `"master"`).

## Motivation

When `refs/remotes/origin/HEAD` exists as a symbolic reference, `reference.name()` returns the ref's own name. Stripping the `refs/remotes/origin/` prefix yields `"HEAD"`, which gets written as `default_branch` in `.tokensave/branch-meta.json`. This corrupts all branch operations (add, remove, gc, MCP tools) that rely on `meta.default_branch`.

Closes #(issue_number)

## Changes

- `src/branch.rs`: use `reference.follow()` to resolve the symbolic ref target before extracting the branch name, instead of using `reference.name()` directly.

## Test plan

- [x] `cargo check --no-default-features --features lite` compiles
- [ ] Verify `branch-meta.json` writes correct default branch name (e.g. `"master"`) after `tokensave sync` in a repo with `refs/remotes/origin/HEAD`
- [ ] Verify fallback to heuristic (`main`/`master`) still works when no `origin/HEAD` ref exists

## Checklist

Closes #25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Default branch detection wrote `"HEAD"` instead of actual branch name** — `detect_default_branch()` used `reference.name()` on the `refs/remotes/origin/HEAD` symbolic ref, which returns the ref's own name. Now resolves through `reference.follow()` to get the target (e.g. `refs/remotes/origin/master`), then strips the prefix correctly.
+
 ## [4.0.3] - 2026-04-16
 
 ### Fixed
--- a/src/branch.rs
+++ b/src/branch.rs
@@ -23,10 +23,12 @@ pub fn current_branch(project_root: &Path) -> Option<String> {
 pub fn detect_default_branch(project_root: &Path) -> Option<String> {
     let repo = gix::open(project_root).ok()?;
 
-    // Try symbolic-ref first
+    // Try symbolic-ref first (refs/remotes/origin/HEAD -> refs/remotes/origin/<branch>)
     if let Ok(reference) = repo.find_reference("refs/remotes/origin/HEAD") {
-        if let Some(name) = reference.name().as_bstr().to_string().strip_prefix("refs/remotes/origin/") {
-            return Some(name.to_string());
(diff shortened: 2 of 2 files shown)


aovestdipaperino:
Merged manually into master — thank you @LucioPg!

CI on all open PRs was broken due to pre-existing formatting and clippy issues on master that caused every branch to fail. Since rebasing wouldn't have fixed it, I merged all four open PRs locally, resolved the CHANGELOG conflicts, ran `cargo fmt` across the codebase, and fixed all clippy errors in a follow-up commit.

Your fix for `detect_default_branch` resolving the symbolic ref via `.follow()` is now live. 🎯

PR #30, opened 2026-04-17
fix: treat no-op upgrade checks as success

lesbass:
## Summary

- treat `tokensave upgrade` returning "already up to date" as a successful no-op instead of a config error
- treat `tokensave channel <current-channel>` as a successful no-op for the same reason
- add regression tests for the no-op upgrade/channel paths

## Motivation

Running `tokensave upgrade` while already on the latest release currently prints a success-looking message:

```text
Current version: v4.0.3 (stable channel)
Checking for updates...
✔ Already up to date (v4.0.3).
```

but then the command exits through `TokenSaveError::Config`, so the CLI also prints:

```text
Error: config error: already at latest version v4.0.3
```

That is misleading UX for a normal no-op state. There is no underlying operational failure in that path; the code was intentionally modeling "already current" as an error, and `main` propagated it with `?`.

The same pattern existed in `tokensave channel <current-channel>`, which reported being already on the selected channel and then still returned a config error.

## Changes

- introduce a small `classify_upgrade()` helper in `src/upgrade.rs` to separate "upgrade available" from "already current"
- make `run_upgrade()` return `Ok(current.to_string())` when no newer version is available
- make `switch_channel()` return `Ok(current.to_string())` when the requested channel matches the current one
- add unit tests covering equal-version no-op upgrades, newer-version detection, and same-channel no-op switching

## Test plan

- [x] `cargo test --lib upgrade::tests`
- [ ] `cargo clippy` has no new warnings

## Checklist

- [ ] `CHANGELOG.md` updated (under `[Unreleased]` if no version bump)

### DIFF
--- a/src/upgrade.rs
+++ b/src/upgrade.rs
@@ -229,6 +229,20 @@ fn replace_binary(new_exe: &Path) -> Result<()> {
     result
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum UpgradeStatus<'a> {
+    AlreadyCurrent,
+    UpgradeAvailable(&'a str),
+}
+
+fn classify_upgrade<'a>(current: &str, latest: &'a str) -> UpgradeStatus<'a> {
+    if cloud::is_newer_version(current, latest) {
+        UpgradeStatus::UpgradeAvailable(latest)
+    } else {
+        UpgradeStatus::AlreadyCurrent
+    }
+}
+
 /// Downloads, extracts, and installs the binary for `version`/`is_beta`.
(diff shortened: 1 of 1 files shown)


aovestdipaperino:
Merged manually into master — thank you @lesbass!

CI on all open PRs was broken due to pre-existing formatting and clippy issues on master that caused every branch to fail. Since rebasing wouldn't have fixed it, I merged all four open PRs locally, resolved the CHANGELOG conflicts, ran `cargo fmt` across the codebase, and fixed all clippy errors in a follow-up commit.

Your fix for treating no-op upgrade/channel checks as success instead of errors is now live.

PR #33, opened 2026-04-17
Fix upgrade flow messaging for Homebrew installs

lesbass:
## Context
- `tokensave upgrade` was run from a Homebrew-installed binary on macOS
- the CLI detected an available update and attempted a self-replace of the running executable
- that path fails for package-manager installs and surfaced as `binary replacement failed: No such file or directory (os error 2)`
- the update messaging was also misleading because it always suggested `tokensave upgrade`, even when the install method required a package-manager command instead

## Summary
- make `upgrade_command()` return an install-method specific update instruction instead of always suggesting `tokensave upgrade`
- make `tokensave upgrade` exit early for Homebrew and Scoop installs with the correct package-manager command instead of attempting self-replacement
- use the install-method specific update instruction in CLI/MCP update notices and adjust the cloud test

## Testing
- cargo test --test cloud_test upgrade_command_matches_install_method
- cargo test --test cloud_test detect_install_method_no_panic

### DIFF
--- a/src/cloud.rs
+++ b/src/cloud.rs
@@ -232,9 +232,10 @@ pub fn detect_install_method() -> InstallMethod {
 }
 
 /// Returns the upgrade command string.
-///
-/// Always suggests `tokensave upgrade` which handles all install methods
-/// and channels automatically.
-pub fn upgrade_command(_method: &InstallMethod) -> &'static str {
-    "tokensave upgrade"
+pub fn upgrade_command(method: &InstallMethod) -> &'static str {
+    match method {
+        InstallMethod::Cargo | InstallMethod::Unknown => "tokensave upgrade",
+        InstallMethod::Brew => "brew upgrade tokensave",
+        InstallMethod::Scoop => "scoop update tokensave",
+    }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -562,9 +562,11 @@ async fn run(cli: Cli) -> tokensave::errors::Result<()> {
                 if tokensave::cloud::is_newer_version(current_version, &latest)
                     && now - config.last_version_warning_at >= 900
                 {
+                    let upgrade_cmd =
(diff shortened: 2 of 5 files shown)


aovestdipaperino:
ah, interesting, I am planning to move in the opposite direction.
I want to be `tokensave upgrade` to be the only safe way to support upgrade since I've had problems with scoop.
The idea is to use brew/scoop for the initial install and a tokensave upgrade that replaces each channel in a channel compatible way. So user can do `brew upgrade tokensave` even after a previous `tokensave upgrade`.

lesbass:
sounds good to me :) I just need to be able to upgrade it even if it was initially installed via brew, while now I cannot do it. Feel free to decline this PR. Thanks @aovestdipaperino!

aovestdipaperino:
Superseded by d43a8fa (pushed to master).

That commit takes a different approach: instead of delegating to `brew upgrade` / `scoop update`, `tokensave upgrade` now handles all install methods directly — it bypasses `self_replace` for Homebrew (atomically replaces the binary at the canonical Cellar path, renames the version directory, updates the symlink and `INSTALL_RECEIPT.json`), and updates Scoop's version directory + junction after `self_replace`. This way `brew` / `scoop status` report the correct version without needing the user to run the package manager separately.

aovestdipaperino:
Closing — superseded by d43a8fa on master.

aovestdipaperino:
Regarding the suggestion for shared indexes across worktrees with differential sync: we've determined that duplicating the DB per worktree (which `tokensave branch add` already does) is the right trade-off. A differential approach would require SQLite-level change tracking (shadow tables, triggers, or WAL replay) to apply/revert file-level deltas against a shared database, and the concurrency story for multiple worktrees hitting the same DB simultaneously adds significant complexity. The cost of a full DB copy (~5-50 MB for typical projects) is negligible compared to that maintenance burden.

PR #34, opened 2026-04-17
feat: follow symlinked source directories during indexing

lesbass:
## Summary

- follow symlinked directories during file discovery in both the plain `walkdir` path and the `.gitignore`-aware `ignore::WalkBuilder` path
- update the fallback source probe to use the same symlink-following behavior
- add Unix integration tests covering indexing through a symlinked source directory with and without `git_ignore` enabled

## Problem

Right now `tokensave` skips directory symlinks while scanning files because both walkers are configured with `follow_links(false)`. That means projects that expose source code through symlinked paths are only partially indexed or not indexed at all, even when the symlinked tree is part of the intended project layout.

## Implementation

This changes the file discovery paths in `TokenSave::scan_files()` to follow symlinks consistently:

- `scan_files_walkdir()` now uses `WalkDir::follow_links(true)`
- `scan_files_with_gitignore()` now uses `ignore::WalkBuilder::follow_links(true)`
- the fallback `has_source` probe also follows symlinks, so the fallback decision matches the actual scan behavior

## Testing

- `cargo test test_index_follows_symlinked_directories --test integration_test`
- `cargo test test_gitignore_scan_follows_symlinked_directories --test integration_test`
- `cargo test test_multiple_files_cross_reference --test integration_test`

### DIFF
--- a/src/tokensave.rs
+++ b/src/tokensave.rs
@@ -880,6 +880,7 @@ impl TokenSave {
                 // causing the ignore-aware walker to skip everything. Fall
                 // back to plain walkdir if source files clearly exist.
                 let has_source = WalkDir::new(&self.project_root)
+                    .follow_links(true)
                     .max_depth(2)
                     .into_iter()
                     .filter_map(|e| e.ok())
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Merged, thanks @lesbass! The `follow_links(true)` change is clean and the tests cover both walker paths well.

The CI failures on the PR were not your code — Clippy/Format failed due to a toolchain setup issue on the older base, and Windows failed from a pre-existing duplicate `tree-sitter` symbol link error (now resolved on master via `tokensave-large-treesitters` 0.3.0). All 1366 tests pass after the merge.

PR #47, opened 2026-04-29
feat: add markdown extraction support

pierreaubert:
## Summary

Add tree-sitter based markdown extraction for the code graph. Enables tokensave to parse markdown files and extract:

- **Headers** as `Module` nodes with hierarchical relationships (`Contains` edges)  
- **Code links** (links to source files) as `Uses` edges for cross-reference tracking

## Files changed

| File | Change |
|------|--------|
| `src/extraction/markdown_extractor.rs` | Main extractor (~350 lines) |
| `src/extraction/mod.rs` | Module registration |
| `tests/markdown_extraction_test.rs` | 14 test cases |
| `Cargo.toml` | Added `lang-markdown` feature + `tree-sitter-markdown-fork` dep |
| `Cargo.lock` | Updated |

## Implementation

- Uses `tree-sitter-markdown-fork` for parsing
- Extracts ATX headings (`# heading`) with parent-child hierarchy
- Extracts links to code files (`.rs`, `.py`, `.ts`, etc.) as `Uses` edges
- Skips external HTTP(S) links and non-code file links

## Tests

All 14 tests pass covering header extraction, link extraction, hierarchy, and edge cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Markdown extraction** — tree-sitter based markdown parser that extracts headers as `Module` nodes with hierarchical `Contains` edges, and code links as `Uses` edges for cross-reference tracking (PR #47)
+
 ### Fixed
 - `java_extraction` avoid panic when parsing empty Java docstrings (fixes #44)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4041,6 +4041,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "tree-sitter",
+ "tree-sitter-markdown-fork",
  "ureq",
  "walkdir",
  "windows-sys 0.61.2",
@@ -4541,6 +4542,16 @@ checksum = "8daaf5f4235188a58603c39760d5fa5d4b920d36a299c934adddae757f32a10c"
 dependencies = [
  "cc",
  "tree-sitter-language",
+]
+
+[[package]]
+name = "tree-sitter-markdown-fork"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3d5ad4a0fbf1d16abe5391387dab9e25977ba0c86661627a2e8680386c8a1c27"
+dependencies = [
+ "cc",
+ "tree-sitter",
 ]
 
 [[package]]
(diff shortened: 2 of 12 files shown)


aovestdipaperino:
Thanks for the PR — the approach is solid and the test coverage is good. A few things to address before this can merge.

## Hard blockers

**Dangling `Uses` edges** (`src/extraction/markdown_extractor.rs`, `visit_link`):

```rust
let target_id = generate_node_id(target_path, &NodeKind::Use, &link_text, 0);
```

`NodeKind::Use` will never match the `NodeKind::File` node that the linked file produces when it gets indexed. Every markdown link in the codebase creates an edge pointing at a phantom node that doesn't exist in the graph, silently breaking `tokensave_callers`, `tokensave_impact`, and any other traversal that follows `Uses` edges. Should be:

```rust
let target_id = generate_node_id(target_path, &NodeKind::File, target_path, 0);
```

**External grammar dependency.** Rather than pulling in `tree-sitter-markdown-fork`, we should add native markdown support directly to `tokensave-large-treesitters`, which we control. That keeps all grammar dependencies in one place, avoids a fork with no maintenance guarantee, and follows the same pattern as every other language we support. Happy to help coordinate that if needed.

## Should fix before merge

**`is_code_extension` is too broad.** It currently includes `yaml`, `yml`, `toml`, `json`, `xml`, `html`, `css`, `scss`, `sql`, `db`, `ipynb`, and others. A `[config.yaml](infra/config.yaml)` link in a README would produce a `Uses` edge, flooding the graph with low-signal cross-references. Suggest tightening to actual programming-language source extensions only.

**Self-loop on parse failure.** When tree-sitter returns `Err`, the code pushes a `Contains` edge from the file node back to itself. A self-loop is meaningless noise — on failure, do nothing (or push to `errors`).

## Minor (can be a follow-up)

- Level detection parses the tree-sitter node kind string (`"atx_h2_marker"` → extract `"2"`) instead of counting `#` characters in the marker text. Brittle if the grammar naming changes. Counting `#` chars is simpler and more robust.
- `setext_heading` (`H1\n===`, `H2\n---`) is not handled. Not a blocker for most repos but worth a TODO comment.

pierreaubert:
Note: test_ignore_crate_nested_gitignore_direct is failing but looks unrelated.

aovestdipaperino:
Thank you @pierreaubert 
Obviously ignore already failing tests, I will fix the separately

aovestdipaperino:
Thanks @pierreaubert for the contribution! The markdown extraction support is a great addition to the graph. We made a few adjustments on top — swapped the external `tree-sitter-markdown-fork` dependency for the grammar now compiled natively inside `tokensave-large-treesitters`, and fixed a Windows path separator issue in the test. Really appreciate the work!

Issue #67, opened 2026-05-14
Homebrew installs can be broken by tokensave upgrade


lesbass:
When tokensave is installed via Homebrew, running `tokensave upgrade` can leave Homebrew's Cellar state inconsistent.

## Observed behavior

1. Install tokensave with Homebrew.
2. Run `tokensave upgrade`.
3. Later run `brew upgrade`.
4. Homebrew may fail with an error like:

```text
Error: /opt/homebrew/Cellar/tokensave/4.3.13 is not a directory
```

## Current behavior in code

The Homebrew-specific upgrade path appears to replace the binary and then rename the Cellar version directory / update symlinks / patch `INSTALL_RECEIPT.json` so that `tokensave` reports the newer version.

That can make the local Homebrew installation disagree with Homebrew's own recorded keg path/version state. A later `brew upgrade` may then look for the original version directory and fail because it was renamed by `tokensave upgrade`.

## Design question

What should `tokensave upgrade` do for Homebrew-managed installs?

Possible options:

1. Refuse self-upgrade when installed via Homebrew and tell users to run:

   ```sh
   brew update && brew upgrade tokensave
   ```

2. Keep self-upgrade but only replace the binary in place, accepting that the Cellar directory name and receipt remain at the Homebrew-installed version.

3. Try to fully simulate a Homebrew upgrade by creating/moving a coherent keg and relinking, though this seems fragile because it depends on Homebrew internals rather than a stable Homebrew API.

4. Change update warnings so Homebrew installs suggest `brew upgrade tokensave` instead of `tokensave upgrade`.

My concern is that option 3 may keep breaking as Homebrew changes metadata/layout behavior. Option 1 seems the safest operationally, but it changes the current “one upgrade command” behavior.

aovestdipaperino:
Fixed in v4.3.17. `tokensave upgrade` now detects Homebrew installs and delegates to `brew update && brew upgrade tokensave`, keeping Homebrew's Cellar state consistent. The single `tokensave upgrade` command still works for all install methods — it just routes through Homebrew when appropriate.

Thanks for the thorough analysis and the four options breakdown — it made the design decision easy. PR #68 from @lesbass implemented the core delegation and is included in this release.

PR #68, opened 2026-05-14
Delegate Homebrew upgrades to brew

lesbass:
## Summary

- Delegate `tokensave upgrade` to `brew upgrade tokensave` when the current install is detected as Homebrew-managed.
- Keep the existing self-upgrade flow unchanged for non-Homebrew installs.
- Add a small unit test for the delegated Homebrew command.

## Context

This follows the discussion in #67. When a Homebrew-installed `tokensave` self-upgrades by mutating the Cellar directly, Homebrew can later fail because its recorded keg path/version no longer matches what is on disk, for example:

```text
Error: /opt/homebrew/Cellar/tokensave/4.3.13 is not a directory
```

After discussing the issue with the maintainer, the agreed direction is to preserve the single user-facing command (`tokensave upgrade`) while letting Homebrew perform the actual package-manager upgrade for Homebrew-managed installs.

## Solution

`run_upgrade()` now checks `InstallMethod::Brew` early and runs:

```sh
brew upgrade tokensave
```

This avoids downloading/replacing the binary inside `tokensave` for Homebrew installs, so Homebrew remains responsible for Cellar layout, linked keg state, receipts, and metadata.

## Testing

```sh
cargo fmt --check
cargo test upgrade::
cargo test brew_upgrade_command_delegates_to_homebrew
```

### DIFF
--- a/src/upgrade.rs
+++ b/src/upgrade.rs
@@ -556,6 +556,31 @@ fn restart_daemon() {
     }
 }
 
+fn brew_upgrade_command() -> (&'static str, [&'static str; 2]) {
+    ("brew", ["upgrade", "tokensave"])
+}
+
+fn run_brew_upgrade(current: &str) -> Result<String> {
+    let (program, args) = brew_upgrade_command();
+    eprintln!(
+        "Delegating upgrade to Homebrew: {program} {}",
+        args.join(" ")
+    );
+
+    let status = std::process::Command::new(program)
+        .args(args)
+        .status()
+        .map_err(io_err("failed to run Homebrew upgrade"))?;
+
+    if status.success() {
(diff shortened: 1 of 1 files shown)


aovestdipaperino (commented):
Looks good — clean fix for #67. Two minor things:

1. **Verify print ordering**: Confirm the early return fires *before* the `"Checking for updates..."` eprintln. The diff context suggests it does, but worth a quick check that Homebrew users don't see that misleading line.

2. **Consider `brew update` first**: `brew upgrade tokensave` can no-op if the formula cache is stale. Running `brew update` (or at least mentioning it in the error message on failure) would make this more robust.

Otherwise LGTM.

PR #78, opened 2026-05-19
fix: make edit prefixes UTF-8 safe

grp06:
## Summary

- Add a private UTF-8-safe byte-budgeted prefix helper for edit tool previews and anchor matching.
- Use it in multi-replace failure messages and long string-anchor searches.
- Add MCP regression coverage for non-ASCII failed edit inputs and update the changelog.

## Motivation

Failed edit requests should return structured failure responses and leave files unchanged for valid UTF-8 input, instead of panicking when an internal byte limit lands inside a multi-byte character.

## Changes

- `src/tokensave.rs`: centralizes safe prefix slicing and applies it to the two edit-tool paths.
- `tests/mcp_handler_test.rs`: covers multi-replace preview and insert-at anchor regressions through the MCP handler boundary.
- `CHANGELOG.md`: notes the fix under `[Unreleased]`.

## Test plan

- [x] `cargo test --test mcp_handler_test test_multi_str_replace_unicode_preview_does_not_panic --no-default-features --features lite`
- [x] `cargo test --test mcp_handler_test test_insert_at_unicode_anchor_prefix_does_not_panic --no-default-features --features lite`
- [x] `cargo test --test mcp_handler_test --no-default-features --features lite`
- [x] `cargo test --workspace`
- [x] `cargo clippy --workspace --all-targets`

## Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - **Kiro agent integration.** `tokensave install --agent kiro` now installs the MCP server, global steering, managed Kiro agent config, default CLI agent selection, and Kiro hook mappings for prompt context, delegated tool context, and post-write re-indexing. Doctor and uninstall support are included with coverage for workspace overrides and idempotent cleanup.
 
+### Fixed
+- **Edit tool UTF-8 failure handling.** `tokensave_multi_str_replace` and `tokensave_insert_at` no longer panic when failure previews or long anchors contain multi-byte UTF-8 characters.
+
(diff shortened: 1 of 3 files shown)


aovestdipaperino:
Thanks for the fix, @grp06! Clean, minimal, and well-tested — exactly the kind of PR that's a joy to review. Merged as f6efd6d.

We're going to follow up by promoting your `utf8_prefix_at_or_before` helper into a shared utility — there's a duplicate of the same walk-back logic in `src/context/builder.rs` that should reuse it, and a couple of BASIC extractors with related byte-slicing assumptions worth hardening. Appreciate the contribution!

PR #77, opened 2026-05-18
feat: add Kiro agent integration

norunners:
## Summary
- add first-class Kiro install, uninstall, and doctor support
- install Kiro MCP config, global steering, managed agent settings, default CLI agent selection, and hook mappings
- add Kiro-specific hook commands for prompt context, delegated tool context, and post-write re-indexing
- document the Kiro integration and update user-facing agent docs
- add focused Kiro integration tests and update agent/hook coverage

## Notes
- Targets `master` as a stable feature.
- Includes an `[Unreleased]` `CHANGELOG.md` entry per `CONTRIBUTING.md`.

## Verification
- `cargo fmt --all -- --check`
- `cargo test`
- `cargo clippy`
- Kiro install/uninstall smoke tests with temporary `HOME`

PR #88, opened 2026-05-26
refactor(kiro): use resource-backed steering

norunners:
## Summary

- Follow-up to #77 now that first-class Kiro support has shipped.
- Switch the managed tokensave Kiro agent to leave `prompt` unset and reference global steering through `~/.kiro/steering/tokensave.md` as a resource.
- Keep the global Kiro MCP config focused on server registration, with tool approval policy owned by the managed custom agent.
- Tighten install, uninstall, and doctor behavior around owned steering, user-managed agents, workspace MCP overrides, hooks, and tool policy drift.
- Update Kiro docs, README/user guide references, changelog, and focused integration tests.

## Notes

- Targets `master` as a stable follow-up to PR #77.
- Includes an `[Unreleased]` `CHANGELOG.md` entry per `CONTRIBUTING.md`.

## Verification

- `cargo fmt --check`
- `cargo test`
- `cargo clippy --workspace --all-targets`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **Kiro steering is loaded as a resource.** The Kiro installer now writes `~/.kiro/steering/tokensave.md`, loads it from the managed agent's `resources` list with an absolute `file://` URI, leaves the custom-agent `prompt` unset so Kiro's default prompt is preserved, keeps MCP approval policy out of `mcp.json`, and installs permissive `tools: ["*"]` plus `allowedTools: ["@builtin", "@tokensave"]` defaults for the managed agent.
+
 ## [6.1.0] - 2026-05-25
 
 ### Added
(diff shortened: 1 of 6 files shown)


norunners:
@aovestdipaperino for review, looks like `multi_mcp_coordination_test` was already broken.

aovestdipaperino:
Thanks @norunners — merged via squash as `ffd1793`. The migration to resource-backed steering with explicit `allowedTools` is a nice cleanup, and the focused doctor checks make drift surface clearly without escalating it to a hard failure. The unrelated Linux CI flake on `two_mcps_on_same_project_coordinate_via_sync_lock` was overridden — it's a pre-existing watcher race that's already being removed in a separate change.

Two minor follow-ups noted during review, both non-blocking:

1. `CHANGELOG.md` could use a quick `[Unreleased]` entry covering the `AGENTS.md` → `tokensave.md` and `prompt` → `resources` migration, plus a one-line note that old `~/.kiro/steering/AGENTS.md` files from #77 are left in place (intentional — tokensave never owned them when they were shared — but worth mentioning so users know to clean up manually).
2. `file_resource_uri` / `percent_encode_file_uri_path` are duplicated between `src/agents/kiro.rs` and `tests/kiro_agent_test.rs`. `pub(crate)` on the originals would let the test import them and remove the copy-paste risk.

Appreciate the thorough test coverage and the contributing-guideline-following PR body. 🙏

PR #95, opened 2026-06-01
[Proposal] Use older Linux for better backward compatibility.

sstepashka:
## Summary

Since the binaries are built on newer version, their libc version is not compatible with older linux distros. Closing the gap by using slightly older version of Linux in order to provide better backward compatibility.

## Motivation

Failure example when running on ubuntu 22.04:
```
.../linux/arm64/tokensave: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by ...)
```

### DIFF
--- a/.github/workflows/release-beta.yml
+++ b/.github/workflows/release-beta.yml
@@ -39,12 +39,12 @@ jobs:
             bottle_tag: arm64_sonoma
             archive: tar.gz
           - name: x86_64-linux
-            runner: ubuntu-latest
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Thank you

sstepashka:
@aovestdipaperino Do you think we can trigger a release to pick up the backward compatible binaries? :) If it's not much work, otherwise, I can always wait for some changes to come in... 

Thank you.

PR #98, opened 2026-06-06
F* language support

hacklex:
## Summary

Added support for F* proof-oriented programming language.
This one fetches pre/post-conditions along with top level definition names, which is crucial for F*.

## Motivation

F* diverged from OCaml / F# far enough

## Changes

Added support to fst / fsti F* files

## Test plan

- [x] `cargo test` passes
- [x] Tested manually against my pet computer algebra project [CuteCAS](https://github.com/hacklex/CuteCAS) (over 50k LoC)

Fixture file added, extraction test added.

## Checklist

- [x] No Breaking Changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- F\* language support (`.fst` / `.fsti`).
+
 ## [6.1.3] - 2026-06-04
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


aovestdipaperino:
Merged — thank you, @hacklex! 🎉 Nicely scoped extractor: the pre/post-condition extraction is a thoughtful touch for F*, the fixture + 36 extraction tests gave me confidence, and it integrated cleanly with no conflicts on master. Appreciate the contribution.

hacklex:
You are welcome! I think this is a great repo, that's why I helped a bit :)

Parsing `(requires/ensures/decreases)` was the only touch from me to be honest, the rest Opus did on its own, given just the ancestor languages files as references. 

They say in F* roughly 9/10 of LoC are the correctness proofs, so without F* Effect Pre/Post conditions, fetching definitions in F* is all but useless (even the best available models fail spectacularly, even when having full info, so having them guess theorem statements by just their names is just asking too much).

As for the tests -- I just enumerated all relevant language capabilities I could remember and asked the AI to wrap them in a nice package. 

I haven't tried to estimate how much this saves, but since it works well against an already big repo in which I already made use of most of F* features I knew of, I think this extractor will work well. If people who work with stateful/effectful F* code get to play with this, they might cover the rest.

Issue #134, opened 2026-06-16
Flaky test: mcp_server_owns_watcher_and_refreshes_token_map_on_change fails under full-suite parallel load


aovestdipaperino:
## Summary
`mcp_server_owns_watcher_and_refreshes_token_map_on_change` (`tests/mcp_handler_test.rs:5056`) is flaky: it passes deterministically in isolation (`cargo test --test mcp_handler_test mcp_server_owns_watcher_and_refreshes_token_map_on_change` → 1 passed) but intermittently fails when the full test suite runs in parallel (as in CI). Observed during unrelated work (PR #133 / #128), which touches only `src/agents/` and cannot be the cause.

## Likely cause
A timing/race assumption in the file-watcher test: it appears to depend on the debounced watcher refreshing the token map within a fixed wall-clock window. Under full-suite parallel load the machine is contended, the debounce/refresh lands late, and the assertion fires before the expected state is observed.

## Why it matters
Intermittent CI failures on unrelated PRs erode trust in the signal and cause needless re-runs / manual "is this real?" triage.

## Suggested fix directions
- Replace fixed `sleep`/deadline assumptions with polling until the expected token-map state appears, bounded by a generous timeout (await-condition rather than await-duration).
- Or make the watcher refresh observable via a signal/channel the test can wait on deterministically.
- Or serialize this test (e.g. a dedicated single-threaded test or a shared lock) if it genuinely cannot tolerate parallel load.

## Acceptance
- The test passes reliably under `cargo test` (full suite, parallel) across repeated runs.
- No fixed-duration sleeps gating the core assertion.

aovestdipaperino:
The condition-polling fix landed in #213 (plus removal of the racy `find_stale_files` detection assert, which could also flake when the server's startup catch-up sync won the race — observed live on #214's CI). Leaving this open briefly to confirm stability across the next several CI runs; will close once it's held.

PR #144, opened 2026-06-18
feature: post checkout hook

ilteoood:
## Summary

Adds a new `post-checkout` git hook that automatically runs `tokensave init` after a fresh clone.

Follows the same pattern as the existing `post-commit` hook in `scripts/`.

-

## Motivation

Reduce manual steps

## Changes

introduced new hook
changed content of readme and setup

## Test plan

- [x] `cargo test` passes

## Checklist


aovestdipaperino:
Thanks for the contribution! Two things:

**CI:** the `Test Windows` ❌ is a flake, not your change — `mcp_handler_test` crashed with `STATUS_ACCESS_VIOLATION` *after all tests passed, on process exit*, and this PR touches zero Rust code. Same class of native flake as #134. I've re-run it.

**Design concern (the thing to resolve before merge):** the `post-checkout` hook auto-runs `tokensave init` on every fresh clone. That conflicts with a deliberate safety guard documented in the README (unchanged by this PR):

> `init` is a one-time opt-in per project, while `sync` only updates projects that were already initialized. **This prevents the global git post-commit hook from silently creating databases in repos you never intended to index.**

Installed as a global hook, this would silently create `.tokensave/` databases in *every* repo you clone — exactly what that guard was designed to prevent. This PR also removes the explicit "run `tokensave init`" step from `setup.sh`, which leans further into auto-init.

I'm not rejecting the idea — reducing manual steps is a real goal — but auto-init-on-clone is a product-direction decision that needs sign-off, and as written it reverses an intentional design. Could we discuss the scope here (e.g. opt-in only / project-local hook only / a prompt rather than silent background init) before merging? Holding merge on that, not on the flaky CI.

aovestdipaperino:
Update: discussed with the maintainer — the intent is to offer post-checkout as an **opt-in, exactly like the existing post-commit hook**, which resolves my earlier concern. I've pushed a commit to wire it up:

- The installer now offers the `post-checkout` hook alongside `post-commit`, gated by the **same** opt-in (`--git-hook` flag / interactive prompt). Previously the PR added `scripts/post-checkout` + docs but `offer_git_post_commit_hook` was never extended, so the hook was never actually installed.
- The generated hook runs `tokensave init` in the background **only on the initial checkout of a fresh clone** (all-zeros previous HEAD), so ordinary branch switches and file checkouts don't trigger indexing.
- It carries its own marker (`# tokensave: auto-init`), so it installs idempotently and is also added on the next run for users who already have the post-commit hook.
- Extracted a shared `write_global_hook()` helper (append-or-create + chmod) used by both hooks; added unit tests for the snippet and the writer.

Because it's behind the same explicit opt-in as post-commit, users who don't want auto-init-on-clone simply decline the git-hook prompt — consistent with how `tokensave sync`-on-commit already works.

Thanks @ilteoood for the contribution!

PR #145, opened 2026-06-18
feat(analysis): add path filtering to dead_code, hotspots, unused_imports + config defaults

sam-fakhreddine:
## Summary

- Add `path`, `path_include`, `path_exclude` params to `dead_code`, `hotspots`, and `unused_imports` — the three analysis tools that were missing path-scoping support.
- Add `default_path_include` / `default_path_exclude` config fields in `.tokensave/config.json` for query-level defaults applied when no explicit args are passed.
- Flip `git_ignore` default to `true` so new projects respect `.gitignore` out of the box.

## Motivation

On monorepos with vendored dependencies (e.g. Swift SPM prebuilts, Go vendor/, JS bundles), analysis tools return thousands of results from code the team doesn't own. `dead_code` on a ~57K-node Swift monorepo returned 9,219 results — 93% noise from vendored code, test mocks, and cross-platform duplicates.

`god_class`, `coupling`, `complexity`, and `largest` already have `path` — but `dead_code`, `hotspots`, and `unused_imports` do not, and no tool has `path_exclude`. This forced users to run direct SQLite queries against `.tokensave/tokensave.db` to scope results.

With these changes, the same query goes from 9,219 → 652 results (with explicit filters) or → 31 results (with config defaults).

## Changes

- `src/mcp/tools/definitions.rs` — Added `path`, `path_include`, `path_exclude` to the three tool schemas
- `src/mcp/tools/handlers/analysis.rs` — Wired `effective_path`, `filter_by_path_lists`, `parse_string_array`, and new `with_defaults` helper into `handle_dead_code`, `handle_hotspots`, `handle_unused_imports`
- `src/mcp/tools/handlers/mod.rs` — Added `with_defaults()` helper (returns caller args if non-empty, else config defaults)
- `src/config.rs` — Added `default_path_include` / `default_path_exclude` fields; flipped `git_ignore` default to `true`
- `CHANGELOG.md` — Updated `[Unreleased]` section

## Test plan

- [x] `cargo test --lib` passes (422 tests)
- [x] `cargo test --test tokensave_test` passes (40 tests)
- [x] `cargo clippy` has no new warnings (all warnings are pre-existing)
- [x] Tested manually on a 57K-node Swift monorepo:
  - `dead_code` with `--path Record-App/Sources`: 431 → 2 results
  - `dead_code` with `path_exclude: ["Tests", "Mocks", "Prebuilts"]`: 9,219 → 652
  - `dead_code` with zero args + config defaults: 431 → 31
  - `hotspots` with `path_exclude`: 10 → 1
  - `unused_imports` with `path_exclude`: 568 → 31
  - Re-index after adding `Prebuilts/**` to indexer exclude: 3,299 → 1,687 files

## Checklist

- [x] Breaking changes documented — `git_ignore` default change only affects new `tokensave init`; existing projects keep their persisted config

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **`path`, `path_include`, `path_exclude` params on `dead_code`, `hotspots`, and `unused_imports`.** These three analysis tools were the only ones missing path-scoping support. Same semantics as `god_class`, `coupling`, `complexity`, and `largest`. `path` filters to a directory prefix; `path_include`/`path_exclude` filter by substring (exclude takes precedence).
+- **`default_path_include` / `default_path_exclude` config fields.** Query-level defaults in `.tokensave/config.json` applied to analysis tools when no explicit `path_include`/`path_exclude` is passed. Useful for permanently excluding test/mock directories from analysis results without repeating args on every call.
+
+### Changed
+- **`git_ignore` now defaults to `true`.** New projects indexed with `tokensave init` will respect `.gitignore` rules out of the box. Existing projects are unaffected (their config is already persisted). Use `tokensave gitignore off` or set `"git_ignore": false` in config to opt out.
+
 ### Fixed
 - **Dead-code analysis resolves calls through receiver bindings, `Self::`, and dotted method calls (#141).** After the trait-impl fix (#137), `tokensave_dead_code` still reported functions/methods as dead when their only call sites used call forms the resolver didn't trace. Three gaps are closed:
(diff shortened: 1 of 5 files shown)


aovestdipaperino:
Thanks for this — the path-filtering feature looks solid and the manual numbers on the Swift monorepo are compelling.

Two CI items, only one of which is a real blocker:

1. **`Format` ❌ is the actual blocker.** `cargo fmt --check` reports diffs in `src/mcp/tools/handlers/analysis.rs` at lines 177, 350, 398, 418. Running `cargo fmt` and pushing should clear it.
2. **`Test macOS` ❌ is a flake, not your change.** `graph_test` crashed with `SIGSEGV` *at binary startup, before any test ran* — this is the native-layer instability tracked in #134, unrelated to this PR's logic (all 422 lib tests + Linux + Windows passed). I've kicked off a re-run; it should go green.

One behavior note for the changelog/reviewers: flipping `git_ignore`'s default to `true` only affects new `tokensave init` — existing projects keep their persisted config, which you've already documented. 👍

Once `cargo fmt` is applied and CI is green, this is good to merge.

PR #159, opened 2026-06-30
build: strip release binaries to fix Scoop shim failure (#156)

### LINKED ISSUE #156: Scoop installer: The term 'tokensave.exe' is not recognized as a name of a cmdlet, function, script file, or executable program.
### tokensave version

7.0.2

### Operating system

Windows

### Installation method

Scoop

### What happened?

When I try to install the package with scoop, I get the error during the installation.

I added the bucket successfully:
```
❯ scoop bucket add tokensave https://github.com/aovestdipaperino/scoop-bucket
Checking repo... OK
The tokensave bucket was added successfully.
```

Next, I install the package and get the error:

```
❯ scoop install tokensave
Installing 'tokensave' (7.0.2) [64bit] from 'tokensave' bucket
tokensave-v7.0.2-x86_64-windows.zip (21.5 MB) [======================================================================================] 100%
Checking hash of tokensave-v7.0.2-x86_64-windows.zip ... ok.
Extracting tokensave-v7.0.2-x86_64-windows.zip ... done.
Linking ~\scoop\apps\tokensave\current => ~\scoop\apps\tokensave\7.0.2
Creating shim for 'tokensave'.
Get-Command: C:\Users\USER\scoop\apps\scoop\current\lib\install.ps1:191
Line |
 191 |              $bin = (Get-Command $target).Source
     |                      ~~~~~~~~~~~~~~~~~~~
     | The term 'tokensave.exe' is not recognized as a name of a cmdlet, function, script file, or executable program. Check the spelling of
     | the name, or if a path was included, verify that the path is correct and try again.
Can't shim 'tokensave.exe': File doesn't exist.
```

If I check the executable, then it is present:
```
C:\work
❯ cd $env:USERPROFILE\scoop\apps\tokensave\7.0.2

apps\tokensave\7.0.2
❯ dir

    Directory: C:\Users\USER\scoop\apps\tokensave\7.0.2

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---          24/06/2026    15:20      159328576 tokensave.exe
```

### Steps to reproduce

_No response_

### Error output

```shell

```

### Component

None

### Additional context

_No response_

aovestdipaperino:
## Summary

Fixes #156. Scoop install fails with `Can't shim 'tokensave.exe': File doesn't exist.` even though the exe is present after extraction.

## Root cause

The manifest and release zip are both correct (exe at archive root, `bin: tokensave.exe`), so this is **not** a bucket bug. The Windows release binary ships **unstripped at ~159 MB** — there is no `[profile.release]` in `Cargo.toml` and the release workflow builds with plain `cargo build --release`.

On install, Scoop extracts the zip and then runs `Test-Path "$dir\tokensave.exe"` / `Get-Command` to create the shim (`create_shims` in Scoop's `install.ps1`). Windows Defender's post-extract scan of a binary that large can still hold the file at that moment, so the `Test-Path` checks fail and Scoop aborts the shim — while the file becomes accessible again seconds later.

## Fix

Add `[profile.release] strip = true`, cutting the binary to a fraction of its size and removing the scan/shim race. No behavioral change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,3 +146,11 @@ criterion = { version = "0.5", features = ["async_tokio", "html_reports"] }
 name = "large_repos"
 harness = false
 
+
+[profile.release]
+# Strip debug symbols from release binaries. The unstripped binary is ~159 MB,
+# large enough that Windows Defender's post-extract scan can still hold the file
+# when Scoop tries to create its shim, causing "Can't shim 'tokensave.exe':
+# File doesn't exist." even though the exe is present (#156). Stripping cuts the
+# binary to a fraction of that size and avoids the scan/shim race.
+strip = true


aovestdipaperino:
Closing — I verified this doesn't actually fix #156.

Built the release binary with `strip = true` and measured: **148M → 144M** (only ~4M saved). Cargo's default release profile already sets `debug = false`, so there were never meaningful debug symbols to strip. `size` shows the binary is ~148M of `__TEXT` — bundled tree-sitter grammar tables (40+ parsers), which neither `strip` nor LTO can shrink.

So the binary is irreducibly large and size-reduction is a dead end for this issue. The real cause of the Scoop shim failure is environmental (Defender holding the freshly-extracted large binary during `create_shims`' `Test-Path`/`Get-Command`), which no Cargo/manifest change reliably fixes. A genuinely deterministic fix would be to ship the raw `.exe` as the release asset (URL `#/tokensave.exe` fragment) so Scoop skips extraction entirely — but that's a release-workflow + manifest change worth discussing separately.

Issue #165, opened 2026-07-02
`PreToolUse` hook (`hook-pre-tool-use`) reads the `TOOL_INPUT` env var, but Claude Code sends the event on stdin — the grep/Read→MCP guardrail never fires


rNoz:
### tokensave version

7.0.3

### Operating system

macOS

### Installation method

Homebrew (brew)

### What happened?

tokensave installs a Claude Code `PreToolUse` hook whose job is to stop `Grep`/`Bash grep`/`Read`/`Agent` on code files in an indexed project and redirect the agent to the tokensave MCP tools (where the token savings come from). That hook does not fire on current Claude Code.

`hook_pre_tool_use()` (`src/hooks.rs`) reads its input from an **environment variable named `TOOL_INPUT`**. Per the official Claude Code hooks reference, Claude Code delivers the hook event as **JSON on stdin** and sets **no `TOOL_INPUT` environment variable** (the docs enumerate every env var Claude sets — `CLAUDE_PROJECT_DIR`, `CLAUDE_CODE_REMOTE`, `CLAUDE_EFFORT`, `CLAUDE_ENV_FILE`; `TOOL_INPUT` is absent). So the hook reads an empty string, decides "allow", and never redirects — the guardrail is a silent no-op.

Consequence: `tokensave monitor` can show **0 saved** for a whole session even in an initialized project, because nothing routes the agent to the MCP tools and the guardrail that should enforce it never runs. It fails "open" (allows), so nothing visibly breaks — which is likely why it went unnoticed.

There is also a **shape** mismatch: `evaluate_hook_decision` inspects *top-level* fields (`pattern`, `command`, `subagent_type`), i.e. the bare `tool_input` object, but Claude's stdin payload nests it under `tool_input` (`{ "hook_event_name", "tool_name", "tool_input": { ... } }`).

### Steps to reproduce

1. Install the Claude hooks (`tokensave install --agent claude`) and `tokensave init` a project.
2. In a Claude Code session in that project, run a `Grep`/`Read` on a source file — it is **not** redirected to the tokensave MCP tools.
3. `tokensave monitor` shows 0 (or under-reported) saved for the session.

Deterministic CLI proof (run in any repo with `.tokensave/tokensave.db`):

### Error output

```shell
$ BIN=/opt/homebrew/bin/tokensave

# 1) Claude's REAL transport — event JSON on stdin (nested), a Grep that should block:
[...]

PR #166, opened 2026-07-02
fix(hooks): read Claude PreToolUse event from stdin, not TOOL_INPUT

rNoz:
## Summary

_Disclaimer_ I am still surprised this is broken or a real bug, but found it early this morning and wanted to give it a try. Filled the issue and then started exploring it. Here the PR to review. I left it in draft in case I am missing something.

- Fix the Claude Code `PreToolUse` hook (`tokensave hook-pre-tool-use`), which never fired: it read tool arguments from a `TOOL_INPUT` env var that Claude Code does not set.
- The handler now reads the hook event from **stdin** (Claude's contract) and unwraps the nested `tool_input` object, restoring the grep/Read/Agent → tokensave-MCP guardrail.
- Keeps `TOOL_INPUT` as a fallback (used when stdin is empty) so nothing regresses.
- Solves https://github.com/aovestdipaperino/tokensave/issues/165

## Motivation

Closes #<ISSUE_NUMBER>.

`hook_pre_tool_use()` in `src/hooks.rs` read its input from `std::env::var("TOOL_INPUT")`. Per the official Claude Code hooks reference, Claude Code delivers each hook event as **JSON on stdin** with the tool arguments nested under a top-level `tool_input` key, and sets **no `TOOL_INPUT` environment variable**. The decision logic was correct but always ran against an empty string, so every `Grep`/`Bash grep`/`Read`/`Agent` call was implicitly allowed and nothing redirected exploration to the tokensave MCP tools. Because the hook fails open (empty output = allow), nothing visibly broke, but the token-savings guardrail was a silent no-op, and `tokensave monitor` can report 0 saved for a whole session in an initialized project.

The sibling Kiro hooks already read stdin correctly via the existing `read_stdin_to_string()` helper; only the Claude path used the env var. The other Claude hooks (`hook_prompt_submit`, `hook_stop`) read no tool input from the environment and are unaffected.

Verified against the current Claude Code hooks docs: stdin JSON, top-level `hook_event_name`/`tool_name`/`tool_input`, deny via `hookSpecificOutput.permissionDecision: "deny"`.

## Design note (narrow scope, fails open)

The decision is a **pure function of the tool-input JSON plus a cheap env snapshot** (does `.tokensave/tokensave.db` exist? is `TOKENSAVE_DISABLE_GREP_HOOK` set?). It never calls the tokensave MCP server or touches the DB, so a broken/slow tokensave cannot make the hook block. It only intercepts: an `Explore` subagent, a code-research-style Agent prompt, a symbol-shaped `Grep` on code files in an indexed project, and a `Bash` command that *starts with* `grep`/`rg`/`ag` (after `rtk`/`sudo`) matching the same. Everything else passes through, like arbitrary Bash, `git grep`, non-code greps, `files_with_matches`/`count` modes, regex sweeps, non-`Explore` sub-agents. Malformed/unexpected JSON, empty input, or a hook that errors all resolve to "allow" (fail open), so the guardrail can never wedge an agent. The deny message names the `TOKENSAVE_DISABLE_GREP_HOOK=1` escape hatch; for a `Bash` grep an agent can even set it inline on the one command (`TOKENSAVE_DISABLE_GREP_HOOK=1 grep …`) since the hook reads the Bash child's env.

## Changes

- `src/hooks.rs`:
  - `hook_pre_tool_use()` now reads stdin via `read_stdin_to_string()`; when stdin is non-empty it delegates to the new `evaluate_claude_pre_tool_use()`, otherwise it falls back to the `TOOL_INPUT` env var.
  - New pure `evaluate_claude_pre_tool_use(raw: &str) -> String` unwraps the nested `tool_input` object then delegates to the existing `evaluate_hook_decision()`; if the wrapper shape is absent it treats `raw` as a flat tool-input object.
  - `evaluate_hook_decision` / `evaluate_hook_decision_with_env` are unchanged, so all existing flat-string unit tests still apply.
- `tests/hooks_test.rs`: 5 new tests covering Claude's nested-stdin shape end-to-end (blocks Explore agent, blocks research prompt with the MCP-tools reason, allows a normal tool call, falls back to a flat `tool_input`, allows invalid JSON).

## Test plan

- [x] `cargo test --test hooks_test` passes (60 tests, 5 new)
- [x] `cargo clippy --workspace --all-targets`; no new warnings from this change
- [x] Tested manually against the built binary:
  ```
  # nested stdin (Claude's real shape); now DENIES (was silent before):
  printf '{"hook_event_name":"PreToolUse","tool_name":"Grep","tool_input":{"pattern":"install_mcp_server","glob":"*.rs"}}' | tokensave hook-pre-tool-use
  # -> {"hookSpecificOutput":{...,"permissionDecision":"deny",...}}

  # env-var fallback still works:
  TOOL_INPUT='{"pattern":"install_mcp_server","glob":"*.rs"}' tokensave hook-pre-tool-use
  # -> deny JSON

  # normal tool call; allowed (no output):
  printf '{"tool_name":"Bash","tool_input":{"command":"cargo test"}}' | tokensave hook-pre-tool-use
  ```
- [x] Verified in a **real Claude Code session** (hook installed, project indexed). Asked the agent to run a Bash symbol grep without using tokensave; the tool call was **blocked pre-execution** and returned verbatim:
  > STOP: This Bash grep targets a code file in a tokensave-indexed project and the pattern `install_mcp_server` looks like a symbol name. Use tokensave_search (definition) or tokensave_callers_for (usages) instead. symbol-indexed lookups are faster and more accurate than text grep. To override for this one call, set TOKENSAVE_DISABLE_GREP_HOOK=1 in the shell.

  Re-running the same command as `TOKENSAVE_DISABLE_GREP_HOOK=1 grep -rn install_mcp_server src/` executed normally (exit 1, no matches). When blocked, the agent redirected to `tokensave_find_exact_symbol` / `tokensave_search` on its own. (Before this fix the same session performed the raw grep and recorded 0 tokens saved.)

  image

  White block is the startup message (company motd or whatever). Tested with Sonnet 5. Claude v2.1.198. Tokensave installed locally in the project where I ran `claude`.

## Checklist

- [x] Breaking changes documented (if any): none; env-var path retained as fallback

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Claude Code `PreToolUse` hook never fired — the grep/Read→MCP guardrail was silently dead.** `hook_pre_tool_use` read the tool arguments from a `TOOL_INPUT` environment variable, but Claude Code delivers the hook event as JSON on **stdin** (with the tool arguments nested under `tool_input`) and sets no such variable. The decision logic was correct, but it always ran against an empty input, so every Grep/Read/Agent call was implicitly allowed and nothing redirected exploration to the tokensave MCP tools — under-reporting `tokensave monitor` savings. The handler now reads Claude's stdin contract (unwrapping the nested `tool_input`), keeps the `TOOL_INPUT` env var as a fallback for empty stdin, and preserves the existing flat-input decision logic. The sibling Kiro hooks already read stdin correctly and are unaffected.
+
 
 ## [7.0.3] - 2026-06-30
 
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -129,16 +129,37 @@ enum PatternShape {
 
 /// `PreToolUse` hook handler for Claude Code's Agent / Grep / Bash matchers.
 ///
-/// Reads the `TOOL_INPUT` environment variable (JSON), inspects the input,
-/// and prints a JSON decision to stdout. Blocks Explore agents,
-/// exploration-style prompts, and symbol-shaped grep/Grep calls against
-/// indexed code files — directing Claude to use tokensave MCP tools instead.
+/// Claude Code delivers the hook event as JSON on **stdin** with the tool
+/// arguments nested under `tool_input`; it sets no `TOOL_INPUT` env var.
+/// Reads stdin, inspects the input, and prints a JSON decision to stdout.
+/// Blocks Explore agents, exploration-style prompts, and symbol-shaped
+/// grep/Grep calls against indexed code files — directing Claude to use
+/// tokensave MCP tools instead. Falls back to the `TOOL_INPUT` env var when
+/// stdin is empty.
 pub fn hook_pre_tool_use() {
-    let tool_input = std::env::var("TOOL_INPUT").unwrap_or_default();
-    let decision = evaluate_hook_decision(&tool_input);
+    let raw = read_stdin_to_string();
+    let decision = if raw.trim().is_empty() {
+        evaluate_hook_decision(&std::env::var("TOOL_INPUT").unwrap_or_default())
+    } else {
+        evaluate_claude_pre_tool_use(&raw)
+    };
     if !decision.is_empty() {
         println!("{decision}");
     }
+}
+
+/// Parse Claude Code's `PreToolUse` stdin JSON and return the decision string.
+///
+/// Unwraps the nested `tool_input` object before delegating to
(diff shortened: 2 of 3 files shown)


aovestdipaperino:
Merged — thank you @rNoz! Great catch, and an even better write-up: the guardrail failing open meant this never showed up as breakage, just as silently missing savings, and your issue #165 + before/after session evidence made the diagnosis airtight. Keeping the TOOL_INPUT fallback and routing everything through the existing flat-input decision logic (so all prior tests still apply) was exactly the right shape. I applied one cosmetic clippy touch-up on top (map().unwrap_or_else() → map_or_else()) to keep the workspace pedantic-clean. Thanks for verifying against a real Claude Code session too!

PR #175, opened 2026-07-05
fix(db): coalesce NULL kind in `get_stats` instead of aborting

rNoz:
## Summary

- Fix `get_stats` hard-failing on a single row with a NULL `kind`, which showed the monitor 0 savings even when savings were real.
- Read `kind` as `Option<String>` and coalesce NULL to `"unknown"` instead of erroring the whole aggregate.
- Solves https://github.com/aovestdipaperino/tokensave/issues/176

## Motivation

Fixes the issue "`get_stats` aborts on a single NULL `kind` row". `query_kind_counts` read the `kind` column as a non-null `String`, so one NULL row (e.g. an older-schema ancestor DB served via the untracked-branch fallback) failed the entire `get_stats` call with `failed to read kind: Null value`. Because the monitor and dashboard read `get_stats`, a genuinely-working tokensave looked completely dead (0 savings). `get_stats` is a read-only aggregate and must fail open, not hard-fail on one malformed row;  the scalar counts in the same function already defend against NULLs.

## Changes

- `src/db/queries/mod.rs` — `query_kind_counts` reads `kind` as `Option<String>`, coalesces NULL → `"unknown"`, and merges counts (so a NULL group and a literal `"unknown"` group combine).
- Regression test `query_kind_counts_coalesces_null_kind` (uses a scratch table with a NULL row, since `nodes`/`edges` are `NOT NULL`).

## Test plan

- [x] `cargo test` passes
- [x] `cargo clippy --workspace --all-targets` clean
- [x] Verified `tokensave_status` / `get_stats` returns full stats on an untracked branch

## Checklist

- [x] `CHANGELOG.md` updated (under `[Unreleased]` → Fixed)
- [x] No breaking changes (nothing to doc)


aovestdipaperino:
Thanks @rNoz — fail-open on a malformed row is exactly right for a read-only aggregate, and the regression test nails it down. Merging. 🙏

Issue #179, opened 2026-07-05
Transparently track the current git branch so tokensave "just works" on feature branches


rNoz:
### Problem

tokensave only serves a per-branch DB for branches recorded in `BranchMeta` (`.tokensave/branches.json`), created by `tokensave branch add <b>`. On any **untracked** branch (the common case: a fork working on `rnoz/integration` while only `master` is tracked), two things happen:

1. Queries silently fall back to the nearest tracked ancestor DB (`find_nearest_tracked_ancestor`, `src/branch.rs:132`); stale but usable.
2. `get_stats` hard-crashed on a NULL `kind` row from that fallback path, showing the monitor **0 savings** (see `get-stats-null-kind-crash.md`).

I ran `tokensave branch add rnoz/integration` by hand on every fork to un-blind the monitor. Every new branch or cherry-pick needs the same manual step, and agents don't remember to. This makes tokensave look broken across projects when it is actually working.

### Current mechanism (for reference)

- `TokenSave::new` (`src/tokensave/mod.rs`) resolves `active_branch` **once at startup** from git (`branch::current_branch`).
- `resolve_db_for_branch` / `open_branch` (`src/tokensave/mod.rs:301,361`) pick the DB: tracked branch → its `db_file`; untracked branch → ancestor fallback.
- `handle_branch_action::Add` (`src/commands.rs:8`) copies the nearest-ancestor DB into a new `db_file`, records `{parent, db_file, last_synced_at}` in `BranchMeta`, and runs an incremental sync.
- `branch_meta::is_tracked` already answers "is this branch known?".

### Design options

**A. Auto-track at query/startup time.** When a `.tokensave` DB and a `BranchMeta` exist, `active_branch` is `Some`, and `!is_tracked(active_branch)`, run the existing `branch add` logic automatically before serving the first query. No git-hook install; works identically for every harness (Claude, Droid, Auggie, Copilot, Codex, bare CLI) because it lives in tokensave itself.

- Refactor: extract the copy-ancestor-DB + record-meta body out of `handle_branch_action::Add` into a reusable `branch::ensure_branch_tracked(project_root, branch) -> Result<bool>` (idempotent: no-op if already tracked). Call it from both the CLI `Add` arm and the new auto path. **No new code duplicated, no new dependency.**
- Slot the call in `resolve_db_for_branch` (or once in `TokenSave::new` after `active_branch` is known).

[...]

PR #180, opened 2026-07-05
feat(branch): transparently auto-track the current git branch

rNoz:
## Summary

- Transparent auto-track of the current git branch when it is untracked, so tokensave "just works" on feature branches with no manual `tokensave branch add`.
- **Two complementary paths**: a `post-checkout` git hook (primary, git-native, reuses the existing hook install infra) plus a best-effort auto-track inside `TokenSave::open` (safety net for harness entry points that don't fire git hooks).
- Off by default; opt in via `config.auto_track` or `TOKENSAVE_AUTO_TRACK=1`. Pairs with the discussion in https://github.com/aovestdipaperino/tokensave/issues/179 and is offered as one concrete solution; the design space (A / B / C) is open there.

## Motivation

On an untracked branch tokensave falls back to the nearest-ancestor DB. Queries still save tokens, but the served index is the ancestor's (stale for branch-changed code) and, until fixed separately, `get_stats` could crash. The operator has to run `tokensave branch add` by hand on every new branch, and agents don't remember to. tokensave already installs a `post-commit` hook (auto-`sync`) and a `post-checkout` hook (auto-`init` on fresh clone); the missing piece is "branch create/switch doesn't track the new branch."

## Changes

- `src/agents/mod.rs`: `post_checkout_snippet` gains a branch-checkout arm. On a real branch checkout (`$3 == 1`, non-fresh-clone) it runs `tokensave branch add` (no-op if already tracked or default). Reuses the existing hook install + chain infra, no new dependency, no `.git/hooks` clobber work.
- `src/branch.rs`: new `track_branch_copy` copies the nearest-ancestor DB and records the branch in `BranchMeta`. No incremental sync (would re-enter `open()`); freshness is the `post-commit` hook's job.
- `src/tokensave/mod.rs`: `TokenSave::open` best-effort auto-tracks an untracked active branch. Skips when there is no branch metadata (single-DB mode), when the active branch is the default branch, or on detached HEAD. Never fails `open()`, only logs.
- `src/config.rs`: new `auto_track` field (default `false`) plus the `TOKENSAVE_AUTO_TRACK` env override (`0/false/no/off/""` all disable). `auto_track` and the env combine so operators can carry per-repo settings via config and disable per run via the env.

## Test plan

- [x] `cargo test` passes; `track_branch_copy` (idempotency / ancestor copy) and `post_checkout_snippet`'s branch arm are covered.
- [x] `cargo clippy --workspace --all-targets` clean.
- [x] E2E on a fresh branch: `open()`-path auto-track copies the ancestor DB, `status` returns full stats without crashing, `TOKENSAVE_AUTO_TRACK=0` opts out cleanly, and `branch add` is idempotent.

## Known limitations (deliberate for the PoC; happy to address per the discussion)

- Copies only the ancestor `.db`, matching the existing `branch add`. An uncheckpointed WAL on the ancestor would be missed. Upgrade path: checkpoint-before-copy, applied to both paths together.
- `save_branch_meta` is a plain write, not atomic (temp + rename). Concurrent `track_branch_copy` on the same branch could lose a metadata write; the idempotent `is_tracked` guard makes a lost race otherwise harmless. Harden separately.
- Every touched branch gets a DB copy; `branch gc` reaps deleted branches and would benefit from a delete-hook wire-up.

## Checklist

- [x] `CHANGELOG.md` updated under `[Unreleased]` (Added).
- [x] Breaking changes documented (none; off by default, additive only).

## Disclaimer (as this is in Draft and for the discussion)

_Note_ I leave it in draft as it should serve as a part of the discussion/proposal of  https://github.com/aovestdipaperino/tokensave/issues/179 but this PR/code is working and tested, although I am still evolving it as I consider better approaches. This is fully functional as it is but, for instance, my next approach (whenever I find more free hours for this project) will be to consider something like this:

   1. Extract env+config parsing into branch::should_auto_track(config_auto_track) -> bool.
   2. TokenSave::open calls the helper.
   3. Add a new hidden subcommand tokensave hook-auto-track-branch (the hook entry point) that calls the helper and tracks only if true; otherwise silent no-op. => So we achieve runtime opt-in if we want.
   4. Change post_checkout_snippet so the hook calls hook-auto-track-branch instead of branch add.
   5. CLI tokensave branch add [branch] stays unconditional — that's the user's explicit command.

Why? Because I am thinking how this will be easier to reason for the user, even requiring some more code changes: same single knob for both complementary paths. Default off even with hooks installed; toggle without reinstall.


rNoz:
Hi @aovestdipaperino: would be great to know your thoughts about this PR approach (and the related issue/discussion), otherwise I plan to dig into what I say in the Disclaimer section, but I would like to know your thoughts/design decisions. Have a good one!

rNoz:
I'll set in ready for review so you can think about this approach and the original issue I opened with the different options, in case you want a layered strategy as in the other discussion issue I brought that you liked.
CC @aovestdipaperino

rNoz:
Rebased and conflicts solved.

aovestdipaperino:
Merged — thank you @rNoz for this contribution and for the clear write-up of the design trade-offs and known limitations (WAL checkpoint on copy, non-atomic branch-meta save). The opt-in default keeps existing behavior safe while making feature branches "just work" for those who enable it. Follow-ups for the listed limitations are welcome!

PR #184, opened 2026-07-07
fix: extract Julia definitions with bundled grammar

p-pietro:
## Summary

- Julia extraction recorded zero functions and structs from every `.jl` file — only file/module stub nodes were indexed, leaving tokensave blind to Julia code (no symbol search, call graph, or impact analysis).
- Root cause: the extractor read definition names via `child_by_field_name("name"/"body"/"function")`, but the bundled `tree-sitter-julia` 0.23 grammar exposes those positionally, not through named fields — so every lookup returned `NULL` and the whole subtree was silently dropped.
- Extractor now falls back through `signature`/`type_head` to the first identifier, traverses module children directly, and derives call callees positionally. Failed name lookups now recurse into children instead of dropping the subtree.
- Adds a `julia_extraction_test` + fixture (mirroring `dart_extraction_test`/`proto_extraction_test`), which Julia previously had none of — the reason this shipped silently.

## Motivation

Inspecting the bundled grammar's `node-types.json` confirms `function_definition`, `struct_definition`, `abstract_definition`, and `macro_definition` carry no `name` field (only `module_definition` does), and `call_expression` has no `function` field. The extractor's field-name assumptions were written against a different (local) grammar build and never re-validated against the version that shipped — the same class of grammar-shape mismatch fixed earlier in `d1b379e` and `afe7386` for the batch/dart/proto extractors. Because `sync` reported plausible node/edge counts, the failure was silent.

## Reproduction

Against `JuliaLang/julia`'s `base/` stdlib (206 `.jl` files):

| kind | before | after |
|---|---|---|
| function | 0 | 4424 |
| struct (class) | 0 | 427 |
| module | 47 | 418 |
| macro | 0 | 191 |
| calls edges | 0 | 14121 |

Minimal repro — `tokensave init && tokensave sync --force` on a directory with one file:

```julia
struct Point
    x::Float64
    y::Float64
end

function distance(a::Point, b::Point)
    return sqrt((a.x - b.x)^2 + (a.y - b.y)^2)
end
```

Before: a single `file` node. After: `Point`, `distance`, and a `calls` ref to `sqrt`.

## Changes

- `src/extraction/julia_extractor.rs`: add `definition_name` (name → `signature` → `type_head` → first identifier, via the shared `find_child_by_kind`/`find_descendant_by_kind` in `ts_state`), `call_callee`/`first_call_child` for positional callees, and `extract_function_calls` that skips the `signature` subtree. `visit_module` now walks children directly. All `visit_*` handlers fall back to `visit_children` on a failed name lookup instead of returning.
- `tests/julia_extraction_test.rs` + `tests/fixtures/julia_module.jl`: assert module/struct/function extraction, `contains` edges, and a `sqrt` call ref.
- `CHANGELOG.md`: entry under `[Unreleased] ### Fixed`.

## Test plan

- [x] `cargo test` — full suite green (452 unit + all integration tests, 0 failures).
- [x] `cargo fmt --all -- --check` clean.
- [x] Re-sync of `JuliaLang/julia` `base/` produces the counts above.

## Checklist

- [x] `CHANGELOG.md` updated under `[Unreleased] ### Fixed`.
- [x] No breaking changes (extraction-only; no schema or API change).


aovestdipaperino:
Thanks for the fix, @p-pietro — rebased onto latest master and tests pass locally. Merging once CI is green. 🙏

Issue #185, opened 2026-07-07
`tokensave_diff` fails with "delegated tool returned non-JSON text" on large or empty diffs


rNoz:
### tokensave version

7.0.3

### Operating system

macOS

### Installation method

Built from source

### What happened?

Calling the MCP `tokensave_diff` tool fails hard with:

```
config error: delegated tool returned non-JSON text: key must be a string at line 672 column 1
```

`tokensave_diff` is an aggregator: it delegates to `handle_diff_context` / `handle_changelog` / `handle_commit_context`, then nests their result under a `changes` envelope. The failure fires whenever the delegated payload is large (any real diff), and also on the trivial "no changes" case.

### Steps to reproduce

1. In an indexed repo with a non-trivial working-tree diff (enough changed symbols that the delegate's JSON exceeds `MAX_RESPONSE_CHARS`), call `tokensave_diff` with no args.
2. It returns the `delegated tool returned non-JSON text` config error instead of a diff.
3. A clean working tree also fails, because the delegate returns the plain status line `No changes detected.` which is not JSON.

### Error output

```shell
config error: delegated tool returned non-JSON text: key must be a string at line 672 column 1
```

### Component

MCP server

### Additional context

Each delegate builds a structured `serde_json::Value`, then formats it with `serde_json::to_string_pretty` and passes it through `truncate_response` before returning it as the `text` field of a `ToolResult`. `handle_diff` then re-parses that text back into a `Value` via a private `extract_value` helper (`serde_json::from_str`).

When the delegate's formatted payload exceeds `MAX_RESPONSE_CHARS`, `truncate_response` appends `\n\n[... truncated at N chars]`, so the text is no longer valid JSON and the re-parse fails. The same happens for the plain-text status lines the delegates emit on the empty ("No changes detected.") and git-error ("git error: ...") paths, which were never JSON to begin with. The whole `tokensave_diff` call errors out as a result.

### Proposed fix

[...]

PR #186, opened 2026-07-07
fix(mcp/git): compose diff delegates as structured Values, not re-parsed text

rNoz:
## Summary

- Fix `tokensave_diff` failing with "delegated tool returned non-JSON text" whenever the delegated payload is large (any real diff) or empty.
- Compose delegate results as structured `Value`s instead of re-parsing their already-formatted-and-truncated text.
- First stab at https://github.com/aovestdipaperino/tokensave/issues/185

## Motivation

Fixes the issue "`tokensave_diff` fails with delegated tool returned non-JSON text on large or empty diffs". `handle_diff` delegated to `handle_diff_context` / `handle_changelog` / `handle_commit_context`, each of which formats its payload with `to_string_pretty` and then `truncate_response`. It then re-parsed that text back into JSON via `extract_value`. When the delegate's payload exceeded `MAX_RESPONSE_CHARS`, `truncate_response` appended a non-JSON suffix, so the re-parse failed and the whole call errored. The plain status lines ("No changes detected.", "git error: ...") broke it the same way. This means `tokensave_diff` was unusable on real diffs.

## Changes

- `src/mcp/tools/handlers/git.rs`: split `handle_diff_context`, `handle_changelog`, and `handle_commit_context` into private `*_value` helpers returning the structured `Value` (plus touched files); the public `handle_*` tools are now thin wrappers that format and truncate once.
- `handle_diff` composes the delegates' raw `Value`s directly and truncates once on the final envelope; the empty and git-error states become structured JSON so they nest cleanly.
- Deleted the `extract_value` re-parse shim.
- Regression test `handle_diff_composes_valid_json_without_reparse` drives `handle_diff` down a non-JSON delegate branch and asserts a parseable envelope.

## Test plan

- [x] `cargo test --lib mcp::tools::handlers::git` passes (incl. new regression)
- [x] `cargo clippy --all-targets` clean
- [x] `cargo fmt --check` clean

## Checklist

- [x] `CHANGELOG.md` updated (under `[Unreleased]` -> Fixed)
- [x] Behavior change: empty/error states now returned as structured JSON (previously plain text); more consistent, no known consumers rely on the old text. But please, review on your end as well.


aovestdipaperino:
Thanks for the fix, @rNoz — merging. 🙏

Issue #187, opened 2026-07-07
Cursor `preToolUse` hook blocks every Grep/Shell call: "Hook returned invalid JSON"


rNoz:
### tokensave version

7.0.3

### Operating system

macOS

### Installation method

Built from source

### What happened?

With the tokensave `hook-pre-tool-use` wired into Cursor's `preToolUse` (matcher `Grep|Shell`), Cursor blocks every Grep and Shell call and reports:

```
Hook ... returned invalid JSON
```

The agent cannot run any shell command or grep. Claude Code with the same binary works fine, so the hook logic is correct; only the stdout contract differs.

### Steps to reproduce

1. Register `tokensave hook-pre-tool-use` as a Cursor `preToolUse` hook (matcher `Grep|Shell`).
2. Ask the agent to run any Shell or Grep tool call.
3. Cursor fails the call with "Hook ... returned invalid JSON" and the tool never runs.

### Error output

```shell
Hook ... returned invalid JSON
```

### Component

Agent integration (Claude, Cursor, etc.)

### Additional context

Cursor's `preToolUse` is a permission gate: it reads a flat `permission` field ("allow" / "deny") from the hook's stdout, and treats any payload without that field as fail-closed, reporting "returned invalid JSON". The hook printed nothing on allow (an empty decision meant "allow" for Claude Code, whose contract is empty-stdout = allow) and, on deny, printed only Claude Code's nested `hookSpecificOutput.permissionDecision` object. Neither carries the flat `permission` field, so under Cursor every call was blocked: allow produced empty stdout, deny produced JSON Cursor did not recognize as a decision.

### Proposed fix

Emit a cross-harness payload from `hook_pre_tool_use`:

- Allow: print `{"permission":"allow"}` instead of nothing.
- Deny: add flat `permission:"deny"` plus `user_message`/`agent_message` alongside the existing nested `hookSpecificOutput`.

Claude Code ignores the unknown flat `permission` field and falls through to its normal permission flow, so one payload satisfies both harnesses. Also skip install/flush maintenance for the `Hook*` subcommands, which run on the per-tool-call hot path and must keep stdout JSON-only.

### Additional context

The failure mode is silent and total under Cursor (no tool works), and it is invisible on Claude Code, so it is easy to misattribute to the Cursor integration or the environment rather than the hook's stdout format.

aovestdipaperino:
[...]

PR #188, opened 2026-07-07
fix(hooks): emit Cursor-native permission JSON from preToolUse hook

rNoz:
## Summary

- Fix the tokensave `preToolUse` hook blocking every Grep/Shell call under Cursor with "Hook returned invalid JSON".
- Emit a cross-harness permission payload that both Cursor and Claude Code accept from one hook.
- Solving https://github.com/aovestdipaperino/tokensave/issues/187, which is nasty and a total blocker. Disclaimer: I have only tested using `agent` (cursor cli, not the GUI app while investigating this and testing implementations/fixes).

## Motivation

Fixes the issue "Cursor `preToolUse` hook blocks every Grep/Shell call". Cursor's `preToolUse` is a permission gate that reads a flat `permission` field from stdout and fails closed when it is missing. The hook printed nothing on allow and only Claude Code's nested `hookSpecificOutput` on deny, so under Cursor every tool call was blocked. Claude Code, whose contract is empty-stdout = allow, worked with the same binary, which is why the format bug was easy to miss.

## Changes

- `src/hooks.rs`: `hook_pre_tool_use` now prints `{"permission":"allow"}` on allow (new `build_allow_message`); `build_block_message` adds flat `permission:"deny"` plus `user_message`/`agent_message` alongside the existing nested `hookSpecificOutput`. Claude Code ignores the unknown flat field.
- `src/main.rs`: `should_skip_agent_install_maintenance` now also skips the `Hook*` subcommands, which run on the per-tool-call hot path and must keep stdout JSON-only for the permission gate.
- Inline tests `allow_message_carries_cursor_permission_field` and `block_message_is_cross_harness`.

## Test plan

- [x] `cargo test --test hooks_test` passes (60)
- [x] `cargo test --lib cursor_decision_tests` passes (2)
- [x] `cargo test --bin tokensave startup_tests` passes (should_skip)
- [x] `cargo clippy --all-targets` clean, `cargo fmt --check` clean
- [x] Manual: `echo '{...}' | tokensave hook-pre-tool-use` prints `{"permission":"allow"}`; Explore-agent input prints `permission:"deny"` + `hookSpecificOutput`

## Checklist

- [x] `CHANGELOG.md` updated (under `[Unreleased]` -> Fixed).
- [x] No breaking changes.

Tested other harnesses (Claude) and it is unaffected (unknown flat field ignored). I won't have Cursor for much longer, but at least giving my 2cents after hours and hours of digging deep into issues (more to come when I tackle them).


aovestdipaperino:
Thanks for the fix, @rNoz — rebased onto latest master and merging. 🙏

Issue #189, opened 2026-07-07
`tokensave init` writes `.tokensave` to the tracked `.gitignore` (why tracking it?)


rNoz:
## `tokensave init` writes `.tokensave` to the tracked `.gitignore` (why tracking it?)

Enhancement / behavior discussion. 

Since I propose changing a default behavior, worth flagging for proper review.

### tokensave version
current `master`

### Component
CLI (`init`)

### What's the situation?
`tokensave init` appends `.tokensave` to the project's **tracked** `.gitignore`. `.tokensave/` is a per-developer index, not a project artifact, so writing it into the tracked `.gitignore` commits a per-developer choice into shared project history: every developer who runs `init` produces a committable diff to a shared file.

Git already provides local, untracked ignore mechanisms for exactly this case, like `.git/info/exclude` and the global excludes file, and tokensave already *reads* them via `git check-ignore` (`src/config.rs: is_ignored_by_git`); it just doesn't *write* there.

### Proposal
On `tokensave init`:
1. If `git check-ignore -q .tokensave/` already succeeds (global excludes, `.git/info/exclude`, or an existing `.gitignore` rule) → do nothing.
2. Otherwise write `.tokensave/` to `.git/info/exclude` (untracked, local-only), resolved via `git rev-parse --git-path info/exclude` so worktrees and custom `$GIT_DIR` layouts work; a non-repo is a no-op.
3. Keep an explicit `[g]` opt-in for the old tracked-`.gitignore` behavior.

### Decision requested (which default?)
This changes the default, so it's the maintainer's call. Two shapes are available, so happy to ship either:

- **Option A (this PR):** default → `.git/info/exclude` (local); opt-in `[g]` → tracked `.gitignore`; `[n]` → nothing. Rationale above.
- **Option B (conservative):** keep the current default → tracked `.gitignore`; add opt-ins for `.git/info/exclude` or nothing. No default change; users who want local-only ignore opt in.

Either way, projects that already ignore `.tokensave/` never prompt, and there is no regression for users who already have `.tokensave` in `.gitignore`.

About to upload a PR I have been working on during the last days.

aovestdipaperino:
Fixed in #190 (commit 4d33c8c) — `tokensave init` now excludes `.tokensave` via `.git/info/exclude` instead of the tracked `.gitignore`. Thanks, @rNoz! 🙏

PR #190, opened 2026-07-07
feat(init): exclude .tokensave via .git/info/exclude, not .gitignore

rNoz:
## Summary

- `tokensave init` currently appends `.tokensave` to the project's **tracked** `.gitignore`. For a per-developer, local-only index, that commits a per-developer choice into shared project history.
- Prefer the **local, untracked** ignore mechanisms instead: skip entirely if `git check-ignore` already ignores `.tokensave/` (global excludes / existing rules); otherwise write to `.git/info/exclude`. Offer the tracked `.gitignore` only as an explicit opt-in.
- First design and implementation of this proposal (Option A) https://github.com/aovestdipaperino/tokensave/issues/189, since I was for days using it and decided to promote as an issue/PR, if you find it useful.

## Motivation

`.tokensave/` is a per-developer index, not a project artifact. Writing it into the tracked `.gitignore` means every developer who runs `tokensave init` produces a committable diff and pushes a per-developer choice into shared project history. `.git/info/exclude` and the global excludes file exist for exactly this "ignore locally, without touching tracked files" case. tokensave already *reads* those via `git check-ignore` (`src/config.rs: is_ignored_by_git`); it just doesn't *write* there.

## Proposed behavior

On `tokensave init`:
1. If `git check-ignore -q .tokensave/` already succeeds (global excludes, `.git/info/exclude`, or an existing `.gitignore` rule) → do nothing.
2. Else, by default, append `.tokensave/` to `.git/info/exclude` (create if missing): untracked, local-only.
3. Keep an explicit opt-in for the old behavior (e.g. `--gitignore` flag or a prompt choice) for users who *do* want it in the tracked `.gitignore`.

## Changes

- `src/config.rs`: add `add_to_git_info_exclude(project_path)`. Appends `.tokensave/` to the repo's local `.git/info/exclude`, creating the file if missing, idempotent (skips if a `.tokensave` entry already exists). The exclude path is resolved via `git rev-parse --git-path info/exclude` (private `git_info_exclude_path` helper), so worktrees and custom `$GIT_DIR` layouts work and non-repos are a no-op. `add_to_gitignore` is unchanged.
- `src/commands.rs` (`init_and_index` flow): the existing "not ignored yet" prompt becomes a three-way choice: default/`[Y]` → `.git/info/exclude` (local), `[g]` → tracked `.gitignore` (explicit opt-in, old behavior), `[n]` → nothing. The `is_in_gitignore` guard (which already consults `git check-ignore`) means an already-ignored project never prompts.
- `src/main.rs` (`status`): the "not excluded" warning now suggests `echo '.tokensave/' >> .git/info/exclude`.
- `tests/config_test.rs`: new tests. Creates the entry (and leaves tracked `.gitignore` untouched); idempotent; appends after `git init`'s seeded comments; git actually ignores the entry (isolated from the developer's global excludes via an empty `GIT_CONFIG_GLOBAL`); and a non-repo is a no-op.
- `CHANGELOG.md` `[Unreleased]`.

## Test plan

- [x] `cargo test --test config_test` passes (28 tests, 5 new)
- [x] `cargo clippy --workspace --all-targets`: no new warnings from this change
- [x] Manual: `tokensave init` in a repo with no ignore → `.git/info/exclude` gets `.tokensave/`, tracked `.gitignore` untouched; re-run is a no-op; a repo whose global excludes already ignore `.tokensave/` is left completely alone; answering `g` writes the tracked `.gitignore` as before.

## Checklist

- [x] No behavior change for users who already have `.tokensave` in `.gitignore`
- [x] Backward-compatible opt-in preserved (`[g]` at the prompt), but the default is changing previous behavior. Please, be sure you review carefully if you like this proposal. Otherwise, we can offer it the other way around: default (gitignore, as it is right now), opt-ins excluded or nothing.

I hope you like it. Totally open for new directions or better approaches, I'll happily mutate this branch and test how it would be locally myself.


aovestdipaperino:
Thanks for the fix, @rNoz — merging. 🙏

Issue #195, opened 2026-07-09
PreToolUse guardrail: should it hard-block agent *delegation*, or only steer?


rNoz:
This is not a bug but a general concern in terms of the design of tokensave (yeah, thanks for this tool). So as in https://github.com/aovestdipaperino/tokensave/issues/179, here I would like to know your vision, I am providing some options I am thinking and starting an implementation considering one that I test locally (not saying is the best based on your criteria, that's why I am asking).

The `PreToolUse` guardrail that steers `grep`/`Read`/agent work toward the MCP tools is a genuinely nice idea, and this is a design question about it, not a bug report. I don't know the full history of the guardrail, so I may be missing intent; please correct me.

While running Claude Code inside a tokensave-indexed project, I asked the agent to delegate an *implementation* task to a cheaper sub-agent (`subagent_type: "general-purpose"`), a normal "cheap model does the bulk, expensive model reviews" workflow. The guardrail denied the delegation with `STOP: Use tokensave MCP tools … instead of agents for code research.` The sub-agent was an implementer, not a researcher; it was denied purely because its prompt text contained research-like words ("explore", "find callers of X", "conventions"). There was no way to override and continue in the session: the agent branch of the hook does not honor the `TOKENSAVE_DISABLE_GREP_HOOK` opt-out that the `Grep`/`Bash` branches honor. The only way forward was to abandon the session, change configuration, and start over; the guardrail interrupted legitimate work rather than guiding it.

## What the code does today

In `src/hooks.rs`, `evaluate_hook_decision_with_env` denies an Agent call when either `subagent_type == "Explore"` or `is_code_research_prompt(prompt)` is true (the second regardless of `subagent_type`, and neither branch checks `env.disable_grep_hook` or `env.cwd_has_tokensave_db`, unlike the `Grep` and `Bash` branches). The `.claude/settings.json` matcher is `Agent|Grep|Bash` (note: `Read`/`Glob` are not matched, so file-reading fan-out is not redirected today). I can see this has been deliberate since the original native hook (v1.2.0), which is exactly why I'm asking rather than asserting.

## Thoughts

1. **Deny vs steer.** Is a hard `deny` the intended behavior for the agent case, or would an allow-plus-`additionalContext` hint (steer, never block) better match the "optimization" intent? A guardrail that blocks a legitimate, deliberate delegation seems to work against its own goal.
[...]

PR #196, opened 2026-07-09
fix(hooks): scope agent guardrail to Explore + honor opt-out

rNoz:
## Summary

- The `PreToolUse` guardrail denied any Agent/Task call whose prompt looked like research, regardless of `subagent_type`, so a deliberately-typed non-Explore delegation (e.g. `general-purpose`, an implementer, a custom or other-harness agent type) was hard-blocked with no override.
- Now: the built-in `Explore` agent is still steered to the MCP tools; an explicitly-typed non-Explore agent is never blocked on prompt text; an *untyped* Agent/Task call is still steered by prompt shape; and the agent path honors the existing `TOKENSAVE_DISABLE_GREP_HOOK` opt-out, giving an explicit override.
- Proof-of-concept for the "open discussion" in https://github.com/aovestdipaperino/tokensave/issues/179 (Option B). **Totally open** to implement a different direction (steer-not-deny / config-driven / inner-tool-only) per that discussion, but I am waiting for your point of view :)

## Motivation

The guardrail's intent is to *steer* code-research work toward the tokensave MCP tools to save tokens. But the agent branch hard-denied on prompt keywords for any `subagent_type`, and (unlike the `Grep`/`Bash` branches) ignored the opt-out env and the index check. Prompt keywords cannot distinguish research from implementation, and callers increasingly use explicit, purpose-built agent types (implementer, reviewer, custom Factory-droid types, other harnesses' task types). The result was that a legitimate, deliberate delegation was blocked with no in-session way to continue; the guardrail interrupted work rather than guiding it. See the linked discussion for the concrete case and the fuller option space.

## Changes

- `src/hooks.rs`, `evaluate_hook_decision_with_env`: restructured the agent branch. Gated the whole agent block on `!env.disable_grep_hook`. Kept the `subagent_type == "Explore"` deny. Applied the `is_code_research_prompt` deny only when `subagent_type` is absent (untyped Task). An explicitly-typed non-Explore agent now falls through to allow. The `Grep` (`pattern`) and `Bash` (`command`) branches and the Kiro delegation path (which also uses `is_code_research_prompt`) are unchanged.
- `tests/hooks_test.rs`: added coverage. A typed non-Explore agent with a research-y prompt is allowed; an untyped research Task is still blocked; the opt-out lets an `Explore` agent through. Existing agent/grep/bash/kiro tests unchanged and still pass.

## Test plan

- [x] `cargo test --test hooks_test`: 64 passed (including 4 new cases: typed non-Explore allowed on a research prompt; untyped research Task still blocked; opt-out lets `Explore` through; blank `subagent_type` treated as untyped).
- [x] `cargo clippy --workspace --all-targets`: no new warnings.
- [x] `cargo fmt --check`: clean.
- [x] Verified against the real hook binary with Claude Code's stdin payload shape: a `general-purpose` agent with a research-y prompt is allowed; a built-in `Explore` agent is still denied; an untyped (and a blank-`subagent_type`) research Task is still denied.

## Checklist

- [x] Breaking changes documented: none. Strictly loosens an over-broad block and adds an opt-out, and the built-in Explore behavior plus the `Grep`/`Bash`/`Kiro` paths are unchanged. Here what's important is to know your views and your design decisions, since Option B was just one way to do it as a PoC and to open discussion.

## Known limitation / out of scope (see https://github.com/aovestdipaperino/tokensave/issues/179)

Unlike the `Grep`/`Bash` branches, the agent branch does not gate on the presence of a `.tokensave` index (`cwd_has_tokensave_db`). This predates this change (it has been so since the original hook) and is left as is here to keep the PoC minimal. Also, it is raised as a consistency question in the linked discussion (the "right layer" / cross-path-consistency concern), since fixing it touches the existing agent-block tests and is really part of the larger design decision.


rNoz:
Hi @aovestdipaperino, I set it in Ready for Review, but because the PoC Option B is done and solves the case in a _simple way_, not that I say this should be the proper design decision to be implemented, that's why I opened it as part of the discussion with a set of options I was considering...

aovestdipaperino:
I will personally take a look. Thank you.

aovestdipaperino:
Thank you @rNoz for the excellent writeup in #195 and the carefully-scoped PoC — the tests covering the blank-subagent_type edge were a nice touch. I resolved the conflict with the `evaluate_hook_decision_core` refactor from master and merged. Follow-ups per the discussion in #195: gating the agent branch on `cwd_has_tokensave_db` (your point 6), and longer-term the tool-level config in `.tokensave/project.json`.

Issue #197, opened 2026-07-09
Feature request: `PreToolUse` guardrail hook for Factory Droid


rNoz:
Factory Droid support (#163 / PR #162) shipped MCP registration + AGENTS.md rules, but **not** the `PreToolUse` guardrail hook that drives tokensave's token savings on the other hook-capable agents. Today only `claude` and `kiro` install a guardrail hook; Droid users get MCP tools but nothing nudges Droid to use them instead of raw `grep`/`Read`/sub-agent exploration.

Factory now ships a hook mechanism (`docs.factory.ai/cli/configuration/hooks-guide`), so the follow-up deferred in #163 is unblocked.

### Proposed behavior

Install a Droid `PreToolUse` hook, registry-wired like the other agents, that reuses tokensave's existing shared decision core (`evaluate_hook_decision_with_env` / `is_code_research_prompt` / `TOKENSAVE_RESEARCH_BLOCK_REASON`) and only adds a thin per-agent wire-format adapter.

### Factory hook contract 

> I have reviewed the docs (structure, hooks) and performed a bunch of tests

- Input: **stdin JSON**, tool payload nested under `.tool_input` (`.tool_input.command` for the `Execute` shell tool).
- Block: **exit code 2 + stderr** (the Kiro mechanism, not Claude's stdout JSON).
- Config: `hooks` object in `~/.factory/settings.json` (same file as the MCP entry).
- Escape hatch honored: hook subprocess inherits Droid env, so `TOKENSAVE_DISABLE_GREP_HOOK` is passed through.

### Design decisions (that would be key!)

- **Fail open**: a broken, slow, or mis-indexed tokensave must never block a Droid tool call. Malformed or empty input lets the call through (same class of bug as #165).
- **Narrow matcher**: only symbol-grep-shaped `Execute` calls on code files in an indexed project, and code-research sub-agent/task launches. Arbitrary shell, `git grep`, non-code files, and specialized sub-agents with normal tasks pass.
- **Only fires in an indexed project** (`.tokensave/tokensave.db` in cwd).

### Expected ACs

`tokensave install --agent droid` writes the hook into `~/.factory/settings.json`, preserving all other keys. `uninstall` removes only tokensave's entry. `doctor --agent droid` reports it. End-to-end tests cover the Droid wire shape, fail-open, narrow-matcher, and sub-agent-passes cases.

rNoz:
The huge contrib of #198 has been just merged! Solved!

Issue #163, opened 2026-07-02
Support Factory Droid


rNoz:
Factory Droid is a widely used agentic CLI that supports MCP servers and the AGENTS.md instruction standard, but tokensave had no integration for it. This brings Droid to parity with the other supported agents so its users get the same code-graph tooling.

aovestdipaperino:
Implemented by #162 (merged, thanks @rNoz): `tokensave install --agent droid` registers the MCP server in `~/.factory/mcp.json` and appends prompt rules to `~/.factory/AGENTS.md`; `--local` writes project-scoped `.factory/mcp.json` and a repo-root `AGENTS.md`. Auto-detected when `~/.factory` exists. Will ship in the next release.

PR #162, opened 2026-07-02
feat(agent): add Factory Droid agent integration

rNoz:
## Summary

- Add **Factory Droid** (the `droid` CLI) as a `tokensave install --agent droid` target.
- Registers the tokensave MCP server in `~/.factory/mcp.json` and appends prompt rules to `~/.factory/AGENTS.md` (with `--local` support for project-scoped config).
- Follows the JSON-MCP + AGENTS.md pattern of OpenCode/Pi/Codex; no new dependencies.
- Solves https://github.com/aovestdipaperino/tokensave/issues/163

## Motivation

Factory Droid is a widely used agentic CLI that supports MCP servers and the AGENTS.md instruction standard, but tokensave had no integration for it. This brings Droid to parity with the other supported agents so its users get the
same code-graph tooling. 

Factory also supports Claude-style hooks; wiring those is intentionally left as a follow-up to keep this PR minimal and consistent with every other non-Claude agent. And even more, start using it asap, as I am doing locally.

## Changes

- **`src/agents/droid.rs`** (new): `DroidIntegration` implementing `AgentIntegration`:
  - MCP server written to `~/.factory/mcp.json` under `mcpServers.tokensave`   with Factory's stdio shape `{ "type": "stdio", "command": <bin>, "args": ["serve"] }`;    `--local` writes `<project>/.factory/mcp.json`.
  - Prompt rules appended to `~/.factory/AGENTS.md` (global) or a repo-root    `<project>/AGENTS.md` (`--local`), reusing the OpenCode marker/idempotency logic.
  - install/uninstall touch only the `tokensave` entry/rules block (unrelated    servers, keys, and AGENTS.md content preserved; files deleted only when empty).
  - No hooks/permissions (Factory uses runtime approval, like OpenCode/Pi).
- **`src/agents/mod.rs`**: register `droid` in the four registry sites, so  `install` / `reinstall` / `uninstall` / `doctor` / auto-detect all cover it  (all registry-driven; no per-command changes needed).
- **`src/cli.rs`**: list `droid` in the `--local` flag's project-scoped agents.
- **Docs**: `README.md` (command list + agent prose lists), `docs/USER-GUIDE.md`,  `CHANGELOG.md` (`[Unreleased]`).

## Test plan

- [x] `cargo test` passes; `tests/droid_agent_test.rs` (new, 18 tests: install/uninstall, idempotency, server/key/content preservation, `--local`,  detection, healthcheck) and `tests/agent_test.rs` (registry count + enumerations).
[...]

PR #198, opened 2026-07-09
feat(droid): add PreToolUse guardrail hook for Factory Droid

rNoz:
## Summary

Watch out, this is a long description because I am providing e2e proofs and manual tests in fresh sessions, repeating the whole process (starting, installing for the agent, dropping custom AGENTS.md rules, verifying hooks isolated, etc). Should serve as "docs" of the methology, as it hasn't been straightforward since I started providing support for Droid. More ideas to come. Furthermore, I have direct communication with Factory folks (in case it is required at some point), so I am also exchanging bug reports and asking for features on parallel. For instance, in the "details" section we can see some discoveries (like double hook execution for no reason, but it is not harmful)

- `tokensave install --agent droid` now also installs a `PreToolUse` guardrail hook, the same class of hook `claude` and `kiro` already ship, so Droid users get the token-saving redirect to tokensave MCP tools instead of raw grep/Bash exploration.
- Droid's hook contract (confirmed live, not just from public docs): stdin JSON with the tool payload nested under `tool_input`; blocks via **exit code 2 + stderr** (the same mechanism as the Kiro hook, not Claude's stdout-JSON decision); the `hooks` object lives in **`~/.factory/settings.json`**, a separate file from `mcp.json`, and the public doc's `hooks.json` is wrong.
- No new decision logic: the Droid adapter is a thin wire-format shim over the same shared core (`evaluate_hook_decision_with_env` → `evaluate_hook_decision_core`) the Claude and Kiro hooks already use, so the fail-open / narrow-matcher / escape-hatch invariants are inherited, not re-implemented.
- Implements https://github.com/aovestdipaperino/tokensave/issues/197

## Motivation

Factory Droid support (#163 / PR #162) shipped MCP registration + AGENTS.md prompt rules but explicitly deferred the guardrail hook because Factory had no documented hook mechanism at the time. Factory now ships one (`docs.factory.ai/cli/configuration/hooks-guide`), so this closes that gap: today `claude` and `kiro` steer exploration to tokensave MCP tools and Droid does not, even though Droid users get the same MCP tools.

## Changes

- **`src/hooks.rs`**
  - Extracted the classification logic shared by every `PreToolUse`-style hook into a private `evaluate_hook_decision_core(tool_input, env) -> Option<String>`. `evaluate_hook_decision_with_env` (Claude's stdout-JSON path) now just wraps that `Option` in `build_block_message`. Pure refactor; no behavior change for Claude or Kiro, and all existing tests pass unchanged.
  - Added `hook_droid_pre_tool_use() -> i32` (reads stdin, prints the reason to stderr and returns 2 on block, 0 on allow; mirrors `hook_kiro_pre_tool_use`), `evaluate_droid_pre_tool_use(raw)` (real-environment convenience wrapper), and `evaluate_droid_pre_tool_use_with_env(raw, env) -> Option<String>` (unwraps the nested `tool_input`, falling back to a flat payload, then delegates to the shared core).
- **`src/agents/droid.rs`**
  - Added `install_hook` / `uninstall_hook` / `doctor_check_hook`, wired into `DroidIntegration::install` / `uninstall` / `healthcheck`.
  - Writes/removes a single array entry under `settings.json`'s `hooks.PreToolUse`, matched on `"matcher": "Execute"` (Droid's confirmed shell tool; see "Confirmed vs. unconfirmed" below), preserving every other top-level key and every other hook entry (including hooks a user manages themselves, e.g. their own permission wrapper).
  - Corrected the module doc comment, which claimed "Factory Droid has no hook system"; no longer true.
  - Global path: `~/.factory/settings.json`. `--local`: `<project>/.factory/settings.json` (same relative layout as `mcp.json`).
- **`src/cli.rs` / `src/main.rs`**: new hidden subcommand `hook-droid-pre-tool-use`, dispatched the same way as `hook-kiro-pre-tool-use` (non-zero exit code propagated via `process::exit`).
- **`tests/hooks_test.rs`**: 12 new cases for the Droid wire shape. Blocks a symbol-shaped `grep` on a Rust file; allows terminal-launched tools (Plannotator, `cargo build`, `git commit`); allows `git grep`; allows when not indexed; honors `TOKENSAVE_DISABLE_GREP_HOOK=1`; a specialized sub-agent with a normal (non-research) task passes; falls back to a flat payload; allows empty/invalid input; the block reason documents the escape hatch.
- **`tests/droid_agent_test.rs`**: 10 new cases. Writes the hook; preserves an existing owner-managed `PreToolUse` wrapper and unrelated `Stop` hook and top-level keys; idempotent; `--local` writes the project-scoped file; uninstall removes only tokensave's entry; uninstall without a prior install doesn't crash; healthcheck reports the hook missing/present.

### Confirmed vs. unconfirmed (matcher scope)

Only the `Execute` matcher is installed. Factory's hook-events list proves Droid has *some* sub-agent/task concept (`SubagentStop` fires for it), but the exact tool name that launches one is not in the public docs and could not be confirmed without spending on a live orchestration run (I will run separately, to be 100% clear we are not implementing it based on hints, but actual proofs - wanted to isolate the **Execute** -confirmed- from the rest). Per the fail-open invariant, we do not guess it for now: an unconfirmed tool name is simply never wired to a matcher, so a call through that tool never reaches this hook at all (the safest form of "fail open" for something we can't verify). The shared decision core (which also inspects `subagent_type`/`prompt` fields, for the Claude Agent/Task shape) is still exercised directly in `tests/hooks_test.rs::test_droid_specialized_subagent_with_normal_task_passes` so the invariant holds if that matcher scope is ever widened later.

## Test plan

- [x] `cargo test --test hooks_test`: 70 passed, 0 failed (12 new Droid cases; all prior Claude/Kiro/Grep/Bash cases pass unchanged).
- [x] `cargo test --test droid_agent_test`: 26 passed, 0 failed (10 new hook cases; all prior MCP/AGENTS.md cases pass unchanged).
- [x] `cargo test` (full workspace): every test binary reports `0 failed` (448 in the main `tests` binary, the rest split across ~90 integration-test binaries; verbatim `test result:` lines captured, all `ok`).
- [x] `cargo clippy --workspace --all-targets`: clean, no warnings.
- [x] `cargo fmt --check`: clean.
- [x] Real `droid exec` session (see "Real-world verification" below): the hook fires exit-2/stderr on a live symbol grep, and the `TOKENSAVE_DISABLE_GREP_HOOK=1` escape hatch resolves it in the same session, without the human restarting.

## Real-world verification

All steps below used `--settings <path>` (droid's "runtime settings file merged for this process only" flag) and `--local`/`--cwd` pointed at a scratch project, so **the user's real `~/.factory/settings.json` was never modified**. The tokensave binary used was this branch's own `cargo build --release` output. 

### 1. Install

```
$ cd /tmp/droid-hook-live-verify && tokensave install --agent droid --local
✔ Added tokensave MCP server to /private/tmp/droid-hook-live-verify/.factory/mcp.json
✔ Appended tokensave rules to /private/tmp/droid-hook-live-verify/AGENTS.md
✔ Added tokensave PreToolUse hook to /private/tmp/droid-hook-live-verify/.factory/settings.json
```

Resulting `.factory/settings.json`:

```json
{
    "hooks": {
        "PreToolUse": [
            {
                "matcher": "Execute",
                "hooks": [
                    {
                        "type": "command",
                        "command": "<tokensave-bin> hook-droid-pre-tool-use"
                    }
                ]
            }
        ]
    }
}
```

### 2. Doctor

```
$ tokensave doctor
Factory Droid integration
  ✔ MCP server registered in /Users/<home>/.factory/mcp.json
  ✔ MCP server args include "serve"
  ✔ AGENTS.md contains tokensave rules
  ✘ PreToolUse hook NOT installed in /Users/<home>/.factory/settings.json; run `tokensave install --agent droid`
```

This ran against the real global `~/.factory/`: `doctor` always checks the global path regardless of a prior `--local` install, matching the pre-existing behavior of every other agent's doctor check in this codebase (not something this PR changes). The hook correctly reports missing globally since only the local scratch install was performed.

### 3. Live deny + escape-hatch re-run (verbatim, captured from the session's `.jsonl` transcript)

Prompt sent via `droid exec --settings <scratch>/.factory/settings.json --cwd <scratch> "Run this exact shell command and show me the output: grep -rn handle_request_for_widget src/main.rs"` against a scratch project containing `src/main.rs` with a `handle_request_for_widget` function, indexed with `tokensave init`.

**First attempt; the guardrail fires (exit code 2, stderr shown to the model):**

```
tool_use: Execute { command: "grep -rn handle_request_for_widget src/main.rs" }
hookStatus: error
hookResults[0]: {
  exitCode: 2,
  stderr: "STOP: This Bash grep targets a code file in a tokensave-indexed project and the
  pattern `handle_request_for_widget` looks like a symbol name. Use tokensave_search
  (definition) or tokensave_callers_for (usages) instead. Symbol-indexed lookups are
  faster and more accurate than text grep. To override for this one call, set
  TOKENSAVE_DISABLE_GREP_HOOK=1 in the shell."
}
tool_result: is_error: true, content: "Error: STOP: ... (same message)"
```

**Model self-recovers using the documented escape hatch, in the same session, without a human touching anything:**

```
tool_use: Execute { command: "TOKENSAVE_DISABLE_GREP_HOOK=1 grep -rn handle_request_for_widget src/main.rs" }
hookStatus: completed
hookResults[0]: { exitCode: 0, stderr: "" }
tool_result: is_error: false, content: "src/main.rs:1:fn handle_request_for_widget() {\nsrc/main.rs:6:    handle_request_for_widget();\n\n\n[Process exited with code 0]"
```

Final assistant message: *"Note: A hook intercepted the initial grep call, suggesting the tokensave MCP tools instead. I ran it with `TOKENSAVE_DISABLE_GREP_HOOK=1` to honor your 'exact command' request."*

This is the exact round trip the handoff asked for: a live deny with the verbatim message, and a re-run under `TOKENSAVE_DISABLE_GREP_HOOK=1` that succeeds, with no session restart.

### 4. Cleanup

```
$ tokensave uninstall --agent droid --local
✔ Removed /private/tmp/droid-hook-live-verify/.factory/mcp.json (was empty)
✔ Removed /private/tmp/droid-hook-live-verify/AGENTS.md (was empty)
✔ Removed /private/tmp/droid-hook-live-verify/.factory/settings.json (was empty)
```

Scratch directory removed; the user's real `~/.factory/settings.json` (`md5 e05a23096ceb84e1c93a48955aaf7f6d`) was unchanged throughout.

### Note on the session transcript

The session's `.jsonl` log shows the `Execute`-matcher hook command listed **twice** in `hookCommands` for the grep call (both identical, running our binary), alongside the owner's own catch-all `factory-permission.sh` wrapper (which has its own unrelated `matcher: ""` entry in the *global* settings and is unaffected by anything in this PR). Only one entry for our hook exists in the actual `settings.json` file on disk; the duplicate invocation appears to be an artifact of Droid's own hook-dispatch/merge behavior when a `--settings` override is combined with the global config, not something this PR's install code writes. It has no effect on correctness: both invocations returned the identical exit-2/stderr decision (the function is a pure, idempotent classifier), and the escape-hatch re-run passed both hook invocations with exit 0. Flagged here for visibility, not fixed, since it lives in Factory's dispatcher rather than in the tokensave-owned code this PR changes.

## Additional live verification: rules-removed hook test

<details>
<summary>Click to expand: methodology, session ID, and per-test-case results</summary>

### Methodology

To confirm the hook works even when Droid has no tokensave prompt rules to steer it, we temporarily removed the rules from the AGENTS.md files Droid reads, then ran direct `Execute` prompts in a `tokensave init`-ed project. The goal was to verify that the `PreToolUse` hook alone blocks symbol-shaped grep and honors the escape hatch.

Files changed before the session:
- Backed up `/Users/<user>/tokensave/AGENTS.md` to `AGENTS.md.bak` and emptied it.
- `~/.factory/AGENTS.md` was also emptied before the test but was restored externally before the session started, so the later ambiguous prompt was not fully rules-free. The direct `Execute` tests are unaffected by this.

Session ID: `62238022-70de-461c-baff-9f4d7c3c5870`

### Test case 1: symbol-shaped grep should be blocked

**Prompt:**
> Run this shell command and show me the output: `grep -rn handle_request src/main.rs`

**Tool call:**
```json
{ "name": "Execute", "input": { "command": "grep -rn handle_request src/main.rs" } }
```

**Hook commands executed (in order):**
1. `/Users/<user>/.local/bin/tokensave hook-droid-pre-tool-use`
2. `/Users/<user>/<other tokensave-unrelated hooks>/factory-permission.sh`
3. `/Users/<user>/.local/bin/tokensave hook-droid-pre-tool-use`

**Hook results:**
- tokensave hook: exit 2, stderr:
  ```
  STOP: This Bash grep targets a code file in a tokensave-indexed project and the pattern `handle_request` looks like a symbol name. Use tokensave_search (definition) or tokensave_callers_for (usages) instead — symbol-indexed lookups are faster and more accurate than text grep. To override for this one call, set TOKENSAVE_DISABLE_GREP_HOOK=1 in the shell.
  ```
- permission wrapper: exit 0
- tokensave hook (second invocation): exit 2, same stderr

**Overall `hookStatus`:** `error`

**Verdict:** Correct. The hook blocked the symbol-shaped grep.

### Test case 2: escape hatch should allow

**Prompt:**
> Run this shell command and show me the output: `TOKENSAVE_DISABLE_GREP_HOOK=1 grep -rn handle_request src/main.rs`

**Tool call:**
```json
{ "name": "Execute", "input": { "command": "TOKENSAVE_DISABLE_GREP_HOOK=1 grep -rn handle_request src/main.rs" } }
```

**Hook results:**
- tokensave hook: exit 0
- permission wrapper: exit 0
- tokensave hook (second invocation): exit 0

**Overall `hookStatus`:** `completed`

**Verdict:** Correct. The opt-out let the command through.

### Test case 3: normal command should pass

**Prompt:**
> Run this shell command and show me the output: `echo hello`

**Tool call:**
```json
{ "name": "Execute", "input": { "command": "echo hello" } }
```

**Hook results:** all three hooks exit 0.

**Overall `hookStatus`:** `completed`

**Verdict:** Correct. A non-grep command is allowed.

### Test case 4: escape hatch repeated (same as #2)

**Prompt:** same as test case 2.

**Result:** same as test case 2 — all hooks exit 0, `hookStatus: completed`.

**Verdict:** Correct.

### Test case 5: ambiguous research request (not fully rules-free)

**Prompt:**
> Find all functions named `handle_request` in this codebase and list their callers.

**Tools used by the agent:**
- `TodoWrite`
- `ToolSearch`
- `tokensave___tokensave_find_exact_symbol`
- `tokensave___tokensave_callers_for`
- `tokensave___tokensave_node`

No `Execute`, `Grep`, `Read`, or `Glob` tool was used. Because the global `~/.factory/AGENTS.md` had been restored to contain the tokensave rules before this prompt, this test case does not prove the hook alone would have redirected the agent; it only shows that the agent chose tokensave MCP tools when rules were present.

### Observations from the session

1. **Duplicate hook invocations.** For every `Execute` event, Droid listed the tokensave hook command **twice** in `hookCommands`. Only one entry exists in `settings.json`; the duplicate is a Droid dispatcher merge artifact. Both invocations returned identical decisions, so there is no functional impact.
2. **No hook firing on non-Execute tools.** `TodoWrite`, `ToolSearch`, and the tokensave MCP tools did not invoke the tokensave hook because the matcher is `Execute` only. This is the intended behavior.
3. **The hook output is shown to the model.** When the hook exits 2, Droid surfaces the stderr message to the model, which then self-recovered using the documented `TOKENSAVE_DISABLE_GREP_HOOK=1` escape hatch.

### Cleanup

After the session, the project-level `AGENTS.md` was restored from its backup.

</details>

## Future directions: a "soft" / wildcard mode

Factory's public hook docs already support regex and wildcard matchers (`*`, `""`, or patterns like `Execute|Grep|Read|Glob|Task`). This PR intentionally registers only the confirmed `Execute` matcher to stay fail-open while the integration is validated.

A follow-up enhancement can widen the matcher to `Execute|Grep|Read|Glob|Task` and classify incoming events by payload shape:

- payload has a `command` string → treat as `Execute` / Bash path
- payload has a `pattern` / `path` / `glob` → treat as Grep path
- payload has `subagent_type` / `prompt` → treat as Agent/Task path
- anything else → allow

Critical safeguards for the wider matcher:
- Always allow tokensave's own MCP tools (`tokensave___*` / `mcp__tokensave__*`).
- Always allow Droid meta tools like `TodoWrite` and `ToolSearch`.
- Treat unknown tool shapes as allow (fail open).

Session analysis for a code-research prompt (`2f785343-5354-4c66-b331-16efd20b23c9`, annotating it here if you want proofs, so I don't forget all the tests that I have done in a fresh session + droid execs) showed the agent went straight to `tokensave___tokensave_search` and `tokensave___tokensave_callers` without using `Execute`, `Grep`, `Read`, or `Glob`. This means the current `Execute`-only hook already covers the worst-case scenario (the agent tries a shell grep) while the prompt rules cover the best-case scenario (the agent uses tokensave MCP tools directly). Widening the matcher is a safe next step once the Droid payloads for `Grep`/`Read`/`Task` are tested.

## Checklist

- [x] Breaking changes documented: none.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Factory Droid `PreToolUse` guardrail hook (follow-up to the Droid MCP integration, #163).** `tokensave install --agent droid` now also registers a `PreToolUse` hook under the `hooks` object in `~/.factory/settings.json` (`<project>/.factory/settings.json` for `--local`) — the same file Factory's own hook wrappers live in, a separate file from `mcp.json`. Droid delivers the hook event as JSON on stdin with the tool payload nested under `tool_input` (the Claude/Kiro shape) and blocks a tool call via **exit code 2 + stderr**, the same mechanism as the Kiro hook. The install only registers the `Execute` matcher (Droid's confirmed shell tool) and reuses the same shared decision core as the Claude and Kiro hooks, so symbol-shaped `grep`/`rg`/`ag` calls against indexed code files redirect to tokensave MCP tools while arbitrary shell, `git grep`, non-code files, and specialized sub-agent tasks pass through untouched; `TOKENSAVE_DISABLE_GREP_HOOK=1` opts out per-call. Install/uninstall preserve every other key and hook entry already in `settings.json` (including hooks the user manages themselves); `doctor --agent droid` reports the hook's status.
+
 ## [7.1.0] - 2026-07-09
 
 ### Added
--- a/src/agents/droid.rs
+++ b/src/agents/droid.rs
@@ -3,10 +3,16 @@
 //!
 //! Handles registration of the tokensave MCP server in Factory Droid's MCP
 //! config (`~/.factory/mcp.json` globally, `<project>/.factory/mcp.json` for
-//! `--local`) under the `mcpServers.tokensave` key, and prompt rules via
+//! `--local`) under the `mcpServers.tokensave` key, prompt rules via
 //! `AGENTS.md` (`~/.factory/AGENTS.md` globally, `<project>/AGENTS.md` for
-//! `--local`). Factory Droid has no hook system or declarative tool
-//! permissions — it uses interactive runtime approval.
+//! `--local`), and the `PreToolUse` guardrail hook in Factory's `hooks`
+//! object (`~/.factory/settings.json` globally, `<project>/.factory/settings.json`
+//! for `--local` — the same file Factory's own `Stop`/`Notification`/
+//! `PreToolUse`/`UserPromptSubmit` wrappers live in, per the live-config
+//! verification in Factory's public hook docs; this is a separate
+//! file from `mcp.json`). Droid blocks a tool call via **exit code 2 +
+//! stderr** — the same mechanism as Kiro's `preToolUse` hook, not Claude
+//! Code's stdout JSON decision.
 
 use std::io::Write;
 use std::path::{Path, PathBuf};
@@ -23,6 +29,21 @@ use super::{
 
 /// Factory Droid agent.
 pub struct DroidIntegration;
+
+/// Hook event name under `settings.json`'s `hooks` object that Droid fires
+/// before a tool call executes.
+const DROID_PRE_TOOL_EVENT: &str = "PreToolUse";
+
+/// The `Execute` tool is Droid's confirmed shell tool — the only tool name we
+/// register a matcher for. The sub-agent/task launch tool name is
+/// unconfirmed in Factory's public docs, so we
+/// deliberately do not guess a matcher for it: an unmatched tool never
+/// reaches our hook subprocess at all, which is the safest form of
+/// "fail open" for a tool name we can't verify.
+const DROID_HOOK_MATCHER: &str = "Execute";
+
+/// Subcommand invoked by the installed hook.
+const DROID_PRE_TOOL_HOOK: &str = "hook-droid-pre-tool-use";
 
 impl AgentIntegration for DroidIntegration {
     fn name(&self) -> &'static str {
@@ -44,6 +65,9 @@ impl AgentIntegration for DroidIntegration {
         let agents_md = droid_agents_md_for(ctx);
         install_prompt_rules(&agents_md)?;
 
+        let settings_path = droid_settings_path_for(ctx);
+        install_hook(&settings_path, &ctx.tokensave_bin)?;
+
         eprintln!();
         eprintln!("Setup complete. Next steps:");
         eprintln!("  1. cd into your project and run: tokensave init");
@@ -58,6 +82,9 @@ impl AgentIntegration for DroidIntegration {
         let agents_md = droid_agents_md_for(ctx);
         uninstall_prompt_rules(&agents_md);
 
+        let settings_path = droid_settings_path_for(ctx);
+        uninstall_hook(&settings_path);
+
         eprintln!();
         eprintln!("Uninstall complete. Tokensave has been removed from Factory Droid.");
         eprintln!("Start a new droid session for changes to take effect.");
@@ -68,6 +95,7 @@ impl AgentIntegration for DroidIntegration {
         eprintln!("\n\x1b[1mFactory Droid integration\x1b[0m");
         doctor_check_config(dc, &ctx.home);
         doctor_check_prompt(dc, &ctx.home);
+        doctor_check_hook(dc, &ctx.home);
     }
 
     fn is_detected(&self, home: &Path) -> bool {
@@ -112,6 +140,19 @@ fn droid_agents_md_for(ctx: &InstallContext) -> PathBuf {
         InstallScope::Global => ctx.home.join(".factory/AGENTS.md"),
         InstallScope::Local { project_path } => project_path.join("AGENTS.md"),
     }
+}
+
+/// Global Factory Droid settings path (`~/.factory/settings.json`) — the file
+/// Factory reads its `hooks` object from.
+fn droid_settings_path(home: &Path) -> PathBuf {
+    home.join(".factory/settings.json")
+}
+
+/// settings.json path for this install: `~/.factory/settings.json` globally,
+/// or `<project>/.factory/settings.json` for `--local` (same relative layout
+/// as `mcp.json`).
+fn droid_settings_path_for(ctx: &InstallContext) -> PathBuf {
+    ctx.base_dir().join(".factory/settings.json")
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


aovestdipaperino:
Thank you

PR #217, opened 2026-07-12
fix: preserve symlinked config files during atomic writes

janstol:
## Summary

- Fixes `safe_write_json_file()` destroying symlinked config files (e.g. a dotfiles-managed `~/.claude/settings.json`) during `install`/`uninstall`/`doctor --fix`.
- Adds `resolve_symlink_target()` so writes land on the symlink's real target instead of replacing the symlink itself, including through multi-hop and dangling symlink chains.
- Resolution now fails safely (returns `Result`, aborts the write) instead of falling back to writing over the symlink when a chain can't be resolved — a cycle, an unreadable link, or a chain deeper than 40 hops.
- Adds regression tests and a CHANGELOG entry.

## Motivation

`safe_write_json_file()` is the shared write helper used by every agent integration (claude, gemini, zed, qwen, augment, droid) across `install`, `uninstall`, `doctor --fix`, and the settings auto-backfill. It stages a `.new` sibling file and does `rename(new, path)` to atomically replace `path`.

`rename(2)` replaces the *directory entry* at `path`. A dotfiles setup commonly tracks `~/.claude/settings.json` in a repo and symlinks it into place. When `path` is that symlink, the rename deletes the symlink and drops a plain regular file in its stead — the dotfiles source is never written, so the live config silently detaches from the repo it was meant to track, and the symlink has to be recreated by hand.

During review, two follow-on gaps in the first version of this fix were found and fixed in this same PR:

1. A multi-hop dangling chain (`config -> intermediate -> not-yet-created target`) broke `canonicalize()` outright, and the fallback only resolved one hop, so `intermediate` itself got replaced by a regular file.
2. Once (1) was fixed, any resolution failure (a cycle, an unreadable link, or an overlong chain) still fell back to writing at the original symlink path — reproducing the original bug for exactly the cases meant to be guarded against. A follow-up fixed this by making resolution return `Result` and aborting the write on failure instead of falling back, plus fixed an off-by-one in the hop-counting loop that rejected a chain of *exactly* the 40-hop limit even when it resolved to a valid target.

I didn't find an existing tracking issue for this; happy to file one separately if you'd prefer it referenced here.

## Changes

- `src/agents/mod.rs`: `safe_write_json_file()` now resolves `path` through `resolve_symlink_target()` before staging and renaming, returning `Result<PathBuf>`.
  - Non-symlink paths resolve to themselves; behavior is unchanged.
  - A live symlink resolves via `canonicalize()`.
  - A symlink `canonicalize()` can't resolve (dangling, possibly multi-hop) is walked by `walk_dangling_symlink_chain()`, following `read_link()` one hop at a time (relative targets resolved against each link's parent) until it reaches a non-symlink path. Cycle detection via a `seen` set; a 40-hop (`MAX_SYMLINK_HOPS`) budget against pathological chains, checked immediately before following a link so a chain of exactly 40 hops that lands on a resolvable target still succeeds.
  - If resolution fails (cycle, unreadable link, chain over 40 hops), the write is aborted with an error and the original symlink(s) are left untouched — no fallback to writing over the symlink itself.
  - The `.new` staging file is written next to the resolved real target (not the symlink's location) so the rename stays on one filesystem and remains atomic. The JSON merge logic itself is untouched; only the write *location* changes.
- `src/agents/mod.rs` (`safe_config_tests` module): new unit tests covering symlink preserved + target updated, target in a different directory than the link, a broken single-hop symlink, a multi-hop dangling chain, a symlink cycle (asserts the write fails *and* both links remain untouched), an overlong dangling chain, a chain of exactly 40 hops (must succeed), and a chain of 41 hops (must still fail).
- `CHANGELOG.md`: `[Unreleased]` / `### Fixed` entry rewritten to describe the full fix.

## Test plan

- [x] `cargo test` passes (`agents::` suite: 110 tests, including the symlink/dangling-chain/cycle/boundary tests)
- [x] Tested manually: built the binary and ran `tokensave install --agent claude` with `HOME` pointed at a throwaway directory containing `~/.claude/settings.json` symlinked to a separate "dotfiles" file. After install, `readlink` still pointed at the dotfiles path, and the dotfiles file (not a new plain file at the `.claude` location) contained the new hooks/permissions, confirmed byte-identical via `diff`. Re-ran install a second time to confirm idempotency — symlink still intact, no leftover `.new` files. The multi-hop/cycle/boundary cases added afterward are covered by the unit tests above rather than re-run manually.

## Checklist

- [x] Breaking changes documented (if any) — N/A, no breaking changes; non-symlink behavior is unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Writing agent config no longer destroys a symlinked config file.** `safe_write_json_file` (used by every agent's `install`/`uninstall`/`doctor --fix`, e.g. `~/.claude/settings.json`) staged a `.new` sibling and atomically renamed it over the target path. `rename(2)` replaces the target's directory entry, so when the target was a symlink — a common dotfiles pattern that tracks config in a repo and symlinks it into `~/.claude/` — the rename deleted the symlink and left a plain regular file in its place, silently detaching the live config from the dotfiles source it was meant to track. The write now resolves the symlink to its real target first and stages/renames there, so the dotfiles source gets updated and the symlink itself survives untouched. Non-symlink targets are unaffected. Resolution walks multi-hop and dangling symlink chains (`config -> intermediate -> not-yet-created target`) hop by hop rather than relying on `canonicalize`, which fails outright on a dangling chain. If a chain can't be resolved safely — a cycle, an unreadable link, or a chain deeper than 40 hops — the write is aborted and the original symlinks are left untouched instead of falling back to writing over the symlink itself, which would reproduce the same destructive bug for that case. A chain of exactly 40 hops that lands on a valid target still resolves; only a chain needing a 41st hop is rejected.
+
 ## [7.2.0] - 2026-07-11
 
 ### Added
--- a/src/agents/mod.rs
+++ b/src/agents/mod.rs
@@ -438,17 +438,38 @@ pub fn safe_write_json_file(
         });
     }
 
-    // 3. Ensure parent dir
-    if let Some(parent) = path.parent() {
+    // 3. Resolve symlinks. If `path` (e.g. `~/.claude/settings.json`) is a
+    //    symlink — common for dotfiles setups that track config in a repo and
+    //    symlink it into place — `rename()` over `path` would delete the
+    //    symlink and drop a plain file in its stead, silently detaching the
+    //    live config from the dotfiles source. Write through the symlink to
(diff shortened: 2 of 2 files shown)


aovestdipaperino:
Thank you, @janstol! Excellent work on this — the fix is well-scoped to the single `safe_write_json_file` choke point, and the follow-on gaps you caught during review (the multi-hop dangling chain and the resolution-failure fallback that would've reproduced the original bug) are exactly the kind of edge cases that make a fix like this genuinely robust. The test coverage is thorough — the cycle, exact-40-hop, and 41-hop boundary cases especially. Merged with your authorship preserved. Much appreciated! 🙏

PR #220, opened 2026-07-13
fix(sync): make subprocess extraction pool return results in file order

krishnaglick:
WorkerPool::extract_files distributed files to N worker threads via a
shared queue and pushed each result into a Mutex<Vec<ExtractTuple>> as
threads finished — a race, so the returned order depended on which
worker happened to complete which file first. That order fed straight
into all_nodes -> DB row order (no ORDER BY on get_all_nodes) ->
ReferenceResolver's name_cache build order, and find_best_match /
resolve_from_filtered_named break ties by picking the first candidate
in that Vec. On unchanged source, three consecutive `sync --force`
runs against a 676-file GDScript project produced cycle membership
[206,2] / [207,2] / [206,2] — different files each time, not just
different counts — because ambiguous same-name calls resolved to a
different tied candidate depending on completion order.

Fix: tag each queued file with its original index, and sort `results`
by that index before returning from `extract_files`, so downstream
node order (and therefore resolver tie-breaks) is stable regardless of
which worker thread finishes first.

## Test plan
- Verified against the real repro: 5 consecutive `sync --force` runs on
  the same project now produce byte-identical cycle membership (was
  flapping every run before this fix).
- Full `cargo test --workspace` (90 suites) and `cargo clippy --lib` pass clean.

### DIFF
--- a/src/extraction_worker.rs
+++ b/src/extraction_worker.rs
@@ -226,8 +226,12 @@ impl WorkerPool {
         F: Fn(usize, usize, &str) + Send + Sync + 'static,
     {
         let total = files.len();
-        let queue: Arc<Mutex<VecDeque<String>>> = Arc::new(Mutex::new(files.into_iter().collect()));
-        let results: Arc<Mutex<Vec<ExtractTuple>>> =
+        // Pair each file with its original position so completion order
+        // (a race between worker threads) never leaks into the returned
+        // node/edge order — see `results` below.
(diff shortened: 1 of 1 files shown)


aovestdipaperino:
Merged — thanks @krishnaglick! Clean, well-analyzed determinism fix: tagging each queued file with its original index and sorting before return stops worker-completion order from leaking into node/resolver tie-break order. The 5-run repro verification sealed it. 🙏

PR #221, opened 2026-07-13
fix(gdscript): capture 3 dynamic-dispatch idioms that read as dead code

krishnaglick:
Prompted by a downstream project's dead-code audit: 4 parallel triage
passes checked 482 raw `tokensave_dead_code` candidates by hand and found
~86% were false positives from extraction gaps in this file, not real
dead code. Fixed 3 concrete gaps, each verified via a parse-tree dump
against the live grammar (not inferred from node-types.json alone, which
under-documents the bare-attribute shape) and a before/after
`tokensave_callers` check against real examples in that downstream repo:

1. `attribute_call`'s `callee_name()` only emitted the bare method name
   (`some_method`), discarding the receiver entirely -- `XScript.
   some_method()` and `some_method()` were indistinguishable to the
   resolver. Now emits `receiver.method`, matching the convention the
   Python/TS/JS extractors already use for the same shape.

2. A function reference passed BY VALUE with no call parens
   (`get_or_create(_h, MyDb._load_from_registry)`, the BaseDatabaseCache
   lazy-init idiom) produced zero extraction output at all -- it's
   neither a `call` nor `attribute_call` node. Added a scan of each
   call's `arguments` node for bare dotted-attribute children (identifier
   pairs with no trailing attribute_call/attribute_subscript), recorded
   as Calls-kind references.

3. Three well-known Godot dynamic-dispatch APIs -- `Callable(receiver,
   "method_name")`, `X.call_deferred("method_name")`, `X.connect(callback)`
   (bare identifier or bare dotted-attribute callback) -- previously
   contributed nothing beyond the ordinary call-to-themselves edge. Added
   narrow, name-matched extraction of the string/bare-reference target
   for each.

## Not fixed by this change
Both need a receiver-type-aware resolver strategy (tracking what a
preload const or local variable is actually an instance of), not just an
extraction-side text-preservation fix:
- Ambiguous resolution when 2+ files define a same-named function: the
  new receiver-qualified text is emitted correctly, but `resolve_one`'s
  bare-name matching can still attribute a reference to the wrong
  same-named candidate.
- `const X = preload("res://...")` alias indirection (`X.some_method()`)
  -- the receiver text is now preserved, but nothing maps the alias name
  back to the file it points at.
- `.has_method("x")`-guarded duck-typed dispatch, and typed-local-variable
  dispatch (`var ctrl: SomeController; ctrl.some_method()`) where the
  variable's declared type would need tracking through the function body.

## Test plan
- Full test suite (90 test binaries, unit + integration + doc tests)
  passes with zero regressions.
- Rebuilt release binary, ran a fresh full sync against a real
  ~500-file GDScript project (62481 -> 74770 edges), confirmed 3
  concrete previously-false-positive examples now show correct callers.

---
**Update:** pushed a follow-up commit (`fix(gdscript): satisfy CI clippy::pedantic and rustfmt`) — doc-comment formatting only (blank lines before trailing paragraphs, backticked two bare identifiers), no logic change. CI is green.

### DIFF
--- a/src/extraction/gdscript_extractor.rs
+++ b/src/extraction/gdscript_extractor.rs
@@ -777,6 +777,7 @@ impl GdScriptExtractor {
     ///   reference passed by value, the shape `BaseDatabaseCache`'s
     ///   lazy-init pattern and similar dispatch tables use. See
     ///   `extract_bare_attribute_args`.
+    ///
     /// Both were previously invisible to this extractor: a callee referenced
     /// only this way had zero recorded edges, making `tokensave_dead_code`
     /// misreport it as unreachable even though it's genuinely live.
@@ -833,10 +834,11 @@ impl GdScriptExtractor {
     /// - `X.connect(callback)` — Godot's signal-connect API; `callback` is a
     ///   bare identifier or bare dotted-attribute function reference (no call
     ///   parens) naming the handler directly.
+    ///
     /// All three were previously invisible: a handler referenced only this
     /// way showed zero incoming edges and misreported as dead code by
     /// `tokensave_dead_code`. Matching is on method name only (not receiver
-    /// type — GDScript has no static typing strong enough to verify the
+    /// type — `GDScript` has no static typing strong enough to verify the
     /// receiver really is e.g. a `Signal`), so this is a heuristic, but a
     /// narrow one: these three names are Godot-API-reserved enough in
     /// practice that a same-named unrelated user method is very unlikely.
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Merged — thanks @krishnaglick! Nicely scoped GDScript extraction fix: receiver-qualified attribute calls plus a deliberately narrow Godot dynamic-dispatch allowlist, all grammar-verified rather than inferred from node-types.json, with thorough tests. 🙏

Issue #204, opened 2026-07-11
Windows: MCP path arguments reject backslash paths despite canonical stored paths


SooYoungJang:
## Summary

Several MCP tools do not normalize native Windows backslash path arguments. Equivalent forward-slash paths work against the same indexed files.

## Reproduction

1. `tokensave_entities({"file":"apps\\admin\\src\\components\\StatCard.tsx"})` returns `symbol_count: 0`.
2. `tokensave_entities({"file":"apps/admin/src/components/StatCard.tsx"})` returns four symbols.
3. `tokensave_files({"format":"flat","path":"apps\\admin\\src\\components"})` returns an empty result.
4. The forward-slash `path` returns the two indexed TSX files.
5. `tokensave_search({"query":"StatCard","path_include":["apps\\admin\\src\\components"],"limit":20})` returns `[]`; the forward-slash include returns the expected symbols.

## Expected

On Windows, tools accepting `file`, `path`, `path_include`, or `path_exclude` should normalize `\` to the canonical stored separator before matching.

## Notes

`tokensave_read` does accept the backslash form, so the issue is limited to particular query/metadata tools rather than all file operations. #87 fixed duplicate stored path variants; this report concerns normalization of incoming MCP arguments.

## Environment

- TokenSave: 7.0.2
- Platform: Windows
- Project type: TypeScript / TSX
- 7.1.0 was reported as available but was not tested; its release notes do not list an explicit fix for this case.
- Repository identity, source bodies, and product-specific details are intentionally omitted.

PR #222, opened 2026-07-13
fix(edit-tools): honor absolute paths verbatim, add project_root override, echo resolved_path

krishnaglick:
The MCP edit tools (str_replace, multi_str_replace, insert_at,
replace_symbol, insert_at_symbol, ast_grep_rewrite) always resolved a
relative `path` against the indexed project root, with no way for a caller
to retarget it. A caller working in a git worktree that passed a relative
path got the edit silently written into the primary checkout instead of
the worktree — caught only by a later `git status`. Absolute paths outside
the indexed root were also rejected outright ("path is not within the
project"), even though an absolute path is itself an explicit, deliberate
statement of where the write should land.

## Fix
- Replace `resolve_path`/`absolute_path` (in the 6 edit methods) with a
  single `resolve_edit_target(path, root_override)` helper: absolute paths
  are honored verbatim (even outside the indexed root — no separate
  opt-out flag needed, since passing an absolute path is itself the
  signal); relative paths resolve against `root_override` when given,
  else the indexed project root (unchanged default). DB reindexing only
  fires when the resolved path actually falls under the indexed root.
- Add an optional `project_root` (alias `cwd`) argument to all 6 edit MCP
  tools, threaded through to the corresponding TokenSave methods, so a
  worktree caller can retarget relative-path resolution per call.
- Add `resolved_path` to EditResult/MultiEditResult/InsertResult/
  AstGrepResult — the fully-resolved absolute path actually read/written —
  so callers can verify the target even when they don't use the override,
  as a cheap guard against silent wrong-tree writes.

## Test plan
- `tests/mcp_handler_test.rs`: resolved_path for the default in-root
  case, absolute-path-outside-root honored for str_replace and insert_at,
  project_root override for str_replace and replace_symbol (the
  DB-resolved symbol-path case), each asserting the primary checkout's
  file is left untouched while the override target is edited.
- 170 mcp_handler_test cases pass (165 existing + 5 new).
- Full `cargo test --workspace` and `cargo clippy --lib` clean.

---
**Update:** pushed a follow-up commit (`fix(mcp): keep edit tools' absolute paths verbatim through arg normalization`) — CI's Windows job caught a real gap: the pre-existing `normalize_path_args` helper (#204) unconditionally rewrote `C:\...` to forward slashes before this PR's "honor absolute paths verbatim" logic ever saw the path. Scoped that normalization's drive-letter exception to just the 6 edit tools (other tools still need forward-slash normalization to match the DB's canonical paths). Also includes the rustfmt fixes CI's Format check flagged. CI is green.

### DIFF
--- a/src/mcp/tools/definitions.rs
+++ b/src/mcp/tools/definitions.rs
@@ -1267,7 +1267,7 @@ fn def_str_replace() -> ToolDefinition {
             "properties": {
                 "path": {
                     "type": "string",
-                    "description": "Absolute or project-relative file path"
+                    "description": "Absolute or project-relative file path. An absolute path is always honored verbatim, even outside the indexed project root (e.g. a git worktree)."
                 },
                 "old_str": {
                     "type": "string",
@@ -1276,6 +1276,10 @@ fn def_str_replace() -> ToolDefinition {
                 "new_str": {
                     "type": "string",
                     "description": "Replacement string"
+                },
+                "project_root": {
+                    "type": "string",
+                    "description": "Optional absolute directory to resolve a relative `path` against instead of the indexed project root. Use this when calling from a git worktree so relative paths land in the worktree, not the primary checkout. Ignored when `path` is absolute. Alias: `cwd`."
                 }
             },
             "required": ["path", "old_str", "new_str"]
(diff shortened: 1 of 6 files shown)


aovestdipaperino:
Merged — thanks @krishnaglick! Rebased onto master (the conflict was an additive test-file collision with the newly-landed #216/#223 tests, resolved by keeping both sets). Nicely designed fix: `resolve_edit_target` cleanly unifies the six edit tools, absolute paths are honored verbatim with `resolved_path` echoed back as a wrong-tree guard, reindex only fires when the target is actually in-root, and the Windows drive-letter carve-out scoped to just the edit tools is the right call. The one Windows CI failure was an unrelated environmental `STATUS_ACCESS_VIOLATION` in go_bug149_test — green on re-run. 🙏

PR #223, opened 2026-07-13
fix(sync): persist unresolved refs from full index so incremental sync stays convergent

krishnaglick:
`index_all_with_progress_verbose()` resolves every cross-file reference
in-memory but never wrote those raw references to the durable
`unresolved_refs` table — only the incremental sync path did that, per
touched file. `sync()`'s incremental path deletes+reinserts a changed
file's nodes via `delete_nodes_by_file()`, which cascades away every edge
touching that file's (about-to-change) node ids, including inbound Calls
edges from other, untouched files that call into it. Those edges only get
recreated if the callers' original references are sitting in
`unresolved_refs` for sync's resolution step to replay — and since a full
index never wrote them there, any inbound edge from an initial
`index_all()` was permanently lost the first time its target file was
later touched by sync.

## Fix
After inserting nodes/edges/files during a full index, also persist the
complete raw unresolved-ref set to the `unresolved_refs` table, mirroring
what the incremental per-file path already does.

## Test plan
Two regression tests added to `tests/integration_test.rs`:
- `test_sync_reresolves_inbound_edges_after_callee_change`
- `test_incremental_sync_edge_count_matches_full_reindex`

Verified RED without the fix (both new tests fail exactly as described:
0/2 edges survive), GREEN with it restored. Full `cargo test --workspace`
passes with zero regressions; `cargo clippy` introduces no new warnings.

### DIFF
--- a/src/tokensave/indexing.rs
+++ b/src/tokensave/indexing.rs
@@ -150,6 +150,24 @@ impl TokenSave {
         self.db.insert_edges(&all_edges).await?;
         self.db.upsert_files(&file_records).await?;
 
+        // Durably record every raw unresolved reference extracted this pass —
+        // not just the leftovers `resolve_all` couldn't bind. A full index
+        // resolves cross-file refs in-memory in one shot and, until now, never
+        // wrote them to the `unresolved_refs` table. That left a later
+        // incremental `sync` with no record of e.g. "file A calls a symbol
+        // defined in file B": when B is edited, `delete_nodes_by_file` cascades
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Merged — thanks @krishnaglick! This is a genuinely valuable convergence fix: persisting the full unresolved-ref set from the full index means incremental sync can re-resolve inbound edges from untouched callers after a callee changes, instead of monotonically losing cross-file call edges. The edge-count parity test against a fresh full reindex is a great regression guard. 🙏

PR #227, opened 2026-07-14
fix(pr_context): resolve default branch from branch-meta

saveriogzz:
## Description

resolve default branch from branch-meta instead of hardcoding "main"

`handle_pr_context` hardcoded "main" as the fallback base ref, causing `git error: cannot resolve 'main'` on repos whose default branch is "master" (or any other name). Now reads `default_branch` from `branch-meta.json`, matching what `handle_branch_diff` already does.

## Summary

- `handle_pr_context` now loads `branch-meta.json` and uses `default_branch` as the fallback base ref instead of hardcoding `"main"`
- Falls back to `"main"` only when no branch metadata exists (pre-branch-tracking projects)
- Added regression test `pr_context_uses_branch_meta_default_branch`

## Motivation

Running `tokensave tool pr_context` (or the MCP tool `tokensave_pr_context`) on a repo whose default branch is `master` fails with:

```
git error: cannot resolve 'main': couldn't parse revision: "main"
```

`handle_branch_diff` already resolves the default branch correctly from `branch-meta.json` — `handle_pr_context` should do the same.

## Changes

- `src/mcp/tools/handlers/git.rs` — `handle_pr_context` loads branch metadata and uses `meta.default_branch` as the fallback for `base_ref`
- `tests/mcp_handler_test.rs` — new test creates a repo with `master` as default branch and calls `pr_context` without `base_ref` to verify it resolves correctly

## Test plan

- [x] `cargo test` passes
- [ ] `cargo clippy` has no new warnings

New regression test `pr_context_uses_branch_meta_default_branch`: creates a git repo with `-b master`, makes a commit, creates a feature branch with a second commit, writes `branch-meta.json` with `default_branch: "master"`, and calls `tokensave_pr_context` with no `base_ref` — asserts no git error and valid diff output.

## Checklist

- [ ] `CHANGELOG.md` updated (under `[Unreleased]` if no version bump)

### DIFF
--- a/src/mcp/tools/handlers/git.rs
+++ b/src/mcp/tools/handlers/git.rs
@@ -812,10 +812,16 @@ pub(super) async fn handle_commit_context(cg: &TokenSave, args: Value) -> Result
 
 /// Handles `tokensave_pr_context` tool calls.
 pub(super) async fn handle_pr_context(cg: &TokenSave, args: Value) -> Result<ToolResult> {
+    let tokensave_dir = crate::config::get_tokensave_dir(cg.project_root());
+    let meta = crate::branch_meta::load_branch_meta(&tokensave_dir);
+    let default_branch = meta
+        .as_ref()
+        .map(|m| m.default_branch.as_str())
+        .unwrap_or("main");
     let base = args
         .get("base_ref")
         .and_then(|v| v.as_str())
-        .unwrap_or("main");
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Thanks @saveriogzz for this contribution! The fix is now merged. 🎉

PR #229, opened 2026-07-14
fix: allow source overrides for test-named paths

### LINKED ISSUE #228: Allow production paths named test to override test-file classification
### tokensave version

7.2.0

### Operating system

macOS (arm64)

### Installation method

GitHub release binary

### What happened?

Production code under a directory named `test` is classified as test-only. The files are indexed and searchable, but test-aware tools such as `tokensave_test_risk` and `tokensave_test_coverage` exclude them from production functions.

The current path check treats any path containing `test/` as a test file. That is ambiguous for applications where `test` is a product feature or route name.

### Steps to reproduce

1. Add a production component at `components/test/Widget.tsx`.
2. Add its unit test at `components/test/__tests__/Widget.test.tsx`.
3. Run `tokensave init`.
4. Call `tokensave_test_coverage` for the production component or `tokensave_test_risk` for `components/test`.

### Actual behavior

The production component reports zero production functions and is counted as test-only.

### Expected behavior

Projects should be able to mark ambiguous paths as production code while keeping explicit test markers such as `__tests__/`, `*.test.*`, and `*.spec.*` classified as tests.

### Workaround

Renaming the directory avoids the heuristic. Passing `path_include` can help search and context queries, but it does not change test/source classification.

A configurable source-path override would avoid forcing projects to rename valid product routes or feature directories.

barbrick:
## Summary

- add a `source_path_overrides` project setting for production paths that match the test-directory heuristic
- use the project-aware classifier across coverage, risk, affected-test, context, workflow, and git tools
- keep explicit markers such as `__tests__/`, `*.test.*`, and `*.spec.*` classified as tests

## Motivation

Some applications use `test` as a product route or feature name. Those files are indexed and searchable, but test-aware tools currently classify them as test-only.

Fixes #228.

## Changes

- added the backward-compatible config field and centralized project-aware classification
- removed the extra hard-coded `/test` exclusion in test-risk analysis
- added unit and MCP integration regression tests
- documented the setting and updated the changelog

## Test plan

- [x] `cargo test` — 2,377 passed across 95 suites
- [x] `cargo clippy --workspace --all-targets` — clean
- [x] `cargo fmt --check` — clean
- [x] synthetic production and unit-test paths covered by integration tests

## Checklist

- [x] No breaking changes


aovestdipaperino:
Thanks, @barbrick! 🙏 Really solid refactor — explicit test markers taking precedence over the overrides is exactly the right ordering, and the end-to-end coverage/risk test makes the behavior unambiguous. The `/test` substring removal in `test_risk` is a nice latent-bug fix on top. Merging.

Issue #211, opened 2026-07-11
Vitest describe/it callbacks are not indexed, leaving directly tested TS functions uncovered


SooYoungJang:
## Summary

A normal TypeScript `*.test.ts` file is indexed only as a file plus imports; its `describe` / `it` callback tests and direct calls to production functions are absent. This breaks `test_coverage`, `test_map`, and `affected`.

## Reproduction

A colocated `date.test.ts` directly imports and tests seven functions from `date.ts`, using 8 `describe` suites and 36 `it` callbacks.

`tokensave_entities({"file":"packages/shared/src/utils/date.test.ts"})` returns only:

- the file node
- the `vitest` import node
- the `./date` import node

It returns zero test-function/callback nodes.

`tokensave_test_coverage({"file":"packages/shared/src/utils/date.ts","include_untested":true,"max_depth":5})` reports `tested: 0`, `untested: 7`.

`tokensave_test_map({"file":"packages/shared/src/utils/date.ts"})` reports no test files and zero covered symbols. Separate TS/TSX samples also cause `tokensave_affected` to miss directly importing colocated tests.

## Expected

- `describe` / `it` callbacks in `*.test.ts` should be recognized as tests.
- Their direct calls/import dependencies should connect to production symbols.
- Coverage, test map, and affected-test tools should surface the colocated tests.

This is a follow-up to #103, whose acceptance criteria explicitly include TypeScript `*.test.ts` conventions.

## Environment

- TokenSave: 7.0.2
- Platform: Windows
- Project type: TypeScript / TSX
- 7.1.0 was reported as available but was not tested; its release notes do not list an explicit fix for this case.
- Repository identity, source bodies, and product-specific details are intentionally omitted.

aovestdipaperino:
Fixed in a9c6f3a. Top-level `describe`/`it`/`test`/`suite` calls (including dotted forms like `it.each`) now become Function nodes named after the suite/case title; nested suites attach beneath their parent, and calls inside the callbacks emit Calls refs. Combined with the existing `.test.ts` path detection this restores `test_coverage`, `test_map`, and `affected` for colocated Vitest/Jest tests. Will ship in the next release.

PR #230, opened 2026-07-15
fix: index .mjs/.cjs/.mts/.cts as JavaScript/TypeScript (#219)

### LINKED ISSUE #219: *.mjs files are skipped by indexing despite being classified as JavaScript
## tokensave version

7.2.0

## Operating system

macOS 26.5 (25F71)

## Installation method

Homebrew (`brew install tokensave`)

## What happened?

`*.mjs` files are silently omitted from the index. After a successful `tokensave sync`, regular `*.js` files appear in `tokensave_files` and the SQLite `files` table, while colocated Node.js tests named `*.test.mjs` do not appear at all. As a result, `tokensave_test_coverage`, `tokensave_test_map`, `tokensave_affected`, and related tools cannot discover those tests.

In the observed project, the source tree contained 17 `*.js` files and 8 `*.mjs` test files. The database contained 17 `*.js` rows and 0 `*.mjs` rows. Repository identity, source bodies, and product-specific details are intentionally omitted.

The current public source appears to have an extension-registration mismatch: `.mjs`/`.cjs` are recognized by the hook file-extension list, formatter, resolver, and query display mapping, but `TypeScriptExtractor::extensions()` registers only `ts`, `tsx`, `js`, and `jsx`.

## Steps to reproduce

1. Create a minimal JavaScript project:

   `src/add.js`

   ```js
   export const add = (a, b) => a + b;
   ```

   `test/add.test.mjs`

   ```js
   import assert from 'node:assert/strict';
   import test from 'node:test';
   import { add } from '../src/add.js';

   test('adds two numbers', () => {
     assert.equal(add(1, 2), 3);
   });
   ```

2. Run `tokensave sync`.
3. Run `tokensave files` or query `.tokensave/tokensave.db`:

   ```sql
   SELECT path FROM files WHERE path LIKE '%.mjs';
   ```

4. Observe that `src/add.js` is indexed but `test/add.test.mjs` is absent.

## Expected behavior

- `.mjs` files should be indexed as JavaScript files.
- `node:test` callbacks and calls inside `*.test.mjs` should be available to test coverage, test map, and affected-test tools.
- A supported extension should not be silently discarded during sync.

## Error output

```shell
$ tokensave sync
sync done

$ sqlite3 .tokensave/tokensave.db \
  "SELECT count(*) FROM files WHERE path LIKE '%.mjs';"
0
```

No warning or extraction error is emitted.

## Component

Indexing / sync

## Additional context

This differs from #211: in that issue, the TypeScript test file was present in the graph but its `describe`/`it` callbacks were missing. Here the `.mjs` file itself is absent from the indexed file set.

It may also be worth checking `.cjs`, because several non-extractor components already classify both `.mjs` and `.cjs` as JavaScript, but only `.mjs` was directly reproduced here.

janstol:
## Summary

- Register `mjs`/`cjs`/`mts`/`cts` in `TypeScriptExtractor::extensions()` so the sync walker no longer silently skips them.
- Route `.mjs`/`.cjs` through the JavaScript tree-sitter grammar (`.mts`/`.cts` already fell through to the TypeScript default correctly).
- Add fixture + extraction tests covering `.mjs`, `.cjs`, and registry lookups for all four extensions.

## Motivation

Closes #219.

`TypeScriptExtractor::extensions()` only registered `ts`, `tsx`, `js`, `jsx`, while the file watcher (`hooks.rs`), context formatter, resolver, and DB language-display mapping already treated `.mjs`/`.cjs`/`.mts`/`.cts` as JavaScript/TypeScript. Since `LanguageRegistry::extractor_for_file` looks up the extractor purely by extension, any file with one of those extensions matched no extractor and was silently dropped from the walk during `tokensave sync` — with no warning. This was most visible for colocated Node.js test files like `*.test.mjs`, which made `tokensave_test_coverage`, `tokensave_test_map`, and `tokensave_affected` unable to discover those tests at all.

## Changes

- `src/extraction/typescript_extractor.rs`: extended `extensions()` to `["ts", "tsx", "js", "jsx", "mjs", "cjs", "mts", "cts"]`; extended the JS arm of `parse_source()`'s grammar-selection match to include `mjs`/`cjs`; updated the module doc comment.
- `tests/fixtures/sample.mjs`: new fixture — ESM source with a `node:test` callback.
- `tests/fixture_extraction_test.rs`: `test_fixture_javascript_mjs`.
- `tests/typescript_extraction_test.rs`: `test_mjs_node_test_callback_indexed` (mirrors the vitest describe/it test from #211, using `node:test` per the issue's repro), `test_cjs_file_parses_as_javascript`.
- `tests/extraction_test.rs`: extension-registry assertions and `extractor_for_file` lookups for all four extensions.
- `CHANGELOG.md`: `[Unreleased] → Fixed` entry.

## Test plan

- [x] `cargo test` passes (full workspace, 0 failures)
- [x] Tested manually: reproduced the issue's exact repro (`src/add.js` + `test/add.test.mjs`, `tokensave sync`, `SELECT count(*) FROM files WHERE path LIKE '%.mjs'`) against a release build — went from `0` rows (as reported) to `1`, and the `test('adds two numbers', ...)` `node:test` callback appears as a `function` node in the graph.

## Checklist

- [x] No breaking changes


aovestdipaperino:
Thanks for the clean fix, @janstol! 🙏 Nicely scoped — the extractor registry was indeed the only gap, and the `node:test` fixture coverage for colocated `*.test.mjs` files is exactly right. Merging.

PR #231, opened 2026-07-15
Fix false MCP permission warning; add compact grant option

janstol:
## Summary

- Fix a false "N tools not permitted" warning from `tokensave sync`/`doctor` when Claude Code's `permissions.allow` already covers every tool via a wildcard or bare server grant.
- Add an opt-in compact permission grant for Claude Code (`--wildcard-permissions`/`--explicit-permissions`, persisted via `wildcard_permissions` in `~/.tokensave/config.toml`) so `install`/`reinstall` can write a single `mcp__tokensave__*` entry instead of enumerating 80+ individual tool permissions.

## Motivation

A user's `~/.claude/settings.json` granted tokensave tools with a single `mcp__tokensave__*` wildcard entry, but `tokensave sync` kept printing:

> warning: 81 new tokensave tool(s) not yet permitted. Run `tokensave reinstall` to update permissions.

The self-checks compared the fully expanded 80+ tool list against `permissions.allow` using literal string equality, so a wildcard or bare server-wide grant — both of which Claude Code fully honors as `allow` rules per [its docs](https://code.claude.com/docs/en/permissions#tool-name-wildcards) — matched nothing and was reported as missing on every run, even though the tools were actually permitted at runtime.

While fixing the recognition side, it made sense to also let users (and `tokensave install`) write that compact form directly, instead of only recognizing one a user typed by hand.

## Changes

- **`src/agents/claude.rs`** — `perm_is_covered()` recognizes an exact tool grant, the bare `mcp__tokensave` grant, the `mcp__tokensave__*` wildcard, or an anchored partial glob (`mcp__tokensave__prefix*`) as covering an expected tool permission. An *unanchored* glob like `mcp__*` is deliberately **not** treated as covering, since Claude Code itself skips unanchored globs in `allow` rules — treating it as covering would hide a real "tools not granted" state. `warn_missing_permissions` and `doctor_check_permissions` use this for their missing/stale checks. `install_permissions` now prunes any previously-installed tokensave-owned entries before writing, so switching between the explicit list and the compact wildcard never leaves a mix of both styles; when the caller hasn't explicitly requested a style this run (`InstallContext::force_permission_style == false`, e.g. a silent reinstall on upgrade), an existing single grant that already covers every tool is left untouched instead of being pruned back to the 80+ entry list.
- **`src/agents/mod.rs`** — `TOKENSAVE_WILDCARD_PERM` constant and `install_tool_perms(wildcard: bool)` helper; `InstallContext` gains `force_permission_style`.
- **`src/user_config.rs`** — new `wildcard_permissions: bool` field (`#[serde(default)]`, defaults to `false`) so the chosen style persists across silent reinstalls on upgrade.
- **`src/cli.rs` / `src/main.rs`** — mutually-exclusive `--wildcard-permissions`/`--explicit-permissions` flags on `install` and `reinstall`; wiring through to `install_tool_perms`/`force_permission_style` at every call site.
- **`docs/USER-GUIDE.md`** — documents the new config field and CLI flags.
- **`CHANGELOG.md`** — `[Unreleased]` entries under `### Fixed` and `### Added`.
- Tests added/updated in `src/agents/claude.rs` (unit) and `tests/claude_agent_test.rs`, `tests/user_config_test.rs`, and the various agent-integration test helpers (integration), covering recognition of each covering-grant form, the negative case for unanchored `mcp__*`, wildcard install, and style-switch pruning.

## Test plan

- [x] `cargo test` passes
- [x] Tested manually: pointed `HOME`/`~/.claude/settings.json` at a `mcp__tokensave__*`-only grant, confirmed `tokensave sync`/`doctor` no longer warn; ran `tokensave reinstall --wildcard-permissions` and `--explicit-permissions` and confirmed `settings.json` switches cleanly between the compact and explicit forms with no leftover entries.

## Checklist

- [x] Breaking changes documented (if any) — none; the fix is unconditional and backward compatible, the compact grant is opt-in and defaults to the existing explicit-list behavior


aovestdipaperino:
Merged — thank you @janstol for the careful handling of the covering-grant semantics (especially deliberately excluding unanchored `mcp__*`) and the thorough test coverage across install/reinstall style switching. Much appreciated!

PR #232, opened 2026-07-15
fix: record Python nested-def calls and first-class value refs (#224)

### LINKED ISSUE #224: Python extractor misses several reference kinds, inflating dead-code / unused-import results
### tokensave version

7.0.3

### Operating system

macOS

### Installation method

Homebrew (brew)

### What happened?

# Python extractor misses several reference kinds, inflating dead-code / unused-import results

## Summary

On a ~280-file Python project (tokensave v7.0.3), `tokensave_dead_code` reported 320 dead
symbols and `tokensave_unused_imports` reported 1,465 unused imports. Spot-checking showed
three whole classes of false positives, each caused by a reference kind the Python extractor
does not record as an edge:

1. **Imports used only as base classes** are flagged unused — `class Foo(StrEnum)` creates
   no `uses` edge to the `from enum import StrEnum` import node.
2. **Functions referenced by name (not called)** are flagged dead — e.g. passed as a keyword
   argument into a spec/callback table (`QuestionSpec(parse=_parse_text)`). First-class
   function references produce no edge.
3. **Some plain same-file calls** are missed — a module-level helper called twice from a
   sibling function in the same file still showed "no incoming edges". Both call sites are
   ordinary `return _helper(...)` statements inside a nested closure of a factory function,
   so the miss may be closure-related.
4. **Some cross-file calls through a plain `from x import y` are missed** — a function
   imported with `from pkg.mod import fn` and called once inside a method of a class in the
   importing file was still reported dead by `tokensave_dead_code`.

The result is that the dead-code and unused-import tools are unusable for cleanup sprints on
Python codebases that use these (very common) patterns: the true-positive rate in our sample
was low enough that every candidate needed manual grep verification anyway.

## Detail

Environment:

- tokensave v7.0.3 (MCP server), macOS (darwin 25.5), Python project, ~7,600 nodes / ~7,700
  edges, `edges_by_kind: {uses: 1648, annotates: 250, extends: 56, calls: 5771}`

Observed vs expected:

| Pattern | Tool | Observed | Expected |
|---|---|---|---|
| `class Foo(StrEnum)` | `tokensave_unused_imports` | `enum.StrEnum` listed as unused | `extends`/`uses` edge from class to import |
| `spec = Spec(parse=_my_parser)` | `tokensave_dead_code` | `_my_parser` listed as dead | `uses` edge for first-class function reference |
| same-file call inside closure | `tokensave_dead_code` | helper listed as dead | `calls` edge |

Notes:

- The `extends` edge kind exists in the schema (56 present in our graph), so base classes of
  *project* classes may be handled; the miss appears specific to bases imported from the
  stdlib/third-party (import node → no edge).
- The unused-import list is additionally dominated by `__init__.py` re-exports (intentional
  public API surface in Python). A built-in heuristic (treat imports in `__init__.py`, or
  names listed in `__all__`, as used) would remove most of that noise.

### Steps to reproduce

Create a fresh project containing one file `repro.py`:

```python
from enum import StrEnum


class Colour(StrEnum):          # (1) base-class use of an import
    RED = "red"


def _parse_text(text: str) -> str:
    return text.strip()


PARSERS = {"text": _parse_text}  # (2) first-class function reference


def _helper(x: int) -> int:
    return x + 1


def make_adder():
    def add(x: int) -> int:
        return _helper(x)        # (3) same-file call inside a closure
    return add
```

Then, from the fresh project directory:

1. Run `tokensave init .` to index the project.
2. Run `tokensave tool unused_imports` → **expected:** empty; **actual:**
   `enum.StrEnum` is reported unused.
3. Run `tokensave tool dead_code --include-public true` → **expected:** only `make_adder`,
   which is genuinely unreferenced in this toy repo; **actual:** `_parse_text`, `_helper`,
   and `make_adder` are reported. This confirms that the ordinary `_helper(x)` call inside
   the nested `add` closure creates no incoming `calls` edge for `_helper`.

Confirmed with tokensave v7.0.3 on macOS: the one-file repro indexes 7 nodes and 0 edges, so
all three missing-reference patterns reproduce without the decorator-registered factory
from the larger project.

### Error output

```shell
See the observations
```

### Component

MCP server

### Additional context

_No response_

janstol:
## Summary

- Fix Python extractor/handler misses that produced false `dead_code`/`unused_imports` results
- Derive Python import identifiers from actual import semantics instead of a generic `::`-based parser
- Emit `Uses` references for first-class function references (dict values, keyword args, parameter defaults, class-level attributes) and recurse into nested function/class definitions

## Motivation

Closes #224. On Python codebases, `tokensave_dead_code` and `tokensave_unused_imports` reported large numbers of false positives, driven by reference patterns the Python extractor and unused-imports handler didn't record:

1. `class Colour(StrEnum)` — a base-class-only import was reported unused because the unused-imports scan fed the Use node's dotted name through a `::`-based (Rust-shaped) parser that never matches Python import syntax.
2. `PARSERS = {"text": _parse_text}`, `Spec(parse=_fn)`, `def invoke(callback=_default_callback)`, `class Registry: CALLBACKS = {"x": _fn}` — a function referenced only by name (never called) produced no graph edge and was reported dead.
3. `visit_function` never recursed into nested `def`s, so a closure's own calls — and the closure itself — were invisible to the graph, e.g. a factory's nested closure calling a same-file helper left the helper looking dead.
4. A reported cross-file `from pkg.mod import fn` miss was re-verified as a symptom of (3); no separate fix was needed once same-file closures resolve correctly.

## Changes

- `src/mcp/tools/handlers/analysis.rs` — new `python_import_identifier` helper derives the actually-bound name from Python import statement semantics (`from a.b import c` binds `c`; `import a.b.c` binds `a`; an `as` alias always wins, with alias search restricted to the region after the `import` keyword so a module name colliding with the imported name — `from datetime import datetime as dt` — doesn't shadow the alias). `__init__.py` re-exports are excluded from `unused_imports` entirely.
- `src/extraction/python_extractor.rs` — bounded `Uses`-ref scan (`scan_value_positions`) now covers call-argument values, assignment/`return` right-hand sides, and parameter defaults; class-body assignments reach the same RHS scan as module-level ones without becoming module-level `Const` nodes. `visit_function` now recurses into nested function/class definitions (including inside `if`/`try`/`with` blocks), attributing them as children with `class_depth` reset so a closure nested in a method is still classified as a `Function`, not a `Method`.
- `tests/python_extraction_test.rs` — extractor-level unit tests for each new reference kind.
- `tests/fixture_extraction_test.rs` / `tests/fixtures/sample.py` — fixture-based assertions per the extraction-change fixture convention.
- `tests/python_bug224_test.rs` — end-to-end integration tests asserting actual `tokensave_dead_code`/`tokensave_unused_imports` output via `TokenSave::init` + `handle_tool_call`.
- `CHANGELOG.md` — `[Unreleased]` entry.

## Test plan

- [x] `cargo test` passes (full suite, default features)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` has no new warnings
- [x] Tested manually: built scratch Python projects for the issue's exact repro, a cross-file `from pkg.mod import fn` call, a `from datetime import datetime as dt` alias-collision case, and a parameter-default/class-level-callback case — `tokensave tool dead_code`/`tokensave tool unused_imports` match expected output in each

## Checklist


aovestdipaperino:
Merged — thank you @janstol for the thorough root-cause work here, especially re-verifying that the cross-file miss was a symptom of the nested-def gap rather than patching it separately, and for the end-to-end tests against actual `dead_code`/`unused_imports` output. Great contribution!

Issue #236, opened 2026-07-16
The Droid `PreToolUse` hook rewrites all agent configs on every tool call


rNoz:
### tokensave version

7.4.0

### Operating system

macOS

### Installation method

Built from source

### What happened?

`should_skip_agent_install_maintenance` (src/main.rs) lists the per-tool-call hook subcommands that must bypass the pre-dispatch startup work, but `HookDroidPreToolUse` is missing while every other hook is present (Claude: `HookPreToolUse`, `HookPromptSubmit`, `HookStop`; Kiro: `HookKiroPreToolUse`, `HookKiroPromptSubmit`, `HookKiroPostToolUse`).
So the Droid `PreToolUse` hook alone takes the normal `run()` startup path on every tool call: a worldwide-counter network flush, `user_config.save()`, `check_install_stale`, and the silent-reinstall loop that calls `ag.install(...)` for every tracked agent when `needs_reinstall` is true (a pending version transition), rewriting each agent's MCP/hook config and leaving `.bak` files.
The function's own doc-comment states the per-tool-call hot path must not run this, so the omission looks like an oversight rather than intent.

### Steps to reproduce

1. Install the Droid `PreToolUse` hook (`tokensave install --agent droid`) in a project with a `.tokensave/` index.
2. Put the install into a pending version transition (a binary whose version differs from the recorded install markers), so `needs_reinstall` is true.
3. Trigger any Droid tool call that fires `hook-droid-pre-tool-use` (for example a `Grep` or `Execute`) and watch the tracked agent config files get rewritten and `.bak` files appear on each call, plus a network flush every call.

### Error output

```shell
# every hook-droid-pre-tool-use invocation rewrites the tracked agent configs and drops .bak files,
# and performs a network flush + user_config.save(), instead of returning the fast JSON-only decision
```

### Component

None

### Additional context

Fix: add `Commands::HookDroidPreToolUse` to the `matches!` list in `should_skip_agent_install_maintenance` so the Droid hook takes the same fast, side-effect-free path as the Claude and Kiro hooks.
On a stable install the reinstall loop fires once per genuine upgrade; on a build with stuck version markers it fires on every call, but either way the network flush and `user_config.save()` run on every call.

aovestdipaperino:
Fixed by #237 (merged, commit 5a5e3e8) — the Droid PreToolUse hook now takes the same fast, side-effect-free path as the Claude and Kiro hooks. Will ship in the next release. Thanks again @rNoz!

PR #237, opened 2026-07-16
fix(Droid): skip install maintenance on the Droid PreToolUse hook

rNoz:
## Summary

- Add `HookDroidPreToolUse` to `should_skip_agent_install_maintenance` so the Droid `PreToolUse` hook stops running the pre-dispatch install-maintenance work on every tool call.
- Brings the Droid hook to parity with the Claude and Kiro hooks, which already take the fast, side-effect-free path.
- One-variant change with no behavior change for any other command.
- Solves https://github.com/aovestdipaperino/tokensave/issues/236

## Motivation

`should_skip_agent_install_maintenance` (src/main.rs) enumerates the per-tool-call hook subcommands that must bypass the pre-dispatch startup work: the worldwide-counter `try_flush` network round-trip, `check_install_stale`, and the silent-reinstall loop that calls `ag.install(...)` for every tracked agent.

It already lists the Claude hooks (`HookPreToolUse`, `HookPromptSubmit`, `HookStop`) and the Kiro hooks (`HookKiroPreToolUse`, `HookKiroPromptSubmit`, `HookKiroPostToolUse`), but `HookDroidPreToolUse` was missing, so the Droid hook took the normal `run()` startup path instead.

As a result the Droid hook always did the network flush plus `user_config.save()`, and whenever `needs_reinstall` was true (a pending version transition) it rewrote every tracked agent's MCP/hook config and left `.bak` files, all in the middle of an unrelated tool call. The function's own doc-comment says the per-tool-call hot path must not run this, so the omission is an oversight.

## Changes

- `src/main.rs`: add `Commands::HookDroidPreToolUse` to the `matches!` list in `should_skip_agent_install_maintenance`, and extend the neighbouring comment to mention the Droid hook.
- `src/main.rs` (tests): add `per_tool_call_hooks_skip_agent_install_maintenance` asserting every per-tool-call hook variant, including `HookDroidPreToolUse`, is skipped.

## Test plan

- [x] `cargo test` passes (`startup_tests`, 5 tests, including the new case).
- [x] `cargo clippy` has no new warnings (the pre-existing `find_descendant_by_kind` warning under `lite` is unrelated).
- [x] Tested manually: `cargo fmt --check` clean; the new test fails without the fix and passes with it.

## Checklist

- [x] Breaking changes documented (none; additive skip-list entry)


aovestdipaperino:
Thank you @rNoz for catching this — the Droid hook now takes the same fast, side-effect-free path as the Claude and Kiro hooks. Great writeup and regression test!

PR #238, opened 2026-07-16
fix: bash grep guardrail sees through env-assignment and cd prefixes

rNoz:
## Summary

- Strip a leading `NAME=value` env assignment and a leading `cd ... &&`/`cd ... ;` prefix before the grep/rg/ag tool check, so a direct code search hidden behind them is redirected instead of silently passing.
- Honor an explicit inline `TOKENSAVE_DISABLE_GREP_HOOK=<truthy>` as a deliberate opt-out (like the exported variable), rather than stripping it as ordinary noise and then blocking.
- Keep the two intentional behaviors: a pipeline (`... | grep`) still passes through, and the exported-env opt-out is unchanged.
- Additive parsing only; no change to which patterns or targets count as a code search.
- The idea mentioned https://github.com/aovestdipaperino/tokensave/issues/195#issuecomment-4996830687

## Motivation

`extract_grep_invocation` recognized a command as a code search only when it literally started with `grep `/`rg `/`ag ` after stripping `rtk`/`sudo`/`time`/`nice`. So `FOO=bar grep -n Sym src/main.rs` and `cd src && grep -n Sym main.rs` were treated as non-grep and bypassed the redirect, even though both are direct code searches on an indexed project.

This is the raw-shell sub-thread of #195: the native `Grep` tool has no such gap, only the Bash branch does. The fix stays inside the guardrail's deliberate narrow, fail-open scope by only unwrapping shapes that are unambiguously a leading wrapper, not by widening what counts as a search.

## Changes

- `src/hooks.rs`: `strip_command_prefixes` now returns a small `StrippedCommand { rest, disables_hook }` and is applied in a loop so combinations unwrap (for example `cd src && FOO=1 grep ...`). It strips the existing `rtk`/`sudo`/`time`/`nice` wrappers, a leading `NAME=value` assignment (`parse_leading_env_assignment`, quote-aware, only when a following command exists), and a leading `cd ... &&`/`cd ... ;` prefix (`strip_leading_cd`, which returns nothing when the first top-level separator is a pipe, so `cd x && ls | grep` still passes).
- `src/hooks.rs`: while stripping, each leading `TOKENSAVE_DISABLE_GREP_HOOK` assignment re-evaluates `disables_hook` from its value (so the shell's "last assignment wins" holds and the opt-out is recognized wherever it sits in the leading noise, not only first); `evaluate_bash_command` short-circuits on it and passes the command through, so a conscious inline opt-out is honored. Truthiness reuses `HookEnv::from_runtime`'s rule (set, non-empty, not `0`, not `false` case-insensitive) via a shared `disable_value_is_truthy` helper, and `unquote` strips one layer of surrounding quotes first.
- `tests/hooks_test.rs`: add cases for the env-prefix block, the cd-prefix block, the inline opt-out being honored (first, after `cd`/`sudo`, interleaved with other assignments, quoted, and `true`), falsey/empty values still blocking, last-assignment-wins in both directions, nested `cd + env`, and pipe-after-cd passing through.
- `CHANGELOG.md`: add a `### Fixed` entry under `[Unreleased]`.

## Test plan

- [x] `cargo test --test hooks_test` passes (93 on this branch, including the new env-prefix, cd-prefix, inline-opt-out, quoting, truthiness, nested, and last-assignment-wins cases; existing `test_bash_allows_grep_in_pipe_after_other_cmd` still passes).
- [x] `cargo clippy --tests` has no new warnings.
- [x] `cargo fmt --check` clean.

## Checklist

- [x] No breaking change: the opt-out is preserved for both the exported variable and the inline form; only the previously-silent env-prefix and cd-prefix bypasses now redirect.

## Scope and layering (why this could ship here, or be left to rtk)

This PR is offered to give #195 a concrete artifact to react to (specifically https://github.com/aovestdipaperino/tokensave/issues/195#issuecomment-4996830687). It is not a request to merge: whether raw-shell prefix handling belongs in the hook at all, or in a command wrapper like rtk, is the owner/maintainer's call. See the raw-shell grep coverage comment on <https://github.com/aovestdipaperino/tokensave/issues/195>.

- Why it is reasonable to ship here: tokensave already intercepts raw-shell `grep`/`rg`/`ag` by design (the Bash matcher and `evaluate_bash_command` are existing behavior, not new). This only closes an obvious gap in that existing feature (two leading shapes) without changing what counts as a search, and it preserves the narrow, fail-open contract, the pipe pass-through, and the opt-out.
- Why it is reasonable to leave to rtk: going further into shell prefixes trends toward a real shell parser (next would be `;` chains, subshells, `$(...)`, aliases, functions), which the guardrail deliberately avoids. Raw-shell interception may be better owned by a command wrapper that already sits in front of shell commands, leaving the hook focused on the harness tool surface (Grep, Read, Agent). The inline-opt-out handling here is a small reminder that fully correct parsing of arbitrary commands is a bigger commitment than a hook should take on.


rNoz:
CI/Test macOS is a flaky test totally unrelated to this PR.

aovestdipaperino:
Merged — thanks @rNoz! I rebased onto master myself (only CHANGELOG conflicted after #252; I moved your entry under `[Unreleased]`). Nice touch honoring the inline `TOKENSAVE_DISABLE_GREP_HOOK` with last-assignment-wins rather than stripping it as noise.

Issue #239, opened 2026-07-16
Database open flakes on a transient libsql `SQLITE_MISUSE` under concurrent test load (macOS)


rNoz:
### tokensave version

7.4.0

### Operating system

macOS

### Installation method

Built from source

### What happened?

`Database::initialize` and `Database::open` (src/db/connection.rs) open a fresh libsql local connection, apply pragmas, and create or migrate the schema in a single shot with no retry.

Under heavy concurrency the libsql local driver intermittently fails while opening a connection, surfaced as `SQLITE_MISUSE` ("bad parameter or other API misuse"). This shows up on the macOS CI runner, where `cargo test --workspace` runs tests in parallel and each test opens its own temporary database, and it flakes roughly one test per run. Linux CI never reproduces it, and the same tests pass locally on macOS, so it is a concurrency/timing flake in the connect path rather than a logic error in the failing test.

Found in this other PR's CI: https://github.com/aovestdipaperino/tokensave/actions/runs/29540212797/job/87760662378?pr=238

### Steps to reproduce

1. Run `cargo test --workspace` on macOS (a `macos-14` GitHub Actions runner reproduces it most reliably).
2. Observe an occasional single-test failure such as `port_status_does_not_match_methods_of_different_parents`, panicking at the `TokenSave::init(...).unwrap()` call.
3. Re-run: a different test may fail, or all pass, which is the signature of a connect-path flake rather than a deterministic bug.

### Error output

```shell
thread 'port_status_does_not_match_methods_of_different_parents' panicked at tests/mcp_handler_test.rs:990:45:
called `Result::unwrap()` on an `Err` value: Database { message: "failed to create schema: SQLite failure: `bad parameter or other API misuse`", operation: "create_schema" }
test result: FAILED. 173 passed; 1 failed; 0 ignored
```

### Component

None

### Additional context

The connection-setup steps are idempotent (a fresh build/connect, `PRAGMA`s, and `CREATE ... IF NOT EXISTS`), so retrying the whole sequence on a transient error is safe.

Fix: retry `initialize`/`open` up to five times with a short exponential backoff (20, 40, 80, 160 ms) when the error looks transient (`misuse`, `database is locked`, `database is busy`). Genuine schema or corruption errors do not match the transient check and still surface immediately, and a normal run never retries so there is no added latency.

aovestdipaperino:
[...]

PR #240, opened 2026-07-16
fix: retry transient libsql connection-setup failures in db init/open

rNoz:
## Summary

- `Database::initialize` and `Database::open` now retry the whole connection-setup sequence up to five times with a short exponential backoff when the error looks transient.
- Genuine schema or corruption errors do not match the transient check and still surface immediately; a normal run never retries, so there is no added latency.
- It should solves CI blocking like this https://github.com/aovestdipaperino/tokensave/actions/runs/29540212797/job/87760662378?pr=238
- Solves https://github.com/aovestdipaperino/tokensave/issues/239

## Motivation

Both entry points opened a fresh libsql local connection, applied pragmas, and created or migrated the schema in one shot with no retry. Under heavy concurrency the libsql local driver intermittently fails while opening a connection, surfaced as `SQLITE_MISUSE` ("bad parameter or other API misuse"). This reproduced on the macOS CI runner, where `cargo test --workspace` runs tests in parallel and each test opens its own temporary database, and it flaked roughly one test per run. Linux never reproduced it. The setup steps are idempotent (a fresh build/connect, `PRAGMA`s, and `CREATE ... IF NOT EXISTS`), so retrying on a transient error is safe.

## Changes

- `src/db/connection.rs`: split `initialize`/`open` into a thin retry loop plus `try_initialize`/`try_open` holding the existing body; add `DB_CONNECT_MAX_ATTEMPTS` (5), `is_transient_connect_error` (matches a `TokenSaveError::Database` whose message contains `misuse`, `database is locked`, or `database is busy`), and `connect_retry_backoff` (20, 40, 80, 160 ms). The directory creation in `initialize` stays outside the retry loop so it runs once.
- `src/db/connection.rs` tests: add unit tests for the transient-error classifier (matches the observed `SQLITE_MISUSE` and lock/busy messages, ignores a genuine "file is not a database" error and non-database error variants) and for the backoff schedule.
- `CHANGELOG.md`: add a `### Fixed` entry under `[Unreleased]`.

## Test plan

- [x] `cargo test --lib db::connection` passes (8, including the 3 new cases).
- [x] `cargo test --lib` passes (520) and `cargo test --test mcp_handler_test` passes (174), exercising `initialize`/`open` throughout.
- [x] `cargo clippy --lib` has no new warnings.
- [x] `cargo fmt --check` clean.

## Checklist

- [x] No breaking change: behavior is unchanged on the success path and for non-transient errors; only a transient connect failure is retried before surfacing.


aovestdipaperino:
Thanks @rNoz — merged. The idempotent retry with a conservative transient-error classifier and short backoff is exactly the right shape here, and the classifier/backoff unit tests make the intent clear. This should put the macOS CI flake to rest. 🙏

PR #243, opened 2026-07-18
fix: index source files with invalid UTF-8 bytes

### LINKED ISSUE #241: non-utf8 bytes in source files lead to incorrectly skipping that file
### tokensave version

7.4.0

### Operating system

macOS

### Installation method

cargo install

### What happened?

Running `tokensave sync` shows a number of errors/warnings regarding one-byte utf-8 sequences. e.g. in  freetype-2.9.1/builds/amiga/include/config/ftconfig.h

Since I just started using tokensave, I'm unsure whether this represents a failure (this file will be ignored) or a warning ("doesn't matter"). Either way, IMO this shouldn't happen.

### Steps to reproduce

During indexing, many valid C/C++/header files are rejected with warnings like:
! .../SDL_RLEaccel.c: invalid utf-8 sequence of 1 bytes from index 1100

Cause: The scanner reads file contents as bytes and decodes them with strict UTF-8 (std::str::from_utf8 / String::from_utf8). The message is the literal Display of std::str::Utf8Error. Legacy C/C++ sources frequently store author names / symbols in ISO-8859-1 or CP437 comments — e.g. ü (0xFC), å (0xE5), µ (0xB5), è (0xE8), ° (0xB0) — which are lone high bytes that aren't valid UTF-8, so those files are dropped from the graph.

Minimal repro (any OS):
```
printf '/* by W\xfcrkner */\n' > latin1_repro.c   # 0xFC = 'ü' in Latin-1 at index 7
```

# then run tokensave indexing over the directory containing this file

The exact code path it hits (standalone Rust):
```
    match std::str::from_utf8(&bytes) {
        Ok(_) => println!("ok"),
        Err(error) => println!("{error}"),
        // prints: invalid utf-8 sequence of 1 bytes from index 7
    }
```

Suggested fix: decode with `String::from_utf8_lossy` (or an encoding-detection crate such as encoding_rs / chardetng) when strict UTF-8 fails, so these files are indexed instead of skipped. Since the offending bytes are almost always in comments, lossy replacement of 0xFC→� has no effect on parsed symbols.

### Error output

```shell
freetype-2.9.1/builds/amiga/include/c
  onfig/ftconfig.h: invalid utf-8 sequence of 1 bytes from index 577
```

### Component

Indexing / sync

### Additional context

_No response_

janstol:
## Summary

- Index source files containing invalid non-BOM UTF-8 instead of skipping them.
- Preserve existing UTF-8, UTF-8 BOM, and BOM-marked UTF-16 handling.

## Motivation

Legacy C and C++ sources commonly contain isolated bytes from encodings such as ISO-8859-1 in comments. Strict UTF-8 decoding caused `tokensave sync` to skip those files entirely.

Closes #241

## Changes

- Fall back to lossy UTF-8 decoding in the shared source-file reader, replacing malformed sequences with U+FFFD.
- Add a unit regression using the issue's `W\xfcrkner` example.
- Add an end-to-end test confirming a C function remains indexed and searchable when its file contains an invalid UTF-8 byte.
- Document the fix under `[Unreleased]` in `CHANGELOG.md`.

## Test plan

- [x] `cargo test` passes

The issue's invalid `0xFC` byte example is covered through the source reader and the full `TokenSave::init` + `index_all` pipeline. The literal CLI reproduction was not run manually.

Additional check: `cargo fmt -- --check` passes.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 
 ## [Unreleased]
 
+### Fixed
+- **Source files containing invalid UTF-8 bytes are indexed instead of skipped (#241).** Legacy C/C++ sources often contain isolated non-UTF-8 bytes in comments; source reads now replace malformed byte sequences with the Unicode replacement character while preserving the existing UTF-8 and BOM-marked UTF-16 handling.
 
 ## [7.4.0] - 2026-07-16
 
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -4,8 +4,10 @@ use std::path::Path;
 use sha2::{Digest, Sha256};
 
 /// Read a source file to a UTF-8 string, transparently handling UTF-16 LE/BE
-/// (detected via BOM). Returns an IO error only when the file genuinely cannot
-/// be read or decoded.
+/// (detected via BOM). Invalid non-BOM UTF-8 sequences are replaced with the
+/// Unicode replacement character so legacy-encoded source can still be parsed.
+/// Returns an IO error only when the file cannot be read or BOM-marked UTF-16
+/// cannot be decoded.
 pub fn read_source_file(path: &Path) -> std::io::Result<String> {
     let bytes = std::fs::read(path)?;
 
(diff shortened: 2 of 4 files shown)


aovestdipaperino:
Thank you @janstol for the fix and the thorough test coverage — legacy C/C++ sources with stray Latin-1 bytes will now index cleanly. Much appreciated!

PR #247, opened 2026-07-19
fix: count MCP overhead and cap baseline in token-savings metric

janstol:
## Summary

- Fixes systematic over-estimation in the MCP `tokensave_metrics` before/after and the persisted `tokens_saved` counter (surfaced as `approx_tokens_saved`).
- Counts overhead the metric previously ignored (tool schema listing, per-call request framing), caps the "before" baseline to what a tool actually returned instead of the full weight of every touched file, and makes the persisted counter track the session's true net signed total instead of summing each call's independently-clamped net.

## Motivation

The savings metric is the thing tokensave uses to show its own value, so it needs to be trustworthy. It was over-counting in several independent ways: never charging for the ~85-tool schema payload or the request itself, crediting full-file weight to tools that only return references or partial content, measuring `after` before some response content was appended, accumulating a gross rather than net counter, and — once those were first addressed — two more subtle gaps: the one-time schema charge mostly vanished instead of offsetting later calls' savings, and a metrics-line cost was folded into `after` even when the line was never sent.

## Changes

- `src/mcp/tools/accounting.rs` (new): `BaselinePolicy`/`cap_baseline` (response-scaled baseline cap), `request_overhead_tokens`, `schema_overhead_tokens`, and `settle_session_debt` (pure carry-forward debt arithmetic), all unit tested.
- `src/mcp/server.rs`: `handle_tools_call`'s accounting block now folds request/schema overhead into `after` before measuring it; gates the schema charge on `tools/list` actually having been served (`schema_served`); caps `before` via `cap_baseline`; carries any per-call shortfall forward as `session_debt` (`AtomicI64`) instead of discarding it; and shares one `emit_metrics_line` condition between appending the metrics line and counting its cost, so the two can't diverge.
- `tests/mcp_server_test.rs`: integration coverage for capped/net metrics, schema overhead surviving a failed first call and requiring `tools/list`, baseline behavior for uncached vs. cached reads, a `before=0` call never carrying a metrics line, and the persisted counter holding flat while session debt from the schema charge is paid down across real calls.
- `CHANGELOG.md`: `[Unreleased]` entry describing all of the above.

## Test plan

- [x] `cargo test` passes (`cargo test --workspace --features test-transport`)
- [x] `cargo clippy` has no new warnings (`cargo clippy --workspace --all-targets --features test-transport -- -D warnings`)
- [x] `cargo fmt --all -- --check` passes
- [ ] Tested manually

## Checklist

- [x] Breaking changes documented (if any) — none; this only changes reported/persisted metric values, not any public API or tool behavior


aovestdipaperino:
Thanks @janstol — merged. This is a careful, well-reasoned fix: pulling the baseline/overhead/debt arithmetic into a pure, unit-tested `accounting` module keeps the accounting logic honest and separately verifiable, and the integration coverage (capped baselines, schema charge surviving a failed first call and requiring `tools/list`, the `before=0` no-metrics-line case, session-debt paydown) gives real confidence the metric now reflects true net savings. Rebased onto master for the CHANGELOG conflict before merging. 🙏

aovestdipaperino:
Hi @janstol — following up now that this is in. The accounting framework here is solid and I'm glad to have it; one small correction on the schema-overhead piece specifically.

The one-time charge is computed over the full `tools/list` payload (`schema_overhead_tokens(&get_tool_definitions())`), i.e. all ~85 tools. But only the three `anthropic/alwaysLoad` tools — `tokensave_search`, `tokensave_context`, `tokensave_status` — are actually loaded into the model's context up front. The rest are marked deferred and their schemas never enter context unless the client fetches one on demand (via ToolSearch). So debiting the whole listing over-states the real up-front cost by more than an order of magnitude, which would eat into the very savings the metric is meant to report.

The rest of your design already handles this correctly once the input is right — the `schema_served` gating, the debt carry-forward, all of it. It's purely a question of *which* definitions feed the estimate.

I've put up the fix in #249: a `get_always_load_tool_definitions()` that filters on the `alwaysLoad` `_meta` marker, with the overhead computed over just that subset (all your schema-overhead tests still pass). Flagging it here so the reasoning is on the record next to the original change. Thanks again for the thorough work on this — the module structure made the correction a one-liner at the call site. 🙏

Issue #251, opened 2026-07-19
Extend the Factory Droid `PreToolUse` guardrail to the native `Grep` tool


rNoz:
The Droid `PreToolUse` guardrail hook (follow-up #163; branch `rnoz/droid-hooks`) registers only the `Execute` matcher, so it sees shell `grep`/`rg`/`ag` but is blind to Droid's **native `Grep` tool**. A symbol-shaped `Grep` on a code file in an indexed project therefore bypasses the guardrail and returns raw match lines where `tokensave_search` would return a compact symbol list. This is the one remaining redirectable tool where Droid is behind the Claude hook.

### Proposed behavior

Widen the install matcher from `Execute` to `^(Execute|Grep)$` (anchored so it matches only those exact tool names, verified live to still fire). The shared decision core (`evaluate_hook_decision_core`) already classifies a `pattern` payload; this is primarily a wiring gap. Two small adaptations keep the core correct for both harnesses:

- read `glob_pattern` (Droid) alongside `glob` (Claude);
- redirect only an explicit `output_mode: "content"`; omitted, malformed, unknown, and cheap path-only modes (Droid's `file_paths`, Claude's `files_with_matches`/`count`) pass through.

### Why only `Grep` (not `Read`/`LS`/`Glob`/`Task`)

Droid's hook contract is a **hard block** (exit 2 + stderr) with no confirmed soft-steer channel. A block is a net win only when the tokensave equivalent is non-lossy, the classifier can tell the call is redirectable, and blocking doesn't strand the agent. Only `Grep` satisfies all three:

- **Read**, path carries no symbol intent; hits markdown/JSON/`.env`; blocking strands legit reads. Would need soft-steer Droid lacks.
- **LS**, already cheap; `tokensave_files` is a lossy subset (no size/mtime).
- **Glob**, paths, not content; no symbol signal to classify on.
- **Task**, Droid always sends a typed `subagent_type`, so the current core (which blocks only Claude's capitalized `Explore` outright and steers untyped prompts) never fires. Droid's lowercase `explorer` subagent IS a read-only research fan-out worth guarding, but that needs a `Task` matcher plus a hard-block false-positive policy, so it is a deliberate fast-follow (`droid-explorer-subagent-block`), not part of this PR.

[...]

PR #252, opened 2026-07-19
feat(droid): extend PreToolUse guardrail to the native Grep tool

rNoz:
## Summary

- Widen the Droid `PreToolUse` install matcher from `Execute` to `^(Execute|Grep)$` (an anchored regex, verified live to still fire on both tools), so a symbol-shaped `pattern` routed through Droid's **native `Grep` tool** (previously invisible to the guardrail) redirects to `tokensave_search`/`tokensave_callers_for`, reaching parity with the Claude hook's Grep coverage.
- The shared core (`evaluate_hook_decision_core`) already classifies a `pattern` payload. It now reads `glob_pattern` alongside Claude's `glob`, and redirects only an explicit `output_mode: "content"`. Omitted, malformed, unknown, and cheap path-only modes pass through, matching both harness defaults and covering Droid's `file_paths` without guessing that an undeclared search requested content.
- Re-installing over an older `Execute`-only entry migrates it in place (matched by the exact binary/subcommand command shape, not matcher value) instead of leaving a duplicate that would fire the hook twice. Lookalike binaries and customized commands with trailing arguments remain foreign.
- Implements the design of https://github.com/aovestdipaperino/tokensave/issues/251

## Motivation

`rnoz/droid-hooks` shipped the Droid `PreToolUse` guardrail but registered only the `Execute` matcher (Droid's shell tool), deliberately staying fail-open while the integration was validated. That leaves Droid's own `Grep` tool unguarded: a symbol-shaped `Grep` on a code file returns raw match lines where a symbol lookup would return a compact list. Measured this session on this repo: `tokensave_search "..."` reported `before=5172 after=67` tokens for a symbol query. `Grep` is the one remaining redirectable tool where Droid is behind Claude.

## Design Decision: Why `Grep`, and only `Grep` (per-tool analysis)

Droid's hook contract is a **hard block** (exit 2 + stderr). There is no confirmed "suggest but allow" (soft-steer) channel like Claude's `additionalContext`. So every matcher we add can only *deny* a call and force the agent elsewhere. A block is a net win only when all three hold: (1) tokensave has a non-lossy equivalent, (2) the classifier can tell *this* call is redirectable with high precision, (3) blocking doesn't strand the agent.

| Tool | tokensave equiv | Classify intent? | Hard-block safe? | Verdict |
|---|---|---|---|---|
| **Grep** | `tokensave_search`/`callers_for`, non-lossy | Yes: identifier-shaped pattern + code target + indexed project | Yes: names exact tool + escape hatch | **Ship** |
| Read | `node`/`body` need a *symbol*; Read gives a *path*; also hits md/json/`.env` | No | No, strands legit reads; needs soft-steer | Defer |
| LS | `tokensave_files` (indexed only, no size/mtime) | n/a | No, LS already cheap, equiv lossy | Defer |
| Glob | `tokensave_files --pattern` (partial) | No symbol signal (paths, not content) | Low reward | Defer |
| Task | only *research* tasks; core blocks `Explore`/untyped | Droid Task always **typed** → classifier ~never fires | Harmless no-op | Skip |

**Parity, not subpar.** Claude's core redirects three shapes: Grep-`pattern`, Bash-grep-`command`, and its `Explore` agent. Droid already covers Bash-grep (via `Execute`) and has no `Explore` agent to match. So `Execute + Grep` = everything Claude's core does that is relevant to Droid.

**Execute pipe-chain rewriting is out of scope (and likely rtk-land).** The hook allows/blocks a whole `Execute`; it never rewrites. Decomposing `cd && git log | grep ...` into stages to block inner greps is fragile shell parsing, and arbitrary-command interception is arguably rtk's domain. The pending grep-target issues already frame ownership as "may belong in rtk, not tokensave."

## Changes

- **`src/agents/droid.rs`**
  - `DROID_HOOK_MATCHER`: `"Execute"` → `"^(Execute|Grep)$"` (Factory treats the value as a regex; the pattern is anchored so it matches only those exact tool names, not a future name that merely contains `Execute`/`Grep`, and was verified live to still fire on both). Doc comment now records the per-tool rationale, the anchoring reason, and the deliberate exclusions.
  - New `is_tokensave_droid_hook(entry)` helper: identifies our hook by an exact tokensave binary plus `hook-droid-pre-tool-use` subcommand rather than the matcher value, so an older `Execute`-only install is recognized without claiming a foreign lookalike wrapper.
  - `install_hook` now **upserts**: if a tokensave entry with the current matcher exists it skips; otherwise it drops any existing tokensave entry (migrating a stale matcher) and appends the current one, preserving every non-tokensave hook. `uninstall_hook` uses the same ownership check. `doctor_check_hook` recognizes the owned entry but separately requires the current matcher, so an older `Execute`-only install is reported as stale instead of healthy.
- **`src/hooks.rs`**
  - `evaluate_grep_tool_input`: read `glob_pattern` as a fallback for `glob`; redirect only explicit content mode so omitted/default, malformed, unknown, and path-only modes fail open. The Droid handler doc comment now notes `Grep` reaches it.
- **`tests/hooks_test.rs`**: +6 Droid native-`Grep` cases (omitted/default mode passes; explicit content on code blocks; `glob_pattern` on `.rs` blocks and on `.md` passes; `file_paths` passes; non-code passes; escape hatch) plus a shared omitted-mode regression.
- **`tests/droid_agent_test.rs`**: existing matcher assertions updated to `^(Execute|Grep)$`; +4 lifecycle cases (stale `Execute`-only entry upgraded in place with no duplicate; uninstall removes a stale entry; foreign lookalike and trailing-argument commands survive install; `doctor` rejects a stale matcher).

## Test plan

- [x] `cargo test --test hooks_test`: 85 passed, 0 failed (+6 Droid Grep cases plus the shared omitted-mode regression; all prior cases unchanged).
- [x] `cargo test --test droid_agent_test`: 30 passed, 0 failed (+4 lifecycle cases; matcher assertions updated to the anchored value).
- [x] `cargo test --test mcp_handler_test`: 174 passed, 0 failed.
- [x] `cargo test` (full workspace): 2,469 passed across 96 suites, 0 failed.
- [x] `cargo clippy --workspace --all-targets`: clean.
- [x] `cargo fmt --all -- --check`: clean.
- [x] Real `droid exec` session (see "Real-world verification" below): Factory fires `PreToolUse` for the native `Grep` tool under both the unanchored and the shipped anchored `^(Execute|Grep)$` matcher; a symbol `Grep` on `src` blocks (exit 2), the block is deterministic on retry, and the model self-recovers (escape hatch, then a plain read); the escape hatch, a non-code `glob_pattern`, and `file_paths` mode all pass through.

### Real-world verification

Testing the anchored matcher. Method: the branch's own `cargo build --release` binary; a scratch project (`/tmp/droid-grep-verify`) indexed with `tokensave init` whose `src/main.rs` defines `handle_request_for_widget`; `tokensave install --agent droid --local` (which writes matcher `^(Execute|Grep)$`); and `droid exec --model minimax-m3 --skip-permissions-unsafe --cwd <scratch>`. The scratch `AGENTS.md` was emptied so the model reaches for the native `Grep` tool rather than being told to use tokensave. Excerpts below are verbatim from the session output.

**The premise this validated:** Factory runs `PreToolUse` hooks for the native `Grep` tool, not only `Execute`, and honors an anchored regex matcher. The block fired identically under the unanchored `Execute|Grep` and the shipped anchored `^(Execute|Grep)$` value, confirming anchoring does not stop the hook from matching the `Grep` tool name.

#### Test A, native `Grep` on a code dir is blocked, model self-recovers

`Grep {pattern:"handle_request_for_widget", path:".../src", output_mode:"content", line_numbers:true}`. The model's verbatim report:

> The native `Grep` tool was intercepted by a project `PreToolUse` hook installed by the `tokensave` MCP installer. [...] `STOP: This Grep targets a code file in a tokensave-indexed project and the pattern handle_request_for_widget looks like a symbol name. Use tokensave_search (definition) or tokensave_callers_for (usages) instead [...] set TOKENSAVE_DISABLE_GREP_HOOK=1 in the shell.` I retried the same `Grep` call once more to confirm the block is deterministic; same verbatim error.

The model then abandoned `Grep`: it ran `Execute` with `TOKENSAVE_DISABLE_GREP_HOOK=1 grep -rn handle_request_for_widget src` (the override the message names) and a `Read`, and noted `tokensave_search` as the project's preferred lookup. Both a `worker`-shaped recovery and the escape hatch are exercised; the block itself is the point.

#### Test A', anchored matcher still fires

Repeating Test A after rewriting the installed matcher to `^(Execute|Grep)$` produced the same block (STOP message, deterministic). This is the evidence behind shipping the anchored form: it is safer (no substring collision) and loses nothing (still matches `Grep` live).

#### Test B, escape hatch

With `TOKENSAVE_DISABLE_GREP_HOOK=1`, the same `Grep` call passes (exit 0). Seen both in the model run (its override grep ran and returned the two match lines) and in the direct-CLI matrix below.

#### Test C, control, non-code target passes

`Grep {pattern, glob_pattern:"**/*.md"}` passes through (confirms `glob_pattern` is read and a non-code target is not redirected).

#### Direct-CLI proof (no model spend), verbatim exit codes

Run from the indexed scratch project, piping Droid-shaped events into `tokensave hook-droid-pre-tool-use` (ambient `TOKENSAVE_DISABLE_GREP_HOOK` unset except the last row):

| Event | Exit |
|---|---|
| `Grep {pattern, path:"src"}` (no `output_mode`) | 0 |
| `Grep {pattern, glob_pattern:"**/*.rs", output_mode:"content"}` | 2 |
| `Grep {pattern, glob_pattern:"**/*.md", output_mode:"content"}` | 0 |
| `Grep {pattern:"TODO", path:".", glob_pattern:"**/*.md", output_mode:"content"}` | 2 (known shared classifier bug, split to `grep-path-glob-precedence`) |
| `Grep {pattern, path:"src", output_mode:"file_paths"}` | 0 |
| `Grep {pattern, path:"src", output_mode:"content"}` | 2 |
| `Execute {command:"grep -rn <symbol> src/main.rs"}` | 2 |
| `Grep {pattern, path:"src"}` with `TOKENSAVE_DISABLE_GREP_HOOK=1` | 0 |

The first row follows Droid's path-only default. Only an explicit `content` request is eligible for redirection. The `path` + Markdown-glob false-positive is documented rather than hidden, and is queued as the next narrow shared-classifier fix before the `explorer` matcher follow-up.

#### Notes from the run

- **The block is independent of AGENTS.md rules.** The scratch project's `AGENTS.md` was emptied and the model was told to use `Grep`; it still got exit
  2. So the guardrail does not rely on prompt-rule discipline, it enforces at the tool boundary.
- **Recovery path depends on what the project offers.** In this rules-free scratch (no tokensave MCP registered, empty AGENTS.md) the model recovered via the named escape hatch (`TOKENSAVE_DISABLE_GREP_HOOK=1` + `Execute` grep) and a `Read`, and explicitly noted `tokensave_search` as the preferred lookup. In a real project with the tokensave MCP server and AGENTS rules present, the model goes straight to `tokensave_search`/`tokensave_callers_for`.
- **Install in a single scope to avoid a double-fire.** Factory merges every settings source and runs each one, so the same logical hook installed in *both* `~/.factory/settings.json` (global) and `<repo>/.factory/settings.json` (project) fires twice. This is not a tokensave defect (each file holds one entry) nor a Factory defect (running each configured source is reasonable); keep the hook in one scope. Both invocations return identical results, so there is no functional impact meanwhile.
- **Token framing.** The redirect replaces a `content` grep with a compact symbol lookup; on this repo a real `tokensave_search` reported `before=5172 after=67` tokens for a symbol query. The win scales with file/match size, not the 8-line scratch fixture.

## Checklist

- [x] Breaking changes documented (none; additive matcher widening + a migrating install that removes a duplicate an older version could have left)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 
 ## [Unreleased]
 
+### Added
+- **Factory Droid guardrail hook now also covers the native `Grep` tool (follow-up to the Droid `PreToolUse` hook).** The install matcher widens from `Execute` to `^(Execute|Grep)$` (an anchored regex, verified live to still fire on both tools, so it can never collide with a future tool name that merely contains `Execute`/`Grep`), so a symbol-shaped `pattern` on a code target routed through Droid's own `Grep` tool (previously invisible to the hook) redirects to `tokensave_search`/`tokensave_callers_for`, reaching parity with the Claude hook's Grep coverage. The shared decision core now also reads Droid's `glob_pattern` field alongside Claude's `glob`, and redirects only an explicit `output_mode: "content"`; omitted, malformed, unknown, and cheap path-only modes (including Droid's `file_paths`) pass through. Re-installing over an older `Execute`-only entry migrates it in place instead of adding a duplicate, `doctor` flags stale matchers instead of reporting them healthy, and install/uninstall preserve lookalike custom wrappers by requiring an exact tokensave command. `Read`/`LS`/`Glob`/`Task` are deliberately not matched (lossy or unclassifiable equivalents, and Droid's hard-block contract has no soft-steer channel); see the PR for the full per-tool analysis.
 
 ## [7.4.1] - 2026-07-19
 
--- a/src/agents/droid.rs
+++ b/src/agents/droid.rs
@@ -34,13 +34,25 @@ pub struct DroidIntegration;
 /// before a tool call executes.
 const DROID_PRE_TOOL_EVENT: &str = "PreToolUse";
 
-/// The `Execute` tool is Droid's confirmed shell tool — the only tool name we
-/// register a matcher for. The sub-agent/task launch tool name is
-/// unconfirmed in Factory's public docs, so we
-/// deliberately do not guess a matcher for it: an unmatched tool never
-/// reaches our hook subprocess at all, which is the safest form of
-/// "fail open" for a tool name we can't verify.
-const DROID_HOOK_MATCHER: &str = "Execute";
+/// Regex matcher (Factory treats the value as a regex) covering Droid's two
+/// tool names whose payloads the shared decision core can classify: `Execute`
+/// (shell, a symbol-shaped `grep`/`rg`/`ag` on a code file) and `Grep`
+/// (Droid's native content search, a symbol-shaped `pattern` on a code
+/// target). Both redirect to `tokensave_search`/`tokensave_callers_for`, which
+/// return a compact symbol list instead of raw match lines. The pattern is
+/// anchored (`^(...)$`) so it can only match those exact tool names, never a
+/// future tool whose name merely contains `Execute`/`Grep` as a substring;
+/// anchored matching was verified live to still fire on both tools.
+///
+/// Deliberately excluded (see the PR doc for the full per-tool analysis):
+/// `Read`/`LS`/`Glob` have only lossy tokensave equivalents and no way to
+/// infer symbol intent from a bare path, and Droid's hook contract is a hard
+/// block (exit 2 + stderr) with no confirmed soft-steer channel, so blocking
+/// them would strand legitimate reads. `Task` always carries a typed
+/// `subagent_type` on Droid, so the research classifier (which only fires on
+/// Claude's `Explore` or an untyped research prompt) would never trigger.
+/// Unmatched tools never reach our hook subprocess, the safest "fail open".
+const DROID_HOOK_MATCHER: &str = "^(Execute|Grep)$";
 
 /// Subcommand invoked by the installed hook.
 const DROID_PRE_TOOL_HOOK: &str = "hook-droid-pre-tool-use";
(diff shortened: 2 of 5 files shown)


aovestdipaperino:
Merged — thanks @rNoz! The per-tool analysis table was especially useful; agreed on shipping Grep only and deferring Read/LS/Glob/Task given Droid's hard-block contract. The `output_mode` default fix (omitted → pass through) is a good catch on its own.

PR #254, opened 2026-07-20
refactor: split volatile state out of config.toml into state.toml

janstol:
## Summary

- Splits `~/.tokensave/config.toml` into a stable, dotfile-friendly `config.toml` (`upload_enabled`, `watcher_debounce`, `extraction_timeout_secs`, `wildcard_permissions`) and a new machine-local `state.toml` holding everything volatile (`pending_upload`, cached version/pricing/flags, timestamps, `installed_agents`).
- `UserConfig`'s in-memory struct and public API (`load()`, `save()`, `is_fresh()`, field access) are unchanged, so none of the ~40 existing call sites need to change. Legacy single-file installs migrate transparently on first load.
- Atomic-write hardening for the new two-file save path: symlink-safe writes (including dangling and multi-hop chains, with cycle detection), a same-process save/save and load/save mutex to prevent torn config+state pairs, a safe Windows non-replacing-`rename` fallback, corrupt-`state.toml` preservation on load, and preservation of each file's existing permission mode (instead of resetting to the umask default) across saves.

## Motivation

`config.toml` mixed stable user preferences with frequently-changing cached/timestamp state, so checking it into a dotfiles repo produced constant, meaningless diffs. This splits persistence so `config.toml` stays stable and diff-free while volatile state moves to an untracked `state.toml`, with no data loss for existing installs and no breaking change to the in-memory config API.

## Changes

- `src/user_config.rs`:
  - Added `state_path()`, private `ConfigFile`/`StateFile` (de)serialization views, and `apply_state()` to overlay `state.toml` onto a `UserConfig` parsed from `config.toml` (which still recovers legacy state fields from an old mixed file for one-time migration).
  - `save()` now writes `state.toml` then `config.toml` via a shared `write_atomic()` (temp file + rename), serialized under a process-wide `SAVE_LOCK` mutex so two same-process savers (e.g. background MCP work racing a shutdown save) can't interleave writes into a mismatched pair.
  - `load()` now takes the same `SAVE_LOCK` while reading both files, so it can't return a config/state snapshot torn across a concurrent same-process `save()`.
  - `resolve_write_target()` walks symlink chains (including dangling and multi-hop ones) hop-by-hop, bounded by `MAX_SYMLINK_HOPS`, so writes go through a dotfiles symlink instead of replacing it; a cyclic chain fails the save rather than guessing a target to write through.
  - `replace_via_backup()` handles Windows' non-replacing `rename` by moving the existing destination aside first and restoring it if the replacement rename fails, instead of deleting-then-retrying.
  - `preserve_corrupt_state_file()` renames an unparseable `state.toml` aside on load rather than letting the next routine save silently overwrite it with defaults.
  - `set_tmp_file_mode()` (Unix only) sets the temp file's mode to match the resolved target's existing mode before the rename (falling back to a restrictive `0o600` for a new file), so saves no longer reset a dotfiles-managed `config.toml`'s permissions to the umask default; the save fails outright if the mode can't be set.
  - 22 unit tests covering round-tripping, the config/state split, migration, concurrent saves (mixing and torn reads), symlink handling (plain/dangling/nested/cyclic), the Windows fallback, corrupt-state preservation, and file-mode preservation.
- `docs/USER-GUIDE.md`, `docs/DESIGN-DOC.md`: updated to describe the two-file layout and clarify only `config.toml` is meant to be dotfile-tracked.
- `CHANGELOG.md`: single `[Unreleased]` entry describing the split and all of the atomic-write hardening above.

No version bump — this needs no forced resync or agent reinstall per `TOKENSAVE-VERSIONING.md`.

## Test plan

- [x] `cargo test` passes (full default feature set)
- [x] Unit tests re-run 3x for the new concurrency/permission tests to rule out flakiness

## Checklist

- [x] `CHANGELOG.md` updated (under `[Unreleased]`, no version bump)
- [x] Breaking changes documented (if any) — none; `UserConfig`'s public API is unchanged


janstol:
The Windows CI job failed with an unrelated, pre-existing flaky crash — not something introduced by this PR.

The `Test Windows` job crashed with `STATUS_ACCESS_VIOLATION` (0xc0000005) inside the `mcp_handler_test` integration test binary (173/174 tests in that binary had already passed; Rust's default test harness doesn't print a per-test start marker, so the crash isn't attributable to a specific test).

Checking recent CI history on `master` turns up the same exact crash signature twice already, on commits that never touch `src/user_config.rs` and predate this branch:
- Run [29659277581](https://github.com/aovestdipaperino/tokensave/actions/runs/29659277581) (commit 5a5e3e8) — crashed in `integration_test` with `0xc0000005 STATUS_ACCESS_VIOLATION`.
- Run [29695422383](https://github.com/aovestdipaperino/tokensave/actions/runs/29695422383) (commit 30e01cc) — crashed in `mcp_handler_test` with the same `0xc0000005 STATUS_ACCESS_VIOLATION`.

This PR's changes are confined to `src/user_config.rs` and are pure safe Rust file I/O (no `unsafe`, no FFI), so they aren't a plausible source of a native access violation. This looks like a pre-existing, intermittent crash on the Windows runner (possibly under test-harness parallelism) rather than something this change caused.

aovestdipaperino:
Thanks a lot for this, @janstol! The config/state split is exactly the right call — keeping `config.toml` diff-free for dotfiles was overdue, and I appreciate that you kept `UserConfig`'s in-memory API unchanged so none of the ~40 call sites had to move.

The atomic-write hardening that came along with it is the part I'm most glad to have: symlink-chain resolution with cycle detection, the save/load mutex against torn config+state pairs, corrupt-`state.toml` preservation, and preserving the existing file mode instead of resetting to the umask default. That's a lot of careful edge-case work, and the 22 tests covering it make it reviewable. Merged.

PR #257, opened 2026-07-21
feat: decouple agent rules into managed files (#256)

### LINKED ISSUE #256: [Suggestion] Inject agent global instructions outside user's AGENTS.md/CLAUDE.md when possible
Instead of injecting that `## MANDATORY: No Explore Agents When Tokensave Is Available` in `AGENTS.md`/`CLAUDE.md`, use these alternatives on these tools:
- **Claude**
  - Store a separate `tokensave.md` in `~/.claude/rules/`, which is loaded the same way. It could be a symlink so it is always up-to-date with current version of tokensave;
- **OpenCode**
  - Edit `~/.config/opencode/opencode.json`:
    ```jsonc
    {
      //...
      "instructions": [
        //...
        "/opt/homebrew/opt/tokensave/instructions.md"
      ],
      //...
    }
    ```
  - Target instruction is within tokensave installation, so any updates to it would be picked up automatically

janstol:
## Summary

- Claude Code and OpenCode rules now live in a tokensave-owned managed file instead of being appended into the user's CLAUDE.md/AGENTS.md.
- Installing over a pre-#256 setup migrates automatically: the old inline block is stripped and the managed file takes its place.
- `tokensave doctor` checks for the managed file (and, for OpenCode, the `instructions` entry) instead of scanning for the old marker text.

## Motivation

Closes #256. tokensave previously appended a ~40-line "prefer tokensave tools" block directly into `~/.claude/CLAUDE.md` and `~/.config/opencode/AGENTS.md`, guarded by a marker check that skipped the append once present — so the block never refreshed on upgrade even as the rule text improved, and uninstall had to locate and cut the block back out of a file the user otherwise hand-maintains.

## Design notes

- **Docs checked before implementing:** Claude Code auto-loads `~/.claude/rules/*.md` recursively at the same priority as CLAUDE.md, with user-scope files loading with no approval dialog ([code.claude.com/docs/en/memory](https://code.claude.com/docs/en/memory)), so a dedicated `rules/tokensave.md` needs no CLAUDE.md edit at all. OpenCode's `opencode.json` supports an `"instructions"` array of file paths/globs, combined with AGENTS.md ([opencode.ai/docs/rules](https://opencode.ai/docs/rules/), [opencode.ai/docs/config](https://opencode.ai/docs/config/)).
- **Why a managed file and not a symlink:** Claude Code's `rules/` dir does support symlinks, but a symlink pointing into tokensave's install directory breaks whenever the binary moves (`cargo install`, Homebrew upgrade, `tokensave upgrade`) and needs elevated privileges to create on Windows. `resync_installed_agents` already re-runs `install` for every agent on minor/major version bumps, so unconditionally overwriting a real file gets the same "content stays current" property as a symlink, with none of that fragility, and works identically cross-platform.

## Changes

- `src/agents/mod.rs`: shared `managed_rules_markdown`, `write_managed_rules_file`/`remove_managed_rules_file`, and a new `remove_legacy_rules_block` migration engine — all atomic (temp-file + rename, backed up first) and symlink-safe (a symlinked CLAUDE.md/AGENTS.md or managed rules file has its real target written/removed while the symlink itself is preserved, for dotfiles-managed setups).
- `src/agents/claude.rs`: rules now written to `~/.claude/rules/tokensave.md` (global) / `<project>/.claude/rules/tokensave.md` (local), auto-loaded via Claude Code's existing `rules/*.md` convention. `uninstall_claude_md_rules` is now a thin wrapper around the shared migration engine.
- `src/agents/opencode.rs`: rules now written to `~/.config/opencode/tokensave.md` (global) / `<project>/tokensave.md` (local), wired in via a deduped entry in `opencode.json`'s `"instructions"` array. Global installs reference the file by **absolute path** — OpenCode resolves a relative `instructions` entry by globbing up from the current *project's* working directory, not from `opencode.json`'s own directory, so a bare `"tokensave.md"` there would never resolve to the actual global file. Local installs keep the relative filename, which resolves correctly and stays portable across checkouts.
- Migration failures now fail the install rather than being silently swallowed; uninstall-path calls stay best-effort, matching this codebase's existing uninstall convention.
- `CHANGELOG.md`: `[Unreleased]` entry describing the behavior change (no `Cargo.toml` version bump).

## Test plan

- [x] `cargo test` passes (full workspace, default features)
- [x] New unit/integration tests cover: managed-file install/migration/idempotency/uninstall for both agents, symlink preservation on uninstall and on migration's empty-file cleanup, atomic-write failure leaving originals untouched, adjacent-user-heading preservation during migration, and the OpenCode global absolute-path `instructions` wiring.

## Checklist

- [x] `CHANGELOG.md` updated (under `[Unreleased]`, no version bump)
- [x] No breaking changes (existing installs migrate automatically on next `tokensave install`/upgrade)


aovestdipaperino:
Thanks @janstol — merged. Clean migration engine with the atomic writes and symlink-safety handled properly, and the absolute-path wiring for OpenCode's global `instructions` was a sharp catch. Appreciated the thorough test coverage. 🙏

PR #261, opened 2026-07-22
fix: don't revert config.toml preferences from stale/concurrent saves

janstol:
## Summary

- Fix `UserConfig::save()` silently reverting `config.toml` preferences (e.g. `wildcard_permissions`) that were changed by a concurrent process since this process's `load()`.
- Preserve an unparseable **or invalid-UTF-8** `config.toml`/`state.toml` on load by moving it aside to a `*.corrupt.<timestamp>.<pid>.<n>` backup instead of silently falling back to defaults and letting the next routine save overwrite it in place.
- Mark `UserConfig` `#[non_exhaustive]` and correct the module doc, since the per-instance merge baseline now lives in a crate-private field that blocks external struct-literal construction.

## Motivation

`UserConfig::save()` wrote its entire in-memory snapshot of `config.toml` as a whole-file overwrite. A long-lived process (the `serve` MCP daemon, a git `post-commit`/`post-checkout` hook running `sync`) that had loaded the file before another process changed a preference — a manual edit, or `tokensave install --wildcard-permissions` — would revert that change back to the stale value on its next routine save. Separately, an unparseable `config.toml`/`state.toml` was silently discarded on load with no way to recover the original content, and — as caught in review — invalid UTF-8 bytes were being treated as "file missing" rather than "file corrupt," bypassing that preservation entirely.

## Changes

- `src/user_config.rs`:
  - `save()` now re-reads the current on-disk `config.toml` and only overwrites the fields this process actually changed since its own `load()` (or most recent `save()`), preserving whatever another process wrote for every untouched field. The per-field baseline advances after every successful save to this process's own in-memory values, not the merged/disk-sourced file just written.
  - The on-disk read-modify-write span (in both `save()`'s merge and `load()`'s own config.toml read) is now held under a new cross-process advisory lock (`config.toml.lock`, via the existing `fs2` dependency) — the in-process `SAVE_LOCK` alone doesn't order writes from a second process.
  - An unparseable `config.toml`/`state.toml` is moved aside to a `*.corrupt.<timestamp>.<pid>.<n>` backup on load instead of being silently discarded; this resolves through symlinks first, so a dotfiles-managed `config.toml -> ~/dotfiles/config.toml` symlink survives (left dangling) instead of being detached from its target.
  - `load()`'s config.toml and state.toml reads now read raw bytes and decode UTF-8 explicitly (rather than `read_to_string`, which collapses "missing" and "present but invalid UTF-8" into the same `None`/`Err`), so invalid-UTF-8 files get the same corruption-preservation as unparseable-TOML files instead of being silently overwritten.
  - The per-field merge baseline lives in a new crate-private field (`pub(crate) loaded_config: Mutex<Option<ConfigFile>>`), preserving `UserConfig`'s `Send + Sync` guarantee (enforced by a compile-time assertion). Since this field isn't constructible via struct literal from outside the crate, `UserConfig` is now marked `#[non_exhaustive]` and the module doc is corrected to describe it as constructor-built (`UserConfig::load()` / `UserConfig::default()`), not literal-constructible — an intentional, documented break in external API surface (this crate's versioning is not a SemVer contract; see `TOKENSAVE-VERSIONING.md`).
- `tests/user_config_test.rs`, `tests/agent_test.rs`: switched from external struct literals to `UserConfig::default()`-then-mutate, the new required construction path.
- `CHANGELOG.md`: `[Unreleased]` entry under `### Fixed`.

## Test plan

- [x] `cargo test` passes (full `cargo test --workspace`, clean)
- [x] `cargo test --no-default-features --features lite --lib user_config` — 32/32 pass, including new regression tests for: stale-preference clobbering, cross-process lock ordering, corrupt-file preservation (unparseable TOML), and corrupt-file preservation (invalid UTF-8) for both `config.toml` and `state.toml`.

## Checklist

- [x] Breaking changes documented: `UserConfig` is now `#[non_exhaustive]` and no longer constructible via external struct literal (documented in the module doc and this description).


aovestdipaperino:
Thanks @janstol — merged into `master` via 13b97aa. Excellent work on this: the read-modify-write merge under a cross-process advisory lock, the corrupt-file preservation for both unparseable-TOML and invalid-UTF-8 cases, and the symlink-safe move-aside are all exactly right, and the regression tests covering the stale-clobber and cross-process ordering paths are much appreciated. The only merge adjustment was folding your `[Unreleased]` `### Fixed` entry into the existing one on master; no code changes.

PR #268, opened 2026-07-23
fix: register VS Code Copilot MCP server in mcp.json, not settings.json

### LINKED ISSUE #266: `install --agent copilot` reports success but never writes settings.json, and checks a legacy MCP location on modern VS Code
**Environment**
- tokensave **7.6.0** (binary at `~/.cargo/bin/tokensave`)
- VS Code **1.130.0**, macOS
- Agent: `copilot`

**Symptom — infinite `install` ↔ `doctor` loop**

`tokensave doctor` reports:
```
GitHub Copilot integration
  ✘ MCP server NOT registered in <VSCODE_USER>/settings.json — run `tokensave install --agent copilot`
  ✔ MCP server registered in ~/.copilot/mcp-config.json
  ✔ MCP server args include "serve"
```

Running the suggested command prints success:
```
$ tokensave install --agent copilot
✔ Added tokensave MCP server to <VSCODE_USER>/settings.json
```

…but settings.json is **not modified** — `diff` before/after is empty and no `mcp` key is added. So `doctor` keeps reporting `✘`, and the install↔doctor cycle never resolves.

**Bug 1 — false success / broken writer.** For the VS Code (`copilot`) target, the installer prints `✔ Added … to settings.json` without actually mutating the file. The success message should reflect a real, verified write: re-read the file and confirm `mcp.servers.tokensave` exists before printing `✔`.

**Bug 2 — wrong/legacy target on modern VS Code.** Since VS Code 1.102+, user MCP servers live in the dedicated `mcp.json` (e.g. `<VSCODE_USER>/mcp.json`), not in settings.json under an `mcp` key. tokensave already writes/checks `mcp.json` correctly — that check is `✔` and the tools work at runtime — but the extra settings.json check targets a deprecated location and produces a permanent false negative.

**Suggested fix**
- Treat `<VSCODE_USER>/mcp.json` as the source of truth for VS Code ≥ 1.102. Make `doctor` pass if **either** `mcp.json` **or** settings.json contains the server, and stop requiring settings.json.
- If the settings.json path is kept for older VS Code, fix the writer so it actually persists `mcp.servers.tokensave`, and only print `✔` after verifying the write.
- Writing to **both** settings.json and `mcp.json` on modern VS Code risks a duplicate server in the MCP UI — prefer writing only `mcp.json`.

**Workaround** that unblocks `doctor`: manually add the `mcp.servers.tokensave` block to settings.json, mirroring `mcp.json`:
```jsonc
"mcp": {
  "servers": {
    "tokensave": { "type": "stdio", "command": "<HOME>/.cargo/bin/tokensave", "args": ["serve"] }
  }
}
```

janstol:
## Summary

- `install --agent copilot` now registers the tokensave MCP server in VS Code's dedicated `User/mcp.json` (1.102+, GA MCP) instead of the legacy `settings.json` `mcp.servers` key.
- `doctor`/`has_tokensave` check `mcp.json` first, fall back to a legacy `settings.json` entry for back-compat, and warn if both are present (a stale duplicate).
- `uninstall` cleans up both locations.

Closes #266

## Motivation

Modern VS Code (1.102+, GA MCP) reads user-level MCP servers from `User/mcp.json` (top-level `servers` key), not `settings.json`. tokensave's `copilot` integration only ever wrote/read `settings.json`'s legacy `mcp.servers.tokensave` key.

Correcting the framing from my own initial investigation: VS Code doesn't just silently ignore a `settings.json` `mcp` entry. It actively migrates it — `McpConfigMigrationContribution` watches `settings.json`, installs any `mcp.servers` entry it finds into VS Code's own MCP registry (the thing that actually makes a server available to Copilot), and then strips the key back out of `settings.json`. Proof: [`mcpMigration.ts`](https://github.com/microsoft/vscode/blob/33b1d77b034ea548b2b43101c3b81782fecfab3e/src/vs/workbench/contrib/mcp/browser/mcpMigration.ts#L48-L60) in `microsoft/vscode`.

So the install may well have worked from VS Code's point of view. The actual bug was narrower: `doctor`/`has_tokensave` only ever inspected `settings.json`, and once VS Code had migrated the entry out (as it's designed to), tokensave kept reporting the registration as permanently missing — an unresolvable false-negative loop that looks identical, from the outside, to a genuinely broken install.

## Changes

- `src/agents/copilot.rs`:
  - New `install_vscode_mcp_json_server` writes `User/mcp.json` (top-level `servers`, with `type: "stdio"`, unlike the JetBrains variant), parsed/written as JSONC so comments and trailing commas in a hand-edited file don't error out `install` or make `doctor`/`has_tokensave`/`uninstall` treat it as empty.
  - `install` writes only `mcp.json` (+ VS Code Insiders, when present) — it never writes or reformats `settings.json`, even to clear a stale pre-#266 entry there, since rewriting a hand-maintained, possibly comment-bearing file during a routine install is more invasive than leaving a leftover for `doctor` to flag.
  - `uninstall` removes the entry from both `mcp.json` and legacy `settings.json`.
  - `doctor_check_vscode_settings` passes on either `mcp.json` or the legacy `settings.json` entry, and warns (doesn't fail) if both are present.
  - `has_tokensave` and `primary_config_path` updated to check/point at `mcp.json` first.
- `tests/copilot_agent_test.rs`: updated install/doctor assertions to target `mcp.json`; added coverage for JSONC parsing, modern-only/legacy-only/both-locations doctor behavior, Insiders, and that `settings.json` is left byte-for-byte untouched (including a `//` comment) by `install`.
- `CHANGELOG.md`: `[Unreleased]` entry under `### Fixed`.

## Test plan

- [x] `cargo test --test copilot_agent_test --test agent_test` passes (169 tests)
- [x] `cargo test --no-default-features --features lite` passes
- [x] `cargo clippy --workspace --all-targets` — no new warnings
- [x] `cargo fmt --all -- --check` — clean
- [ ] Not tested against a real VS Code install — I verified this by running `cargo test` against a temp `$HOME` (which is what the automated suite above already does), not by installing and opening actual VS Code. If you want to confirm against a real install:
  ```bash
  # macOS, using a temp $HOME so this doesn't touch your real config
  export HOME=$(mktemp -d)
  cargo run -- install --agent copilot
  cat "$HOME/Library/Application Support/Code/User/mcp.json"   # expect servers.tokensave
  cat "$HOME/Library/Application Support/Code/User/settings.json" 2>/dev/null  # expect: no such file, or no mcp.servers.tokensave key
  cargo run -- doctor   # expect the VS Code Copilot check to pass
  ```

## Checklist

- [x] No breaking changes (installer output path changes, but this is the bugfix itself; existing `settings.json` registrations keep working as a read-only fallback until VS Code migrates or the user removes them)


aovestdipaperino:
Merged — thank you @janstol! This was an exemplary PR: the corrected root-cause analysis (VS Code actively migrating the `settings.json` entry via `McpConfigMigrationContribution`, turning `doctor`'s check into a permanent false negative) saved a lot of investigation time, and the test coverage — JSONC handling, all three doctor combinations, and the byte-for-byte `settings.json` untouched assertion — is exactly what this area needed. The Windows CI failure was a pre-existing flake in an unrelated test binary (crash-at-exit, also seen on master) and passed on re-run. #266 is closed by this merge.

Issue #279, opened 2026-07-24
Grep guardrail false-positive when a broad `path` and non-code glob are both set


rNoz:
### tokensave version

7.4.1

### Operating system

macOS

### Installation method

Built from source

### What happened?

The shared Grep guardrail reads both `path` and the harness-specific glob field (`glob` for Claude, `glob_pattern` for Factory Droid), but `target_looks_like_code()` ignores the glob whenever `path` is non-empty. A request such as `{"pattern":"TODO","path":".","glob_pattern":"**/*.md","output_mode":"content"}` is therefore hard-blocked as a code search because `path:"."` wins, even though the effective target is restricted to Markdown files. The same request with only `glob_pattern:"**/*.md"` correctly passes.

This is user-visible on Droid because its native `Grep` commonly combines a search root with `glob_pattern`. The hook contract is exit 2 plus stderr, so this false-positive denies a legitimate documentation search and tells the agent to use symbol tools that cannot answer it. The classifier is shared, so Claude payloads using `path` plus `glob` have the same defect.

### Steps to reproduce

1. In an indexed project, build tokensave and pipe `{"tool_name":"Grep","tool_input":{"pattern":"TODO","path":".","glob_pattern":"**/*.md","output_mode":"content"}}` to `tokensave hook-droid-pre-tool-use`.
2. Observe exit code 2 and the symbol-search redirect.
3. Remove `"path":"."` and repeat.
4. Observe exit code 0.

### Error output

```shell
STOP: This Grep targets a code file in a tokensave-indexed project and the pattern `TODO` looks like a symbol name. Use tokensave_search ...
```

### Component

Agent integration (Claude, Cursor, etc.)

### Proposed behavior

Treat an explicit file filter as more specific than its broad search root. An explicit all-code `glob`/`glob_pattern` should classify as code; an explicit non-code or mixed code/non-code glob should pass through; an unclassifiable glob should fall back to the existing `path` rule. Keep the explicit `type` filter highest priority. Implement this as a conservative tri-state filter classification (`code`, `non-code`, `unknown`) so unknown syntax remains fail-open or falls back without turning every arbitrary glob into a hard block.

### Acceptance

- `path:"."` plus `**/*.md` passes for both Droid (`glob_pattern`) and Claude (`glob`).
- `path:"src"` plus `**/*.md` passes.
- `path:"."` plus `**/*.rs` still blocks.
[...]

PR #280, opened 2026-07-24
fix(hooks): Honor Grep file globs before broad search paths

rNoz:
## Summary

- Make the shared Claude and Factory Droid Grep guardrail classify an explicit `glob` / `glob_pattern` before a broad `path`.
- Preserve `type` as the highest-priority filter.
- Redirect code-only globs, pass non-code or mixed-extension globs, and fall back to the existing path behavior when the glob has no classifiable extension.
- Solves https://github.com/aovestdipaperino/tokensave/issues/279

## Motivation

Claude and Factory Droid can send both a broad search root and a narrower file filter. The guardrail previously discarded the filter whenever `path` was present:

```json
{
  "pattern": "TODO",
  "path": ".",
  "glob_pattern": "**/*.md",
  "output_mode": "content"
}
```

That request was hard-blocked as a code search even though it can only return Markdown. The redirect then points the agent at symbol tools that cannot answer the documentation search.

The fix uses conservative filter precedence:

1. Explicit `type` remains authoritative.
2. A syntactically classifiable glob is more specific than its path.
3. Every classified extension must be a tokensave code extension to redirect.
4. Any non-code extension makes the filter pass through.
5. A glob with no classifiable extension preserves the previous path classification.

## Changes

- `src/hooks.rs`: added a small glob-target classifier supporting ordinary extension globs, terminal brace lists, and unchanged extensionless-glob fallback behavior.
- `tests/hooks_test.rs`: added Claude and Droid regressions for broad paths, code-only and mixed filters, terminal suffixes, directory braces, unknown filters, and explicit-type priority.

## Test plan

- [x] Red tests reproduced the Claude and Droid false-positive before implementation.
- [x] `cargo test --test hooks_test`: 114 passed.
- [x] Direct Claude hook probe: `path:"."` + `glob:"**/*.md"` allowed.
- [x] Direct Claude hook probe: `path:"."` + `glob:"**/*.rs"` denied.
- [x] Direct Droid hook probe: `path:"."` + `glob_pattern:"**/*.md"` exited 0.
- [x] Direct Droid hook probe: mixed `**/*.{rs,md}` exited 0.
- [x] Direct Droid hook probe: Rust-only `**/*.rs` exited 2.
- [x] `cargo test`: 2,548 passed across 96 suites.
- [x] `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- [x] `cargo fmt --all -- --check`: clean.

## Scope

This change only fixes shared target classification. It does not widen Droid's matcher, add a new tool interception path, or change Bash command parsing.

## Checklist

- [x] User-visible behavior documented in `CHANGELOG.md`
- [x] No new dependencies
- [x] No secrets or private overlay files in the publishable branch


aovestdipaperino:
Merged. The filter-precedence ordering here is the right call — keeping `type` authoritative, requiring *every* classified extension to be code before redirecting, and falling back to the old path rule when a glob has no classifiable extension means this can only ever unblock searches that were previously false positives. The brace-expansion and terminal-suffix cases (`**/*.{spec,test}.ts` vs `**/*.{ts,js}.bak`) were the tricky ones and you covered both. Thanks for the fix and for tracking down #279.

Issue #281, opened 2026-07-24
Extend the Factory Droid `PreToolUse` guardrail to Droid's built-in `explorer` research subagent


rNoz:
The Droid `PreToolUse` guardrail (base hook + follow-up `Execute|Grep` matcher, branch `rnoz/droid-hooks-extra`) redirects symbol-shaped greps to tokensave's MCP symbol tools. It does not touch agent delegation. Droid ships a built-in `explorer` subagent described as a "fast read-only codebase exploration agent" that searches files, finds patterns, and reads code. That is exactly the search/read fan-out tokensave's MCP tools replace, and on an indexed project it is the single most wasteful unguarded path (a whole sub-session of grep/read where one `tokensave_context` call would answer the question).

It is unguarded today for two independent reasons:

- The Droid install registers only the `Execute|Grep` matcher, so a `Task` (subagent launch) tool call never reaches our hook subprocess at all.
- Even if it did, the shared decision core (`evaluate_hook_decision_core` in `src/hooks.rs`) only hard-blocks the *capitalized* `subagent_type == "Explore"` (Claude Code's built-in). Droid names its agent lowercase `explorer`, so the existing check never fires. Untyped prompts are steered by keyword, but Droid always sends a typed `subagent_type`, so that path never fires either.

### Proposed behavior

Register a `Task` matcher for the Droid agent and normalize its exact lowercase built-in `explorer` type in the Droid adapter before invoking the existing shared research block. This blocks it the same way Claude's `Explore` is blocked, with a message pointing at `tokensave_context` / `tokensave_search`, without widening Claude's accepted type names.

- Only the `explorer` type is blocked. Typed non-research subagents (`worker`, an implementer, any custom droid) are deliberate delegations and must pass, in line with the accepted #195 Option B (a typed non-Explore agent is never blocked on prompt text).
- The `.tokensave` index gate and the `TOKENSAVE_DISABLE_GREP_HOOK=1` escape hatch apply exactly as they do for the grep path.

### Why this is deferred (gates to clear first)

- **Droid's hard-block contract.** The hook can only deny (exit 2 + stderr); there is no confirmed soft-steer channel (see ROADMAP item "Guardrail soft-mode + Factory outreach"). Blocking a mis-typed or genuinely non-code `explorer` call strands the agent, so the classification must be high precision, or we accept "block + escape hatch" as the contract.
[...]

PR #282, opened 2026-07-24
feat(droid): Guard Droid's built-in `explorer` subagent with the PreToolUse hook

rNoz:
## Summary

- Route Droid `Task` subagent launches through the existing `hook-droid-pre-tool-use` path by widening `DROID_HOOK_MATCHER` to `^(Execute|Grep|Task)$`.
- Normalize the exact lowercase `explorer` type in Droid's adapter, redirecting it through the existing shared research block without changing Claude's case-sensitive behavior.
- Keep `worker` and every other typed subagent passing, and reuse the existing `TOKENSAVE_DISABLE_GREP_HOOK` escape hatch.
- Implements new feature https://github.com/aovestdipaperino/tokensave/issues/281

## Motivation

Droid ships a built-in `explorer` subagent described as a fast read-only codebase exploration agent that searches files and reads code; that is the exact search/read fan-out tokensave's MCP tools replace, and on an indexed project it is the single most wasteful unguarded path. Today it never reaches the hook because the Droid install only registers an `Execute|Grep` matcher, and even if it did the shared core only blocks Claude's capitalized `Explore`.

## Live decisions

- A real `droid exec` run emitted `tool_name: "Task"` with `tool_input.subagent_type: "explorer"`.
- The anchored matcher `^(Execute|Grep|Task)$` captured the real event, so the shipped matcher stays exact and collision-safe.
- The approved policy is an unconditional block for the exact lowercase `explorer` type on indexed projects, matching Claude's exact `Explore` block.
- `worker` and every other typed subagent remain allowed regardless of prompt text.
- The existing `TOKENSAVE_DISABLE_GREP_HOOK` escape hatch remains shared with agent redirection; no second environment variable is added.

## Changes

- `src/agents/droid.rs`: widen `DROID_HOOK_MATCHER` to `^(Execute|Grep|Task)$`; the install upsert migrates an existing `Execute|Grep` entry in place.
- `src/hooks.rs`: keep the shared core's exact Claude `Explore` rule unchanged; in the Droid adapter, normalize only a real `Task` with exact lowercase `explorer` and pass every other `Task` without prompt classification.
- `tests/hooks_test.rs`: add Droid `Task` cases for `explorer` block, `worker`/custom/untyped pass, escape-hatch pass, non-indexed pass, and unchanged lowercase Claude behavior.
- `tests/droid_agent_test.rs`: assert the widened matcher is written, the `Execute|Grep` to `Execute|Grep|Task` migration leaves exactly one tokensave entry, and uninstall removes it.

## Test plan

- [x] `cargo test --test hooks_test` passes with the new Droid `Task` cases.
- [x] `cargo test --test droid_agent_test` passes with the matcher migration cases.
- [x] Full `cargo test` passes.
- [x] `cargo clippy --all-targets --all-features -- -D warnings` is clean.
- [x] End-to-end `droid exec` run confirms an `explorer` launch is blocked and a `worker` launch passes, with the escape hatch working.

## Checklist


aovestdipaperino:
Merged. Thanks @rNoz — clean implementation, and the migration path from the `^(Execute|Grep)$` matcher plus the escape-hatch tests were exactly what this needed. Closes #281.

PR #284, opened 2026-07-24
feat(ruby): track method visibility modifiers

janstol:
## Summary

- Track per-class/module Ruby method visibility instead of hardcoding every node as `Visibility::Pub`
- Retroactively apply the symbol-list form (`private :foo, :bar`) and the singleton forms `private_class_method` / `public_class_method`
- Extract the inline form (`private def foo; end`), which was previously dropped entirely because `call` nodes weren't visited

## Motivation

The Ruby extractor marked every method as public regardless of `private`/`protected`/`public` modifiers. `dead_code`, `module_api`, and `impact` all key off `visibility`, so private helpers were being indexed and reported as public API — skewing dead-code detection, API-surface reports, and impact analysis for every Ruby project tokensave indexes. Self-initiated; no open issue tracks this.

## Changes

- `src/extraction/ruby_extractor.rs` — new directive parser: `resolve_visibility_keyword`, `resolve_class_method_keyword`, `visit_visibility_directive`, `mark_method_visibility`
- `src/extraction/ts_state.rs` — two new `ExtractionState` fields, `visibility_mode` and `singleton_method_ids`, both inert for every other language extractor
- `tests/ruby_extraction_test.rs` — 23 new tests (master has 9) covering all four directive forms
- `CHANGELOG.md` — entry under `[Unreleased] → Fixed`

Non-obvious points for the reviewer:

- `protected` maps to `Visibility::Private` — the enum has no `Protected` variant, and both mean "not public API"
- Retroactive marking (symbol-list and singleton forms) is scoped by `qualified_name`, so a same-named method in a sibling class is untouched
- `singleton_method_ids` disambiguates `def self.run` from `def run` — same kind, same qualified name, different visibility rules
- A receiverless-call check keeps a call like `policy.private` from being mistaken for a visibility directive

## Test plan

- [x] `cargo test` passes — full suite, `--no-default-features --features medium` (the tier the Ruby extractor lives in), including all 32 Ruby extraction tests
- [x] `cargo clippy` has no new warnings — `cargo clippy --workspace --all-targets`
- [x] Tested manually — `cargo fmt --all -- --check` is clean; extraction verified against fixtures covering bare directives, symbol-list form, singleton forms, and inline `private def`

## Checklist

- [x] Breaking changes documented (if any) — none; this only corrects previously-wrong `visibility` values


aovestdipaperino:
Merged. Nice piece of work — the visibility mode is saved and restored per class/module body, receiverless calls are distinguished from `policy.private`, interpolated symbols are skipped instead of guessed at, and `singleton_method_ids` correctly separates `private :foo` from `private_class_method :foo` where kind and qualified name collide. Switching `visit_method` to the `name` field also picks up operator and setter defs that the `identifier` lookup was missing. The 23 tests made the edge-case reasoning easy to verify. Thanks!

PR #286, opened 2026-07-25
fix(cpp): extract declarations wrapped in extern "C"

cescox:
## Summary
- `extern "C" void foo() {}` (single declaration) and `extern "C" { ... }` (block
  form) both parse as a `linkage_specification` tree-sitter node, which had no case
  in the C++ extractor's dispatch — anything wrapped in `extern "C"` was silently
  invisible to extraction (zero symbols, no error).
- `linkage_specification` (and its block-form `declaration_list` body) now dispatch
  through the extractor's existing generic child-walk.
- No behavior change for anything not wrapped in `extern "C"`.

## Root cause
The dispatch is a fixed match over tree-sitter-cpp node kinds; anything not listed
falls into a silent no-op. `linkage_specification` was never added, so both `extern
"C"` forms — a pattern near-universal in C/C++ interop headers — were dropped with
no error surfaced anywhere. The block form's `declaration_list` needed its own case
too: confirmed against tree-sitter-cpp's grammar that `declaration_list` only ever
appears as the body of `linkage_specification` or `namespace_definition`, and the
namespace path already unwraps its own `declaration_list` before dispatch — so
adding it here generically can't double-visit namespace members.

## Tests
- 3 new cases in `cpp_extraction_test.rs`: single declaration, block form
  (declared-only + defined member), and a regression check that plain functions are
  unaffected.
- All three verified to fail (0 functions extracted) against the pre-fix dispatch,
  pass after.
- Full suite: 2591 passed, 0 failed; `cargo fmt`/`clippy` clean.


aovestdipaperino:
Merged. Good catch on a gap that fails silently — a fixed match over node kinds with a no-op default means an unhandled kind loses symbols with nothing surfaced. I verified the `declaration_list` reasoning against `visit_namespace`, which unwraps its own body before dispatch, so the generic arm can't double-visit namespace members. Thanks for tracking down the root cause rather than just the symptom.

Note for future PRs: I rebased the branch onto master to resolve a CHANGELOG conflict, and had to approve the workflow run manually — first-time fork contributions need that, which is why CI showed nothing initially.

PR #287, opened 2026-07-25
feat(extraction): add CUDA/HIP language support (.cu, .cuh)

cescox:
## Summary
- `.cu`/`.cuh` had no registered extractor at all — every CUDA/HIP file indexed as
  zero symbols regardless of size, with no error surfaced.
- New `CudaExtractor` delegates to `CppExtractor` (same pattern as Metal, since
  CUDA is a C++ superset) — no new grammar dependency.
- `<<<grid, block>>>` kernel-launch syntax derails tree-sitter-cpp badly enough to
  drop the *enclosing function* entirely; every `<<<...>>>` span is now blanked to
  same-length spaces before parsing — the same "dialect-blanked copy" technique the
  GLSL/Godot extractor uses for its own dialect syntax.
- Ships in `full` as `lang-cuda`. Depends on #286 for `extern "C"` functions
  (the common real-world pattern for host-callable kernel wrappers) to extract.

## Tests
- 8 new cases in `cuda_extraction_test.rs`: struct/fields, device/global functions,
  kernel-launch survival (the enclosing function + both call references), device
  constant, docstring, extensions, registry dispatch.
- Verified against real-world CUDA/ROCm source (not just the fixture): three files
  that indexed as 0 symbols on master now extract 606, 12, and 84 real
  functions/structs respectively, with 0 parse errors.
- Full suite: 2599 passed, 0 failed; `cargo fmt`/`clippy` clean.


aovestdipaperino:
Merged. Delegating to `CppExtractor` rather than pulling in another grammar is the right shape for a C++ superset, and the same-length blanking of `<<<...>>>` keeps every byte offset valid so signatures and excerpts still index against real positions. Leaving an unterminated `<<<` untouched instead of blanking to EOF is a nice detail. Verifying against real CUDA/ROCm sources — not just the fixture — is what made this easy to trust.

I rebased onto master after #286 landed and adjusted one README line: CUDA/HIP now sits beside the shading languages rather than inside that list, since it's GPGPU rather than a shading language. Thanks for both PRs.

PR #291, opened 2026-07-25
fix(ruby): stop dropping class << self methods; fix receiver/name attribution bugs

janstol:
## Summary

- Ruby `class << self` (and `class << expr`) method bodies are now indexed instead of silently dropped, with singleton methods correctly attributed to the receiver they actually belong to.
- A qualified constant receiver (`class << Outer::Inner`) is now recognized as the enclosing class/module, so `private_class_method` can reach methods defined that way.
- Compact Ruby class/module definitions (`class Outer::Inner`, `module A::B`) are indexed by their real name instead of `<anonymous>`.

## Motivation

`class << self` is an extremely common Rails idiom for declaring class methods, and the extractor's `visit_node` dispatch had no arm for the `singleton_class` node kind it produces — every method inside one was silently absent from the graph (no node, no `Contains` edge, no call sites, no callers), which threw off `dead_code`, `module_api`, `impact`, and `callers`.

Review of that fix surfaced five further receiver-attribution bugs (all fixed in the first commit) plus a sixth found in a later review pass: `is_enclosing_receiver` didn't recognize a *qualified* constant receiver (`class << Outer::Inner`), a `scope_resolution` node, so it always fell through to `Foreign` even when it named the exact class/module it appeared inside — even though Ruby genuinely privatizes methods defined that way.

Verifying that sixth bug exposed a second, independent, pre-existing defect (not introduced by this branch): `visit_class`/`visit_module` located a class/module's name by searching its *direct* children for a `constant` node, but tree-sitter-ruby types the name field as `choice($.constant, $.scope_resolution)` — a compact `class Outer::Inner` has no direct `constant` child at all, so it indexed as `<anonymous>`, and so did every method inside it (`file::<anonymous>::foo`). This hits every Rails app using the common `class Admin::UsersController` compact form, and it's also why the qualified-receiver fix alone can't cover the compact case: no receiver text can ever equal `<anonymous>`.

## Changes

- `src/extraction/ruby_extractor.rs`
  - `is_enclosing_receiver`'s `"constant"` arm now also accepts `"scope_resolution"`, matched against the enclosing class/module scope chain via a new `matches_enclosing_scope_path` helper (relative-suffix matching at node-stack-entry granularity, plus a `::`-anchored absolute-path form).
  - `visit_class`/`visit_module` now read `node.child_by_field_name("name")` instead of searching direct children for a `"constant"` node, so compact definitions resolve to their full path instead of `<anonymous>`.
- `tests/ruby_extraction_test.rs` — 8 new tests covering the qualified-receiver forms (nested, partial-relative, absolute, and each with a negative counterpart) and the compact-name fix (class, module, and the compact-receiver interaction between both fixes).
- `CHANGELOG.md` — extended the existing `class << self` `[Unreleased]` entry with item (6), and added a new `### Fixed` bullet for the compact-name defect.

## Test plan

- [x] `cargo test --no-default-features --features medium --test ruby_extraction_test` — 66 passed (63 → 66)
- [x] `cargo test --no-default-features --features medium` (full suite) — all green
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --workspace --all-targets` — clean

## Checklist


aovestdipaperino:
Merged. Thanks @janstol — the `SingletonScope`/`VisibilityTarget` split is a genuine correctness win over the old boolean, and the negative-case tests made it easy to trust.

PR #295, opened 2026-07-25
fix(ranking): down-rank documentation files

map588:
# fix(ranking): down-rank prose documentation files

## Summary

- Documentation files (`.md`, `.markdown`, `.mdx`, `.rst`, `.adoc`, `.txt`) get a 0.3× factor in `path_rank_multiplier`, so code outranks doc headings that merely share vocabulary with the query.

## Motivation

Markdown headings are indexed as `Module` nodes with `Pub` visibility, so the kind (1.2×) and visibility (1.5×) boosts hand doc headings a **1.8× multiplier over neutral code**. In practice a symbol search can rank `CHANGELOG.md` above the module that defines the symbol. Observed on this repo: `tokensave_search "queryignore"` surfaced `CHANGELOG.md` as a top hit ahead of `src/config.rs`, which defines the feature.

## Changes

- `src/context/ranking.rs` — `is_doc_file` (extension match, case-insensitive) checked first in `path_rank_multiplier`, returning 0.3× — between the existing fixture (0.1×) and test (0.4×) factors. Shared by both `tokensave_search` and `tokensave_context` ranking, so one change covers both paths.
- Non-obvious: proportional, not a filter — an exact heading match still surfaces, and `path_include: ["docs"]` overrides when docs are the target. Extension-only matching: a directory named `docs/` containing code is unaffected.
- Tests: doc factor below the test factor across all six extensions; a doc `Module`+`Pub` heading loses to a private function with equal FTS score in `rerank_candidates`; `mdbook/render.rs` stays neutral.

## Test plan

- [x] `cargo test` passes
- [x] Tested manually: `tokensave tool search "queryignore"` on this repo — top 10 previously led with `CHANGELOG.md`, now all `src/`

## Checklist

- [x] Breaking changes documented (if any) — none; ranking-only, doc files remain indexed and reachable


aovestdipaperino:
Thanks @map588 — the core idea is right and the `is_doc_file` implementation is clean (last `.`-segment of the normalized path, so `mdbook/render.rs` correctly reads as `rs` and extensionless paths fall through; I could not construct a false positive). Holding on a few things before merge.

**1. The other consumer is the one that matters here.** `path_rank_multiplier` has two call sites. In `rerank_candidates` it is multiplicative alongside the other boosts, which is what your justification reasons about. But `src/tokensave/query.rs:66` applies it to an **additive** score that already includes a `+10.0` exact-name bonus. There `0.3` scales the entire score, so an exact heading match in a `.md` drops from 10+ to ~3 and can land below partial matches in code. That path is where the `tokensave_search "queryignore"` symptom you opened with actually lives, and it is where the penalty bites hardest. Could you add a test against the `query.rs` path and sanity-check the factor there? 0.3 may still be right, but right now nothing measures it where it counts.

**2. Ordering vs. the vendor branch.** The doc check is inserted before the vendor check in `path_rank_multiplier` (`src/context/ranking.rs:132`), but the function's own rustdoc says "Vendor classification wins over both non-production and app classification." `node_modules/foo/README.md` now goes 0.4 → 0.3. Numerically harmless, but either move the check below the vendor branch or update the doc comment — right now the code and its contract disagree.

**3. The `path_include` claim is not accurate.** The PR body and the CHANGELOG entry both say `path_include: ["docs"]` overrides the penalty when docs are the target. It does not — `path_include` filters candidates, it does not bypass the multiplier. The *outcome* is fine (all surviving candidates take the same 0.3, so relative order is preserved, and likewise in docs-only repos), but the stated mechanism is wrong and I'd rather not have it enter the CHANGELOG that way.

**4. Consider dropping `txt`.** `.md/.rst/.adoc/.mdx` are unambiguously prose. `.txt` is often fixture or data content whose "headings" are not prose at all. Weakest entry in the list.

**5. Format CI is red** — purely cosmetic, `cargo fmt --all` fixes it (`is_doc_file` and two test assertions want rewrapping).

Minor: `test_path_rank_multiplier_docs_below_tests` asserts `path_rank_multiplier(path) < path_boost(...)`, comparing values from two different functions that get multiplied together at the call site. It encodes 0.3 < 0.4 correctly but is a confusing invariant to leave behind. Removing the `README.md`/`docs/guide.md` lines from `test_path_rank_multiplier_neutral_is_baseline` was right.

Happy to merge once 1-3 and 5 are addressed.

map588:
All addressed in 13eadcb.

(1): writing the test revealed the query.rs path is tier-guarded — headings are Module nodes (tier 3) and the sort is tier-first, so an exact heading match ranks below partial code matches with or without the penalty; the factor only orders within tier 3, where all doc candidates share the 0.3× and relative order is preserved. Test pins both invariants, factor unchanged. 

(2) vendor branch now runs first, doc comment updated. 

(3) CHANGELOG reworded — path_include filters candidates, it doesn't bypass the multiplier. 

(4) txt dropped. 

(5) formatted. Also replaced the cross-function assert from the minor note with direct value asserts.

aovestdipaperino:
Merged. Thanks @map588 — you'd already addressed all five points before I got back to this, and the additive-path test is better than what I asked for: pinning that the tier-first sort (headings are `Module`, tier 3; definitions tier 0) is what guarantees the ordering, so no value of the doc factor can regress it. I verified that in `kind_tier` independently — my original concern about the +10 bonus being scaled below partial code matches was overstated for that path, and your framing is the correct one. Dropping `.txt` and moving the check below the vendor branch were the right calls too.

Issue #258, opened 2026-07-21
tokensave_search literal mode ignores path_include filtering


not-antoni:
### Problem

`tokensave_search` can return literal matches from unrelated files even when `path_include` contains a specific source path.

### Generic reproduction

```json
{
  "query": "some_literal",
  "literal": true,
  "path_include": ["src/target-file.js"],
  "limit": 20
}
```

When the same literal exists elsewhere, results outside `src/target-file.js` are included.

### Expected

Literal-mode results should be filtered to paths containing one of the requested `path_include` values, consistent with the tool schema. `path_exclude` should continue to take precedence.

This report intentionally omits project names, source text, and other proprietary details.

aovestdipaperino:
Fixed in v7.6.0. Literal-mode `tokensave_search` now applies `path_include`/`path_exclude` with the same substring filtering as ranked search (and `path_exclude` takes precedence), matching the tool schema. Thanks for the clear, self-contained repro. 🙏

Issue #116, opened 2026-06-15
Project-level ignore config for graph/query (gitignore-like)


aovestdipaperino:
Split from #112.

## Goal
Optional project-level ignore config for the graph/queries (gitignore-like) so known-noise paths can be excluded from indexing and/or query results persistently, without passing `path_exclude` every call.

## Motivation (from #112)
Users want to declare noise paths once.

## Proposed behavior
- A project config (e.g. under `.tokensave/`) listing ignore globs applied at index and/or query time.
- Define interaction with existing `.gitignore` handling (`tokensave gitignore`).

Part of #112.

aovestdipaperino:
Implemented and released in **v6.3.3** (#112).

PR #296, opened 2026-07-25
fix(search): apply queryignore to literal mode

map588:
# fix(search): apply queryignore to literal mode

## Summary

- Literal search (`tokensave_search` with `literal: true`) now honors `.tokensave/queryignore`, matching the ranked search path.

## Motivation

The literal branch of `handle_search` returns before the ranked path's queryignore filter runs, so paths a project explicitly suppressed via `.tokensave/queryignore` (#116) still appear in literal matches — e.g. literal hits from a vendored tree or `CHANGELOG.md` that ranked search correctly suppresses. Same gap shape as the `path_include`/`path_exclude` literal-mode fix in #258.

## Changes

- `src/mcp/tools/handlers/graph.rs` — `handle_literal_search` filters the indexed file list with the project queryignore patterns before scanning, right after the existing `path_include`/`path_exclude` filter.
- `tests/mcp_handler_test.rs` — regression test: a queryignore pattern matching the fixture file drops the literal match that `test_search_literal_finds_string_in_body` otherwise finds.

## Test plan

- [x] `cargo test` passes
- [x] Tested manually: on the tokensave repo itself, literal queries for strings present in `CHANGELOG.md` stop matching once the path is listed in `.tokensave/queryignore`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), not SemVer.
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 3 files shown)


aovestdipaperino:
Merged. Thanks @map588 — good catch that the literal path returned before the ranked path's queryignore filter, and the fix mirrors the ranked semantics exactly.

PR #297, opened 2026-07-25
feat(context): show docstring first line for entry points

map588:
# feat(context): show docstring first line for entry points

## Summary

- `tokensave_context` entry points show the first non-empty docstring line beneath the signature.

## Motivation

Entry points previously listed only name, kind, location, and signature. Docstrings are already extracted and indexed (`extract_docstrings` defaults to true) but never surfaced in context output — yet the one-line summary frequently answers the query outright, saving the follow-up `include_code` fetch or `tokensave_body` call. One line per entry point is a small token cost against a whole avoided round-trip.

## Changes

- `src/context/formatter.rs` — `format_context_as_markdown` prints the first non-empty docstring line (trimmed) after the signature line; entries without a docstring are unchanged.
- Test asserts the first line appears and later lines don't.

## Test plan

- [x] `cargo test` passes
- [x] Tested manually: `tokensave tool context` on this repo shows doc summaries under each documented entry point

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 
 ## [Unreleased]
 
+### Added
+- **`tokensave_context` entry points now show the first line of the symbol's docstring** beneath the signature. The one-line summary often answers the query outright, saving the follow-up code fetch; entries without a docstring are unchanged.
+
 ### Fixed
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Merged. Thanks @map588 — the one-line summary earns its space, and skipping empty leading lines was the right detail. I rebased for the CHANGELOG collision. Worth a follow-up: there's no length cap, so a single-line block docstring lands in full in every response, which cuts against the token argument.

PR #298, opened 2026-07-25
feat(context): point truncated code blocks at tokensave_body

map588:
# feat(context): point truncated code blocks at tokensave_body

## Summary

- Code blocks truncated at `max_code_block_size` now end with `... [truncated — full body: tokensave_body node_id=<id>]` instead of a bare ellipsis.

## Motivation

A truncated snippet is a dead end for the calling agent: the `...` gives no indication of how to fetch the rest, so the agent either re-derives the symbol by name (an extra search round-trip) or reads the file directly (defeating the token savings). The node id is already on hand in `extract_code_blocks`; embedding it makes the follow-up a single `tokensave_body` call.

## Changes

- `src/context/builder.rs` — `extract_code_blocks` truncation arm appends the handle; the line-boundary preference for the cut point is unchanged.

## Test plan

- [x] `cargo test` passes
- [x] Tested manually: `tokensave tool context` on this repo with a long entry point shows the handle, and the emitted `node_id` resolves via `tokensave_body`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 ### Added
 - **`tokensave_context` entry points now show the first line of the symbol's docstring** beneath the signature. The one-line summary often answers the query outright, saving the follow-up code fetch; entries without a docstring are unchanged.
 
+### Changed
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Merged. Thanks @map588 — turning the dead-end ellipsis into a fetchable handle is a real improvement. I rebased for the CHANGELOG and added the missing test coverage: pulled the truncation into a pure `truncate_code_block` so it's testable without a ContextBuilder, and pinned the handle, the line-boundary cut, verbatim pass-through, and a multi-byte char at the cutoff. Note the suffix pushes the emitted block ~50 bytes past `max_code_block_size`, up from 3 — fine unless a caller treats that as a hard bound.

PR #299, opened 2026-07-26
feat(ruby): extract include/prepend/extend as implements edges

janstol:
## Summary

- Ruby `include` / `prepend` / `extend` of a named module now emit a resolvable `Implements` edge from the including class or module — these keywords weren't handled at all, so no Ruby composition was visible to the graph.
- The resolver gains a Ruby-only `Implements → NodeKind::Module` arm, since Ruby mixin targets are modules and every other language's `Implements` allow-list excludes `Module`.
- `visit_node` now recurses through statement containers (`if`, `case`, `begin`/`rescue`, loops, modifiers), so a mixin — or any `def`/`class`/`module`/constant — nested in one is no longer silently dropped.

## Motivation

`include Comparable` parsed as an ordinary `call` node and was discarded. The consequence was that `dead_code`, `impact`, `module_api`, and `type_hierarchy` saw *no* Ruby composition whatsoever: a module's methods never appeared as part of the including class's API, and a module included by a dozen classes showed zero inbound relationships. For Rails codebases that covers `ActiveSupport::Concern`, `Comparable`, `Enumerable`, and every `app/models/concerns/*` file.

## Changes

Two commits, each green on its own:

- **`fix(resolution): allow Ruby module targets for implements refs`** — `kind_compatible` now takes the whole `&UnresolvedRef` (it needs `file_path` to know the language, not just the `EdgeKind`) and gains a `EdgeKind::Implements if lang_from_path(…) == "ruby"` arm matching **exclusively** `NodeKind::Module`. Exclusive rather than additive on purpose: Ruby itself rejects `include SomeClass` with `TypeError: wrong argument type Class (expected Module)`, and an additive rule would let a project with both a `class Foo` and a `module Foo` bind to whichever sorts first in the suffix index. It would also — incorrectly — let `Extends`/`DerivesMacro` target a `Module`, which Ruby never emits.
- **`feat(ruby): extract include/prepend/extend as implements edges`** — new `visit_mixin_directive` emits one `Implements` ref per `constant`/`scope_resolution` argument (`include A, B` → two; `include ActiveSupport::Concern` resolves the scoped name). Deliberately *not* handled, rather than fabricating a ref: dynamic arguments (`extend self`, `include some_variable`), an explicit receiver (`mod.include Bar`), and top-level `include` (which mixes into `Object`, with no class/module node to attach to).
- **Shared dispatch arm with #284's visibility handler.** Both `visit_visibility_directive` and `visit_mixin_directive` now run from the same `"identifier" | "call" | "method_call"` arm. They compose safely: each early-returns on an explicit receiver, they're gated on disjoint method-name sets (`private`/`protected`/`public`/`*_class_method` vs `include`/`prepend`/`extend`), and `visit_mixin_directive` early-returns on an `identifier` node (no `method` field). Calling both is a no-op for any name neither owns.
- **Statement-container recursion, and its one knock-on effect.** Containers wrap statements without opening a definition scope, so the enclosing class/module stays the parent and `include Foo if enabled?` is extracted as if written directly in the body — the "static possibility" reading. Because the same arm carries visibility directives, a `private` reached only through a conditional now also applies past the container's `end`. That matches Ruby's own runtime scoping and is asserted by a new test (`test_ruby_visibility_directive_in_conditional_applies_after_end`) rather than left implicit. Method bodies are still not traversed this way — `visit_method` routes those through `extract_call_sites`.

Out of scope (each is its own change): `class << self` bodies, `included do … end` / `class_eval do … end`, a distinct `EdgeKind` separating `extend`'s singleton semantics from `include`/`prepend`, `attr_*`, `alias`/`alias_method`, `define_method`, `module_function`, and modules reopened across files.

PR #291 (`feat/ruby-singleton-class`) touches the same `visit_node` dispatch. I re-ran the merge against this branch: `ruby_extractor.rs` merges cleanly, and only `tests/ruby_extraction_test.rs` and `CHANGELOG.md` conflict, both because each PR appends at the same spot. The two are independent — whichever lands second needs a trivial rebase, which I'll handle.

## Test plan

- [x] `cargo test` passes
- [x] Tested manually (described below)

22 new tests — 16 in `tests/ruby_extraction_test.rs`, 6 in `tests/resolution_test.rs`. Beyond the happy path they cover the negative cases (`extend self`, explicit receiver, top-level `include`, `include` inside a method body), the container cases (`if` modifier, `unless` modifier, `if`/`else` body, `begin`/`rescue`), and the resolver's kind gate from both directions: a Ruby `Implements` ref must *not* reach a `Class`, a non-Ruby `Implements` ref must *not* reach a `Module`, `Extends` must not reach a `Module`, and a module is preferred over a same-named class.

Verified on the default/full feature set and on `--no-default-features --features medium`, plus `git rebase origin/master --exec 'cargo test …'` so each commit is green in isolation. `cargo clippy --workspace --all-targets` surfaces one pre-existing `map_unwrap_or` pedantic warning in `src/mcp/tools/handlers/git.rs:850`, untouched by this branch.

Also built the binary (`--features medium`) and ran it against a small Rails-shaped fixture outside the repo: a `module Trackable` in `app/models/concerns/trackable.rb` and a `class Order` that `include`s it and calls one of its methods. After `init`/`sync`, the raw edges table shows exactly `implements | Order (class) | Trackable (module)`, and `tokensave_type_hierarchy` on `Trackable`'s node prints `Trackable (module) -- ...` followed by `|- implements Order (class) -- ...` — the mixin edge resolves and surfaces through the existing hierarchy tool with no changes needed there.

## Checklist

- [x] Breaking changes documented (if any) — none. The `kind_compatible` signature change is private to `src/resolution/resolver.rs`; the Ruby `Implements` arm only *adds* resolutions that previously failed, and every other language keeps its existing allow-list.


aovestdipaperino:
Reviewed and looks good — holding only on a rebase. #291 just landed, which collides with this in `tests/ruby_extraction_test.rs` (both append at the same spot; `src/extraction/ruby_extractor.rs` and the CHANGELOG auto-merge fine). Could you rebase on master, @janstol? Happy to merge once CI is green on the rebased head.

One thing worth a test while you're in there: the new `body_statement` arm also enables extracting defs inside arbitrary `do…end` blocks (`included do`, `class_eval do`, RSpec `describe`), which previously fell through undispatched. That seems like the right behavior, but it's a wider extraction-surface change than the PR body suggests and nothing currently pins it.

aovestdipaperino:
Merged. Thanks @janstol — the resolver gate is the part I'd have most expected to go wrong, and making the Ruby `Implements` arm exclusive to `Module` (rather than additive) is the right call; the both-directions tests made it easy to verify. I rebased twice as master moved.

Follow-up worth a test: the new `body_statement` arm also enables extracting defs inside arbitrary `do…end` blocks (`included do`, `class_eval do`, RSpec `describe`), which previously fell through undispatched. Looks like the right behavior, but nothing pins it.

Issue #301, opened 2026-07-26
Namespace the remaining tokensave-owned environment variable


rNoz:
The runtime environment-variable inventory has one project-owned outlier: `DISABLE_TOKENSAVE`. Every other implemented tokensave-owned variable already uses the `TOKENSAVE_` namespace, including `TOKENSAVE_UPDATE_CHECK`, `TOKENSAVE_AUTO_TRACK`, `TOKENSAVE_DISABLE_GREP_HOOK`, `TOKENSAVE_WORKERS`, `TOKENSAVE_DISABLE_SUBPROCESS`, and `TOKENSAVE_WORKER_TOKEN`.

`DISABLE_TOKENSAVE=true` shipped in v4.3.8 through issue #19. It makes `tokensave serve` exit 0 before database, watcher, or MCP transport initialization, and the maintainer explicitly documented it for per-project MCP configuration. Renaming it without compatibility would silently re-enable tokensave in projects that disabled it to avoid high memory usage.

### Proposed behavior

- Add `TOKENSAVE_DISABLE_SERVER=true` as the canonical server opt-out.
- Keep `DISABLE_TOKENSAVE=true` as a deprecated compatibility alias with no scheduled removal.
- Let an explicit `TOKENSAVE_DISABLE_SERVER=false` override an inherited legacy disable.
- If the new value is absent or unrecognized, preserve a legacy `DISABLE_TOKENSAVE=true` disable.
- Preserve the legacy variable's current exact lowercase `true` contract. The canonical variable recognizes exact lowercase `true` and `false`.
- Do not print a runtime deprecation warning because MCP clients can surface stderr from the intentional early exit as an error.

### Future guard

Add a test that parses first-party Rust sources with the repository's bundled tree-sitter Rust grammar and inspects environment-variable API calls. Tokensave-owned names must start with `TOKENSAVE_`; operating-system, Cargo, Git, and agent-owned contracts require an explicit allowlist rationale. `DISABLE_TOKENSAVE` remains the sole legacy exception. Document the policy in CONTRIBUTING so future variables fail tests instead of introducing another unnamespaced public API.

### Acceptance

- New and legacy names both exit `serve` cleanly before project initialization.
- Conflicting and malformed values follow the documented canonical-authoritative precedence.
- Existing legacy configurations keep working unchanged.
- README uses `TOKENSAVE_DISABLE_SERVER` in examples and labels `DISABLE_TOKENSAVE` deprecated.
[...]

Issue #19, opened 2026-04-14
Suggestion: I'd like to choose when to start tokensave in my project


alkampfergit:
I've enabled tokensave, now every project starts its own version of tokensave and this is good. The only problem is that I have big project, so for some instances the usage of RAM is  almost 1G.

It would be nice (maybe I'm missing something) to enable tokensave per-project with some settings.

<img width="795" height="159" alt="Image" src="https://github.com/user-attachments/assets/78192d2a-ede6-4e31-af2e-36781fa6bc3b" />

aovestdipaperino:
I will keep the bug open, but have you tried 
`tokensave daemon --enable-autostart` ?
It should start as a service for automatic syncing. But I will take a look at RAM usage, there shouldn't be any RAM usage at all when idle.
Thanks for reporting.

alkampfergit:
Four vscode opened, each one has code and codex, two of them are idle at this moment. If you want I can create a minidump and upload somewhere :)

<img width="813" height="242" alt="Image" src="https://github.com/user-attachments/assets/068e27f2-9017-42f7-b6a4-4dcb96d34b8c" />

aovestdipaperino:
How big is the largest tokensave DB size?

aovestdipaperino:
Fixed in v4.3.8.

You can now set `DISABLE_TOKENSAVE=true` in the MCP server environment to opt out per-project. The `serve` command exits cleanly (exit 0) without initializing the database, watcher, or transport — so the host won't retry or show errors.

**Per-project usage** (recommended): add the env var in your project's `.claude/settings.json` so it only affects that workspace:

```json
{
  "mcpServers": {
    "tokensave": {
      "command": "tokensave",
      "args": ["serve"],
      "env": {
        "DISABLE_TOKENSAVE": "true"
      }
    }
  }
}
```

You can also set it globally via shell (`DISABLE_TOKENSAVE=true claude`), but that disables tokensave for every project in the session.

PR #302, opened 2026-07-26
feat(config): namespace server disable env

rNoz:
## Summary

- add `TOKENSAVE_DISABLE_SERVER` as the canonical MCP server opt-out
- preserve `DISABLE_TOKENSAVE=true` as a deprecated compatibility alias
- enforce the `TOKENSAVE_` namespace for future first-party environment variables
- Implements proposal of namespacing tokensave env vars from now on (https://github.com/aovestdipaperino/tokensave/issues/301)

## Motivation

`DISABLE_TOKENSAVE=true` was introduced through #19 as a per-project workaround for repositories where a tokensave server used too much memory. It is the only implemented tokensave-owned environment variable without the `TOKENSAVE_` prefix.

A hard rename would silently re-enable servers for users relying on that safeguard. This change adds a namespaced spelling without breaking the existing public behavior.

## Changes

- `tokensave serve` reads `TOKENSAVE_DISABLE_SERVER` before the legacy alias.
- Exact `TOKENSAVE_DISABLE_SERVER=true` disables the server.
- Exact `TOKENSAVE_DISABLE_SERVER=false` explicitly enables it, even if an inherited `DISABLE_TOKENSAVE=true` is present.
- An absent or unrecognized canonical value falls back to exact `DISABLE_TOKENSAVE=true`.
- The early exit remains before project database, watcher, and MCP transport initialization.
- No runtime deprecation warning is printed because MCP clients can surface stderr from this intentional clean exit as an error.
- A tree-sitter-based integration test scans `src`, `benches`, optional `examples`, and `build.rs`. Tokensave-owned literal names must use `TOKENSAVE_`; operating-system, Cargo, Git, and agent-owned contracts are narrowly allowlisted with rationale.
- README, CONTRIBUTING, and CHANGELOG document the canonical name, compatibility alias, and future naming policy.

### Resolution matrix

| `TOKENSAVE_DISABLE_SERVER` | `DISABLE_TOKENSAVE` | Result |
| --- | --- | --- |
| `true` | any value | disabled |
| `false` | any value | enabled |
| absent or unrecognized | `true` | disabled |
| absent or unrecognized | any other value | enabled |

Values are case-sensitive to preserve the legacy exact lowercase contract.

## Test plan

- [x] `cargo test --bin tokensave startup_tests::`
- [x] `cargo test --test serve_disable_test`
- [x] `cargo test --test env_var_namespace_test`
- [x] `cargo test` passes, 2,654 tests
- [x] Tested manually with isolated scratch projects

The binary proof runs both supported spellings against separate empty directories:

```bash
TOKENSAVE_DISABLE_SERVER=true target/debug/tokensave serve --path "$canonical_scratch"
DISABLE_TOKENSAVE=true target/debug/tokensave serve --path "$legacy_scratch"
```

Both commands exit 0 and neither creates a `.tokensave` directory. The integration test also isolates the child process's home, working directory, and standard streams, and includes a negative control so a populated global project database cannot produce a false pass.

Independent Claude Opus 4.8 review found and prompted that isolation fix, then approved the corrected subprocess proof and final tree-sitter policy guard.

## Checklist

- [x] Compatibility behavior documented


rNoz:
Working on the fix for Windows (CI Test caught it) => Done :)

aovestdipaperino:
Merged — thanks @rNoz. Canonical `TOKENSAVE_DISABLE_SERVER` with `DISABLE_TOKENSAVE` kept as a deprecated alias is exactly the compatible shape for a safeguard people already depend on, and the tree-sitter policy guard should keep future first-party env vars namespaced without needing review vigilance.

Issue #303, opened 2026-07-26
Warn when `install` persists a Cargo build-output binary path


rNoz:
### tokensave version

7.6.2 (`upstream/master` at `0d850e5`)

### Operating system

macOS

### Installation method

Built from source

### What happened?

`which_tokensave()` prefers `std::env::current_exe()` and writes that absolute path into agent hook and MCP command entries. For a source build, the persisted path is commonly `<checkout>/target/{debug,release}/tokensave`.

Issue #146 and PR #147 fixed the Homebrew form of this problem by replacing a versioned Cellar path with Homebrew's stable `bin` symlink. Cargo build output has no equivalent managed symlink.

When `install`, `reinstall`, or automatic agent resync runs from a Cargo build inside a temporary worktree, the worktree-local path is written into persistent global config. Removing the worktree, or running `cargo clean`, leaves every owned hook and MCP entry pointing at a missing executable:

```
UserPromptSubmit hook error
Failed with non-blocking status code: ENOENT: no such file or directory,
posix_spawn '/Users/<u>/projects/<repo>/.claude/worktrees/agent-XXXX/target/release/tokensave …'
```

A broken PreToolUse command also disables the guardrail until the user notices the non-blocking error.

### Steps to reproduce

```bash
# 1. Build in a disposable worktree.
scratch="$(mktemp -d)/tokensave-wt"
git worktree add "$scratch" HEAD
(cd "$scratch" && cargo build --release)

# 2. Use that binary for a global install.
"$scratch/target/release/tokensave" install --agent claude

# 3. Observe the embedded path.
jq -r '.hooks.PreToolUse[].hooks[].command' ~/.claude/settings.json
#   -> .../tokensave-wt/target/release/tokensave

# 4. Remove the worktree (normal cleanup).
git worktree remove "$scratch" --force

# 5. Every session now errors on that hook (ENOENT); the guardrail is dead.
```

### Error output

```shell
UserPromptSubmit hook error
Failed with non-blocking status code: ENOENT: no such file or directory,
posix_spawn '/.../tokensave-wt/target/release/tokensave ...'
```

### Component

Agent integration (Claude, Cursor, etc.)

### Additional context

Global config writes should not silently persist a path under Cargo build output. When preserving the invoked source-build binary, tokensave should emit one actionable warning that:

- names the persisted path,
- explains that worktree removal or `cargo clean` will break the integration,
[...]

Issue #146, opened 2026-06-18
Homebrew install issue with hooks


cavanaug:
### tokensave version

6.4.2

### Operating system

macOS

### Installation method

Homebrew (brew)

### What happened?

When running tokensave install it creates a versioned reference to the tokensave binary in the config files for hooks.   This then breaks then next time there is a brew upgrade.

For example in claude settings.json

```
      {
        "hooks": [
          {
            "args": [
              "hook-pre-tool-use"
            ],
            "command": "/home/linuxbrew/.linuxbrew/Cellar/tokensave/6.4.2/bin/tokensave",
            "type": "command"
          }
        ],
        "matcher": "Agent|Grep|Bash"
      }
```

```
$ which tokensave
/home/linuxbrew/.linuxbrew/bin/tokensave

$ readlink -e /home/linuxbrew/.linuxbrew/bin/tokensave
/home/linuxbrew/.linuxbrew/Cellar/tokensave/6.4.2/bin/tokensave
```

So the next new version of tokensave (or after a brew cleanup) claude will not be able to find tokensave 

### Steps to reproduce

1.
2.
3.

### Error output

```shell

```

### Component

None

PR #304, opened 2026-07-26
fix(install): Warn when agent setup persists a Cargo build binary

rNoz:
## Summary

- Warn when global `install`, successful `reinstall`, or an automatic agent resync that actually ran persists a tokensave executable under Cargo `target/debug`, `target/release`, or the cross-compile `target/<triple>/{debug,release}` layout.
- Preserve the existing exact-invoked-binary selection, Homebrew stable-path rewrite, project-local behavior, existing MCP-command preservation, and resync version-marker semantics.
- Cover the path classifier with focused unit tests and a real sandboxed global Claude install that asserts both the warning and the unchanged persisted command.
- Solves https://github.com/aovestdipaperino/tokensave/issues/303

## Motivation

`which_tokensave()` intentionally prefers `current_exe()` so a source build installs the exact binary the user invoked. In a temporary worktree that path commonly lands under `target/{debug,release}`; removing the worktree or running `cargo clean` then leaves global hooks and MCP entries pointing at a missing executable. The Homebrew form was fixed by #146/#147 because Homebrew provides a managed stable symlink, but a generic Cargo build has no equivalent.

Automatically substituting an arbitrary `PATH` entry could silently install an older binary instead of the source build the user selected. A bare command depends on the hook runner inheriting the install shell's `PATH`, a hard refusal breaks supported source-build workflows, and a managed copy would require a new binary-update lifecycle. This change follows the narrow guardrail used by comparable hook installers: preserve behavior, but make the persistence hazard explicit and actionable.

## Changes

- `src/agents/mod.rs`: add a pure whole-component classifier and warning message for Cargo profile paths, including cross-compile target triples and stable-path near misses.
- `src/main.rs`: emit only after global install state is written, after a successful reinstall with tracked agents, and after automatic resync when `ResyncOutcome.ran` is true. `which_tokensave()` itself stays silent, so ordinary source-build commands do not gain startup noise.
- `tests/install_warning_test.rs`: copy the real test binary into a scratch `target/debug` path, run `install --agent claude --git-hook no` with a sandboxed home, and assert the actionable warning plus the exact persisted path.

## Test plan

- [x] `cargo test --lib which_tokensave_tests` passes, 6 passed.
- [x] `cargo test --test install_warning_test` passes, 1 passed.
- [x] `cargo test --test mcp_handler_test` passes, 179 passed after one non-reproducible SIGABRT during the first concurrent full-suite run.
- [x] `cargo test` passes on rerun, including 588 library tests and all integration suites.
- [x] `cargo clippy --workspace --all-targets` exits 0; upstream's existing `clippy::map_unwrap_or` warning at `src/mcp/tools/handlers/git.rs:850` remains unchanged.
- [x] Sandboxed end-to-end install emits the warning, writes the scratch `target/debug/tokensave` path into all three Claude hook commands, and leaves those commands dangling after that exact binary is removed.
- [x] `git overlay check worktree-ephemeral-bin` reports READY on commit `ba1cdf8`.

```text
warning: agent config references Cargo build output:
  <scratch>/target/debug/tokensave
  `cargo clean` or removing its worktree will break tokensave hooks and MCP servers.
  Re-run `tokensave install` from a stable `cargo install`, Homebrew, or release binary.

.claude/settings.json:
  "command": "<scratch>/target/debug/tokensave"
  "command": "<scratch>/target/debug/tokensave"
  "command": "<scratch>/target/debug/tokensave"

binary_exists_after_removal=no
```

## Checklist


aovestdipaperino:
Merged — thanks @rNoz. The narrow guardrail was the right call here: keeping `current_exe()` selection intact and making the persistence hazard explicit avoids silently swapping in an older PATH binary, and gating on `ResyncOutcome.ran` keeps ordinary source-build commands quiet.

Issue #306, opened 2026-07-26
Resolution path still materialises the whole node graph at once


aovestdipaperino:
Follow-up from #253, which cut the resolver's cache copies (#305) but left two allocation sources untouched:

1. **The caller materialises the full \`Vec<Node>\` before resolution.** \`get_all_nodes()\` is loaded in one shot in \`src/tokensave/indexing.rs\` (three sites) and stays resident for the whole resolution pass. Chunking or streaming it would remove the last graph-sized allocation from the sync path.
2. **Several analysis handlers independently call \`get_all_nodes()\`** (health, analysis, redundancy), so a single tool call on a large project can re-trigger a graph-sized spike. Converting these to SQL aggregates or paged reads would bound them.

Neither is a leak; both are peaks proportional to graph size. Lower priority than #253 was, but they are what stands between the current state and a genuinely bounded footprint on very large projects.

Also unresolved from that thread: a reporter saw orphaned PID-1 processes at ~9.2 GiB burning 20% CPU, which the #305 analysis does not explain and which may be an unrelated failure mode.

aovestdipaperino:
Filed #409 for the design decision this depends on, tagged `PTaL`.

Short version: item 1 here — chunking or streaming `get_all_nodes()` in the resolution path — **cannot be done as a call-site change.** Every cache `ReferenceResolver` builds borrows from the node slice for its lifetime, and more fundamentally resolution needs a *global* name index: binding a cross-file reference means looking up a name that may be defined anywhere in the graph, by short name, qualified name, or any `::` suffix. Chunk the input and a reference silently fails to find a target outside its chunk — a correctness regression dressed as a memory win.

So the peak isn't an accident of how the caller loads nodes; it's a property of resolving names against a whole graph in one pass. #409 lays out four options (interned arena, SQL-backed name index, name sharding, or accept it and bound only the trigger) with the trade-offs, and the questions I need answered before writing code — chiefly whether *bounded* is actually the goal or whether much-smaller-but-still-linear is enough.

#407 is filed against item 1 as a partial: it drops `docstring` and `signature`, which resolution never reads and which are unbounded TEXT, cutting roughly a third of node string payload on this repo. It does not bound anything.

[...]

Issue #253, opened 2026-07-20
`tokensave serve` (MCP stdio server) leaks memory to 8GB+ RSS, causes system-wide OOM crashes


himanishpuri:
**Version:** 7.4.1 (installed via `tokensave upgrade`, latest release at time of writing; changelog for 7.4.2 only touches `tokensave_health` output formatting, doesn't appear to address memory)

**Environment:** Fedora Linux (Dell Inspiron 14 5430), stock kernel, Claude Code CLI, tokensave installed as an MCP server via `tokensave install` (`tokensave serve` over stdio).

**Problem:**
The `tokensave serve` process's RSS grows unbounded during a long-running Claude Code session and eventually consumes most of system RAM, triggering the kernel OOM killer and taking down the whole desktop session (not just the process itself — cgroup-wide cascade kill of the parent terminal scope).

Observed twice across separate boots/sessions:
- Session 1 (prior to upgrading 7.0.2 → 7.4.1): `tokensave serve` was the dominant RSS consumer at OOM time, identical victim process across repeated kernel OOM-kill events in `journalctl -k`, despite `oom_score_adj=200` (deprioritized) — it was still the largest consumer by a wide margin.
- Session 2 (today, on 7.4.1): caught it live — `ps` showed `tokensave serve` at **~8.0GB RSS** (50%+ of 15GB total system RAM), with zram (8G) fully saturated and 4GB+ pushed into disk swap, system down to ~230MB free before I manually `kill -9`'d it. Memory recovered immediately after the kill (RSS 0, ~9GB available).

No crash occurred when running Claude Code with `--safe-mode` (which disables all MCP servers/plugins/hooks), which is what isolated the leak to an MCP server rather than Claude Code's core Bash tool.

**Repro pattern:** appears to correlate with session length / number of `tokensave_context` or graph queries issued over a long-lived stdio connection rather than any single call — I don't have a minimal reproduction, just repeated real-world sessions hitting it. Happy to gather more diagnostics (e.g. periodic RSS sampling, heap profile) if there's a debug build or flag that would help narrow it down.

**Impact:** Had to uninstall tokensave (`tokensave uninstall`) to stop recurring full-system crashes, since the leak survived a version upgrade and the latest patch release doesn't appear to touch memory handling.

No proprietary code/project details included above — this is process/memory-level info only.

himanishpuri:
[...]

Issue #154, opened 2026-06-25
Feature request: Support documentation in MD format


alkampfergit:
In big legacy project we inevitably have complex classes. We noticed that AI Agent uses lots of token reading big code files to understand what the code does.

Apart splitting the code, for complex classes or set of classes we usually have dedicated readme.md that explain what the class does and how it is organized. Lots of time the AI needs only that information, not the entire class file.

We usually go by convention

xxx/yyy/super_big_class.cs
xxx/yyy/super_big_class.readme.md

Could it be a good idea for tokensave to index also this summary documentation? Maybe an MCP method GetFileSummaryInfo and something like that?

It could work with convention over configuration or allow small configurations that explain the convention.

aovestdipaperino:
What do you think about a tokensave-docs directory as child of the project root?
It doesn't have to be by file, but the front matter would tell tokensave when to inject the file in the context. You would be able to have a parallel hierarchy without polluting the source code. This way people might decide to gitignore it.
@alkampfergit ?

alkampfergit:
Sounds good, it could also be possible to put an option that tells tokensave where is the doc directory so it could be placed anywhere.

Wildcard patterns are useful, a practical case:

**/*.es8.cs -> elasticsearch 8 based class, I have specific instruction for them
**/*.es7.cs -> elasticsearch 7 completely different C# driver, different instructions

*Omni search*.cs -> file related to Omnisearch subsystem, I have small hints on how to handle them.

Etc.

aovestdipaperino:
Here's a concrete plan for this, with the trade-offs at each decision point.

## Where docs live: support both conventions

**Sidecar files** (`super_big_class.cs` + `super_big_class.readme.md`) — zero configuration, the doc travels with the code in reviews, and discovery is a pure filename match. Cost: it lives in the source tree, and it can't express one-doc-to-many-files.

**Docs directory** (`tokensave-docs/` by default, location overridable in `.tokensave/config.json`) with YAML front matter — keeps the source tree clean, gitignore-able, and the front matter carries an `applies_to` list of globs, which is exactly the `**/*.es8.cs` vs `**/*.es7.cs` case: one instructions file covering a whole family of files. Cost: docs can silently drift out of sync with the globs they claim to cover, and contributors have to learn the front-matter schema.

[...]

PR #307, opened 2026-07-26
feat(docs): index companion Markdown documentation (#154 phase 1)

aovestdipaperino:
Implements phase 1 of #154. Whole-doc granularity; **section anchors are deliberately out of scope** (see below).

## What it does

Discovers Markdown that documents source files, so an agent can read a summary instead of a 3000-line class. Two conventions, both shipped in one pass because the discovery walk is shared:

- **Sidecar** — `BigClass.cs` alongside `BigClass.readme.md`. Zero config, travels with the code in review. This is the convention the reporter already uses.
- **Docs directory** — `tokensave-docs/` by default, relocatable via a new `docs_dir` config field (empty string disables it). Each doc's YAML front matter carries an `applies_to` glob list, which is the `**/*.es8.cs` vs `**/*.es7.cs` case from the thread: one instructions file covering a family.

Both resolve to the same internal mapping and land in the graph as `Doc` nodes with `Documents` edges to the covered `File` nodes, so coverage is queryable through existing graph tools rather than a parallel subsystem.

Surfaced two ways, per the plan:
- `tokensave_doc(file)` — returns the doc path, content, every file that doc covers, and the drift signal. `include_content: false` for a coverage check without the body.
- `tokensave_entities` now reports `has_doc` / `doc_path` / `doc_hint`, so the agent sees a summary exists **before** deciding to read the file. Both surfaces share one helper, so they cannot disagree.

## Drift

A doc whose covered code was committed *after* it reports `doc_stale: true`. Without git history the field is **`null` (unknown)**, not `false` — a confident "not stale" would be a lie on a fresh clone or a non-git project.

## Decisions I made, worth a look

- **Sidecar is `<stem>.readme.md`**, matched case-insensitively. It never documents another doc (including itself), never matches a sibling stem prefix (`FooBar.cs`), and never reaches into a nested `Foo.d/` directory.
- **Docs covering nothing are dropped** rather than indexed. A glob legitimately goes empty when files are removed; that is not an error, but there is no mapping worth storing.
- **Front matter is hand-rolled.** One key holding a list of strings does not justify a YAML dependency. It handles inline and block sequences, CRLF, and a BOM, and degrades to "covers nothing" on anything it cannot read — never a hard error. An unterminated fence is rejected rather than swallowing the document.
- **`Doc` nodes rank below code** (`kind_tier` 3, `kind_rank_bonus` -1.5). A doc should be surfaced by the marker and the tool, not by outranking the code it describes — consistent with #295, which just landed.
- A covered file with no `File` node yields no edge. Not every extractor emits one, and a dangling edge is worse than a missing one.

## Not in this PR

Section anchors, in either form discussed: the addendum's line ranges with fingerprint resync and in-place Markdown rewriting, and alkampfergit's `## Section m:MethodName` symbol references. That choice was still open at the end of the thread, and whole-doc granularity is what the reporter said covers most of the need. Phase 2 (FTS over doc content, budget-capped context injection) and phase 3 (`inject: always`) are likewise untouched.

## Testing

- 17 unit tests for discovery, front-matter parsing, and coverage resolution.
- 10 integration tests through the real MCP handler: both conventions, the multi-file coverage case, the undocumented case, `include_content: false`, backslash-path normalization, the entities marker in both states, tool registration, and drift in **both** directions — unknown without git, and `true` with a real repo where code was committed after the doc.
- Verified by hand over a live `tokensave serve` stdio session on a scratch project, not only in fixtures.
- Full suite green; clippy `-D warnings` clean.

Incidental: the `test_tool_definitions_complete` count guard caught the new tool, exactly as designed — bumped 81/82 → 82/83 and added an explicit name assertion. Also refreshed three stale `#253` comments in `indexing.rs` that still claimed the resolver clones every node; it borrows now, and the remaining peak is tracked in #306.


rNoz:
@aovestdipaperino let me know if you want some support here, for instance, deep e2e testing this feature in a set of scenarios.

Issue #310, opened 2026-07-26
Redirect symbol grep when its target is the absolute project root


rNoz:
### tokensave version

7.7.0

### Operating system

macOS

### Installation method

Built from source

### What happened?

The PreToolUse guardrail redirects symbol-shaped `grep`/`rg` searches only when the target is recognized as code. It recognizes `.`, `./`, an omitted target, a known code-directory basename, or a supported source extension. It does not recognize an absolute path equal to the indexed project root.

Real agents frequently spell the repository target absolutely:

```text
rg -n 'target_looks_like_code' /Users/<user>/projects/<repo>
```

For an absolute directory the classifier only compares the last path component against `CODE_DIRS`, so any repository whose folder name is not one of `src`, `lib`, `app`, and friends passes through and produces no tokensave savings. The decision core is shared by Claude Bash/Grep, Kiro, and Factory Droid, so all three harnesses leak the same way.

### Steps to reproduce

1. `tokensave init` in a repository whose directory name is not a known code directory (for example `~/projects/tokensave`).
2. From that directory, send the hook a Bash grep on the absolute root: `printf '%s' '{"tool_input":{"command":"grep -rn handle_request /Users/<user>/projects/tokensave"}}' | tokensave hook-pre-tool-use`.
3. Repeat with the relative spelling `.` instead of the absolute path.

### Error output

```shell
grep/rg <symbol> /absolute/project/root       -> {"permission":"allow"}   (bug)
grep/rg <symbol> /absolute/project/root/docs  -> {"permission":"allow"}   (correct)
grep/rg <symbol> . | src/ | src/file.rs       -> {"permission":"deny"}    (redirected)
```

### Component

Agent integration (Claude, Cursor, etc.)

### Proposed behavior

Treat a target as the existing whole-project code bucket when:

1. the hook knows the indexed project root;
2. the raw target, after exact `~` or `~/` expansion, can be canonicalized; and
3. its canonical path equals the canonical project root.

The check runs only after an explicit `type` filter and after code/non-code glob classification, so the precedence landed in PR #280 is preserved. On a missing root, an unknown home directory, or any expansion or canonicalization failure, the classifier falls through to its current rules.

### Non-goals

- Do not classify every path inside the project as code. `<root>/docs`, `<root>/data`, parents, and siblings keep their current behavior.
- Do not use string-prefix matching. A sibling named `<root>-old` must still pass.
[...]

PR #311, opened 2026-07-26
fix(hooks): Redirect symbol grep whose target is the absolute project root

rNoz:
## Summary

- This topic branch took days of work built on top of previous Grep-* improvements, achieving even more features to save tokens! ... and almost another 5h of tests/e2e today (now rebased onto 7.7.0 as you cut release 2h ago). So all should be ready!
- Recognize a grep/Grep target whose canonical path equals the canonical indexed project root as the same whole-project search the classifier already redirects for `.`, `./`, and an omitted target.
- Carry the indexed root in `HookEnv`, derived from the working-directory probe the hook already performed, so the classifier stays a pure function of `(tool_input, env)` with no database or MCP access.
- Keep the match exact: siblings sharing the root's string prefix, parents, and every path inside the project are unchanged, explicit `type` and glob classification still take precedence, and any unknown root or failed canonicalization falls through to the existing rules.
- Solves #310 

## Motivation

`target_looks_like_code()` classifies a whole-project target from `.`, `./`, an omitted target, a known code-directory basename, or a supported source extension. For an absolute directory it only compares the last path component against `CODE_DIRS`, so `rg -n handle_request /Users/me/projects/app` passes through while `rg -n handle_request .` is redirected. Absolute repository paths are a common agent spelling, and every repository whose folder name is not a recognized code directory silently loses the savings the guardrail exists to produce.

The fix is deliberately the narrowest one that closes that gap. Anything below the root keeps its current classification, because a broad "inside the project" rule would hard-block legitimate searches over data, fixtures, and prose. Comparison is on canonicalized paths rather than string prefixes, so a symlinked or `..`-laden spelling still matches while `<root>-old` does not.

## Changes

- `src/hooks.rs`: `HookEnv` gains `project_root: Option<PathBuf>`, set from the same `.tokensave/tokensave.db` probe that already produced `cwd_has_tokensave_db`, so both fields always agree and activation semantics are untouched. `target_looks_like_code()` takes `&HookEnv` and, after the explicit `type` and glob rules, returns `true` when `path_is_project_root()` matches. Only absolute (or `~`-rooted) targets reach that comparison, which keeps the two `canonicalize` calls off every other hook invocation. `path_is_project_root_with_home()` isolates `~` / `~/…` expansion behind an explicit home argument so it can be tested hermetically; `~user` is never expanded.
- `src/hooks.rs`: new `project_root_target_tests` unit module covers tilde expansion, a tilde path below the root, a missing home directory, `~user`, a quoted absolute root, and a relative target.
- `tests/hooks_test.rs`: fourteen regressions over a real temporary indexed project (absolute root, trailing slash, prefix-sharing sibling, absolute `<root>/docs`, root reached through `..`, the root's parent, a root that is not the first target, structured `Grep` root path, non-code glob precedence, unknown root, non-existent root, opt-out, Droid `Execute`, Droid native `Grep`).

## Scope and known limitations

Deliberately out of scope, each pre-existing and unchanged by this PR:

- **Activation stays working-directory based.** The guardrail still only arms when the hook process itself runs at the project root; a session started in a subdirectory has working MCP tools and a silent guardrail. That is a separate defect with its own fix.
- **Only the first target of a Bash command is classified.** `grep -rn sym docs <root>` still passes, exactly as `grep -rn sym docs src` does today.
- **Bash narrowing flags are not parsed.** `--include='*.md'`, `-g`, and `--exclude` are skipped as flags, so `grep -rn sym --include='*.md' <root>` is redirected just as `grep -rn sym --include='*.md' .` already is. Teaching the Bash extractor about narrowing flags would fix both at once and belongs in its own change.

## Test plan

- [x] `cargo test --test hooks_test` passes, 133 passed. The four new blocking regressions fail on the unpatched classifier and pass after it.
- [x] `cargo test --lib project_root_target` passes, 7 passed.
- [x] `cargo test` passes, 2795 tests across 107 suites.
- [x] `cargo clippy --workspace --all-targets` is clean.
- [x] Real headless hook proof from an indexed repository with the release binary, covering both harness channels (stdout JSON and exit code plus stderr).

<details>

<summary>Headless proof (release binary, indexed repository)</summary>

```text
$ printf '%s' '{"tool_input":{"command":"grep -rn target_looks_like_code <root>"}}' \
    | target/release/tokensave hook-pre-tool-use
{"permission":"deny", ... "Use tokensave_search (definition) or tokensave_callers_for (usages) instead"}

hook process cwd = <root>
  bash, absolute root                    -> deny
  bash, absolute root, trailing slash    -> deny
  bash, absolute root through `..`       -> deny
  bash, ~/projects/tokensave             -> deny
  bash, absolute <root>/docs             -> allow
  bash, parent of the root               -> allow
  bash, root as the second target        -> allow   (pre-existing first-target rule)
  structured Grep, path = absolute root  -> deny
  structured Grep, root + **/*.md        -> allow   (#280 precedence)
  Droid Execute, absolute root           -> deny    (exit 2)
  Droid native Grep, absolute root       -> deny    (exit 2)
  Droid Task/explorer                    -> deny    (exit 2, #282 unchanged)
  Droid Task/worker                      -> allow   (exit 0, #282 unchanged)
  TOKENSAVE_DISABLE_GREP_HOOK=1          -> allow
```

</details>

## Checklist

- [x] Breaking changes documented (none). `HookEnv` gains a public field; it is a hook-internal type constructed through `from_runtime()`.


rNoz:
Working on the Windows CI Test issue.

aovestdipaperino:
Merged, thanks.

rNoz:
> Merged, thanks.
> 
> You have commit access here — no need to wait on me for your own PRs going forward, just merge once CI is green.

Thanks! I really appreciate the trust.

I think a good workflow going forward would be:

* For larger or more cross-cutting changes, I'll still open a PR and tag you for review. I'd rather have a second pair of eyes on anything that could have broader impact.
* For smaller, self-contained fixes, I'll merge them myself once CI is green.

I think that gives us a good balance between speed and safety.

PR #312, opened 2026-07-26
fix(ruby): scope self-denotation and block/method definee classification correctly

janstol:
## Summary

- Traverses `def`s nested inside `do…end`/`{…}` block bodies and inside method bodies, which were previously silently dropped — including `ActiveSupport::Concern`'s `included`/`class_methods`, `class_eval`/`instance_eval`, and ordinary blocks like `each`/`describe`.
- Classifies each block's actual definee correctly (own class, receiver's class, receiver's singleton, or unresolvable/opaque) instead of treating every block as either "always inherits" or "always a scope boundary".
- Requires positive evidence (`extend ActiveSupport::Concern`, or a shaped `concern`/`concerning` call) before treating `included`/`prepended`/`class_methods`/`concern`/`concerning` as Concern DSL, so ordinary same-named methods aren't misclassified.
- Scopes what `self` denotes correctly across plain `def`, `def self.x`, and `class << self` bodies, so a nested `instance_eval`/`def self.foo`/`class << self` inside a plain instance method is no longer misattributed as a class method of the enclosing class.

## Motivation

Several reachability and classification bugs in the Ruby extractor caused real, commonly-written Rails-shaped code (`ActiveSupport::Concern` modules, `class_eval` blocks, RSpec `describe` blocks) to either have its method definitions silently dropped from the graph, or attributed to the wrong owner (e.g. a per-instance `instance_eval` definition wrongly showing up as a class method, matchable by `private_class_method`). Both corrupt `callers`/`impact`/`dead_code` results for any codebase using these idioms. All fixes are grounded in behavior verified against Ruby 3.4.5/3.4.7 and activesupport 7.2.2.2/8.1.3 (probes included in the commit message).

## Changes

- `src/extraction/ruby_extractor.rs` — new `visit_block_body`/`visit_expression_blocks` traversal entry points; `BlockScope`/`BlockReceiver` classifiers modeling Ruby's default-definee rule; method bodies now traversed for nested definitions; `ExtractionState.in_concern_scope` and `self_is_instance` gate the Concern DSL and self-denotation edge cases respectively.
- `src/extraction/ts_state.rs` — new `SingletonScope`, `in_concern_scope`, and `self_is_instance` state.
- `tests/ruby_extraction_test.rs` — extensive regression coverage for the above (142 Ruby tests total).
- `CHANGELOG.md` — `[Unreleased]` entry.

See the commit message for the full differential-probe evidence and reasoning behind each classification decision.

## Test plan

- [x] `cargo test` passes (full workspace, including `cargo test --workspace`)
- [x] Tested manually: scratchpad fixture + `tokensave sync` + direct `sqlite3` query against `.tokensave/tokensave.db` confirming per-instance `instance_eval` definitions are correctly absent from the graph
- [x] Each new classification fix mutation-checked individually (temporarily reverted, confirmed only its own test(s) fail)

## Checklist

- [x] No breaking changes


aovestdipaperino:
Merged, thanks.

Issue #313, opened 2026-07-26
Context retrieval: bounded FTS scores discarded, exact-name provenance, index-side stemming (residuals of #264/#202, measured)


map588:
## Summary

Dogfooding `tokensave_context` on this repo (v7.7.0) traced the residual "context misses what search finds" behavior from #264 to specific root causes in candidate scoring, plus two adjacent findings. Filed as one issue so fixes can land as small PRs against an agreed diagnosis.

Measurement: a 12-query golden eval (natural-language queries, known expected entry points, hit@3 / MRR, indexed checkout of this repo at `adc47a8`). Baseline v7.7.0: **hit@3 33%, MRR 0.250**. With the fixes below: **hit@3 66%, MRR 0.528**, no query regressing. The harness is available on request.

## Finding 1 (bug): `search_nodes_bounded` discards the BM25 rank it computes

#264's fix (154a28a) added `ORDER BY bm25(...)` to the bounded per-term fetch — fixing *which rows* enter the pool — but `search_nodes_bounded_fts` still returns `score: 1.0` for every row. Its sibling `search_nodes_fts` returns the negated rank. The flat score is documented as intentional ("context building layers its own multi-signal ranking"), but that reranking is *multiplicative*: boosts applied to a flat base amplify pool noise instead of FTS relevance. Direct search ranks correctly; context building runs blind.

Fix: return `score = -rank` — the value is already computed, same bounded query, no repo-wide sorting, zero added cost. Requires two consumer recalibrations in `find_entry_points`: the exact-name floor (sized for ~1e-6 scores) and the per-term fetch width (3 rows/term is too few once row order is meaningful).

Eval impact alone: 33% → 58% hit@3.

## Finding 2 (bug): exact-name supplement matches symbols the user never wrote

`find_entry_points` feeds every extracted symbol to the exact-name supplement, including single-word segments from compound splitting. Repro on this repo: a query mentioning `tokensave_context` splits to `tokensave`, which case-insensitively exact-matches the `TokenSave` struct; the API-family supplement then floods the pool with its methods at exact-tier scores.

Repro (v7.7.0): `tokensave tool context --task "How does tokensave_context build its result: candidate selection, ranking, token budgeting, snippet truncation"` → `TokenSave`-adjacent entry points instead of `find_entry_points` / `rerank_candidates` / `build_context`.

[...]

Issue #264, opened 2026-07-23
tokensave_context misses indexed top-level asset generator for explicit query


RELnemtzov:
## Environment

- TokenSave: 7.6.0
- Operating system: Linux
- Installation: GitHub release binary
- Component: MCP server / context retrieval
- Project shape: medium JavaScript/TypeScript monorepo

## What happened?

`tokensave_context` returned unrelated application UI screens for an explicit natural-language request about application-icon generation. It omitted the indexed top-level generator and its adjacent test.

A direct `tokensave_search` for the generic filename term `generate-icons` returned the relevant generator functions and test with high scores, so the files and symbols were already present in the graph. This appears to be context candidate selection/ranking rather than an indexing failure.

Repository identity, product terminology, source code, generated assets, configuration values, and local paths are intentionally omitted.

## Steps to reproduce

1. In an indexed JavaScript/TypeScript monorepo containing a top-level icon generator and test, run:

```json
{
  "task": "Find the code that generates application icon assets and controls consistent geometry across the launcher, splash screen, OAuth consent logo, and browser favicon. Include its tests and build configuration.",
  "mode": "plan",
  "include_code": false,
  "max_nodes": 16
}
```

2. Observe that the returned entry points are unrelated application UI screens and the icon generator/test are absent.
3. Run `tokensave_search` with:

```json
{
  "query": "generate-icons",
  "limit": 10
}
```

4. Observe that the generator functions and its test are returned near the top, confirming they are indexed.

## Expected behavior

The explicit combination of “generates,” “icon assets,” “launcher,” “splash,” “OAuth logo,” and “favicon” should strongly rank the indexed icon generator and its test. Unrelated UI screens should not become the primary entry points.

If `tokensave_context` intentionally down-ranks top-level scripts or filename-only matches, it would be helpful for the result to explain that limitation or recommend a keyword/path refinement.

## Error output

No error was emitted; the context result was valid but irrelevant.

## Additional context

This reproduces on 7.6.0 and seems distinct from the spaced CamelCase retrieval problem fixed in #202: the relevant symbols use ordinary generic names, and direct search by the generator’s filename term succeeds.

RELnemtzov:
Follow-up after a read-only inspection of the public source at `f7aababa339369a44c2f08e16620dad146dd4bfb`:

[...]

Issue #120, opened 2026-06-16
Context: cap entry points before BFS expansion so traversal budget isn't spread too thin


aovestdipaperino:
## Problem

`find_entry_points` caps candidates to `max_nodes` via `apply_per_file_cap`, and `expand_subgraph` then runs BFS from **every** surviving entry point. When many roots survive, the per-root traversal budget collapses — e.g. with 36 entry points and `max_nodes = 120`, each root gets only ~3 nodes of expansion, which is too shallow to be useful.

## Fix

Cap the number of entry points used as BFS roots to `search_limit` (the candidates are already score-sorted after `rerank_candidates` + boosts), so each root gets a meaningful traversal budget. The remaining lower-ranked candidates can still be included directly as nodes, but should not each seed a full BFS.

## Acceptance

- Number of BFS roots is bounded by `search_limit`.
- For a broad query that produces many entry points, the resulting subgraph shows deeper expansion around top roots rather than uniformly shallow fan-out.
- Add a context test covering the many-entry-points case.

PR #314, opened 2026-07-26
fix(search): return real BM25 scores from search_nodes_bounded (#313)

map588:
Finding 1 + Finding 4 of #313.

## Problem

`search_nodes_bounded_fts` computes `bm25()` to ORDER BY it, then returns `score: 1.0` for every row — while `search_nodes_fts` in the same file returns the negated rank. Context building consumes the bounded variant, so its multiplicative reranking (kind/visibility/path boosts) runs on a flat base and amplifies pool noise instead of FTS relevance. This is the remaining half of #264: the fetch was ranked in 154a28a, but the rank never reaches the reranker.

## Changes

- Return `score = -rank`. Same per-term bounded query — no repo-wide sorting, the value is already computed, zero added query cost.
- `EXACT_MATCH_SCORE` 20 → 100: negated-BM25 runs ~5–30 before boosts; the exact-name floor must clear that band (it was sized for ~1e-6 flat scores).
- Per-term fetch widened to `(search_limit * 3).max(10)`: 3 rows/term is too few for the right symbol to survive merging once row order is meaningful. The BFS-root cap (#120) is unchanged.
- New `test_symbol_penalty` (0.4) in `rerank_candidates`: the wider fetch surfaces inline `#[cfg(test)]` functions, which carry a production path and escape `path_boost`. Detected by symbol shape (`::tests::`/`::test::` qualified-name segment or `test_` prefix); soft, so exact-name matches still surface.

## Verification

- New tests: bounded search returns positive, distinct, descending scores (name match outranks docstring match); `test_symbol_penalty` hits inline test fns, spares production symbols including names merely containing "test".
- Full suite: 2776 passed, 0 failed.
- 12-query eval from #313: hit@3 33% → 58%, MRR .250 → .444, no query regresses.

Note: if the retrieval-diagnostics PR lands first, its "strong" tier threshold should move 1.0 → 10.0 alongside this change (and vice versa — whichever merges second carries the one-line bump).

### DIFF
--- a/src/context/builder.rs
+++ b/src/context/builder.rs
@@ -188,10 +188,11 @@ impl<'a> ContextBuilder<'a> {
         symbols: &[String],
         options: &BuildContextOptions,
     ) -> Result<Vec<Node>> {
-        // Base score for an exact name match. Far above any realistic BM25
-        // score (~1e-6 in practice), so a perfect name match always wins the
-        // MAX merge over FTS noise and ranks ahead of it.
-        const EXACT_MATCH_SCORE: f64 = 20.0;
+        // Base score for an exact name match. Negated-BM25 scores from
+        // `search_nodes_bounded` run roughly 5–30 before structural boosts,
+        // so this must sit well above that band for a perfect name match to
+        // win the MAX merge over FTS hits and rank ahead of them.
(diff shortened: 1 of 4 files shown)


aovestdipaperino:
Merged, thanks.

Issue #202, opened 2026-07-11
tokensave_context misses CamelCase TSX symbol for spaced natural-language query


SooYoungJang:
## Summary

`tokensave_context` can miss a known indexed React Native/TypeScript component when the natural-language task uses the spaced form of its CamelCase name.

In this case, the indexed component is named `PromoBanner`. A task containing `promo banner` returned only an unrelated date helper from a test file, even though the query was path-filtered to the relevant UI area. Direct symbol search finds `PromoBanner`, and a control context query containing the exact identifier also finds it.

Repository identity, source code, and product-specific details are intentionally omitted.

## Environment

- TokenSave: 7.0.2
- 7.1.0 was reported as available; its release notes do not mention this retrieval case
- Platform: Windows / PowerShell
- Project: React Native, TypeScript/TSX
- Graph size: approximately 4,400 nodes

## Reproduction

1. Confirm that the symbol is indexed:

```json
{
  "query": "PromoBanner",
  "limit": 10,
  "path_include": ["apps/mobile/src/screens"]
}
```

Result: `PromoBanner` is returned as a function in a TSX screen file.

2. Run `tokensave_context` with a natural-language UI task:

```json
{
  "task": "Redesign the React Native home promo banner with a cleaner left-aligned text hierarchy, a right-side product visual, transparent-image support, responsive sizing, dates, and accessibility.",
  "mode": "plan",
  "include_code": true,
  "max_code_blocks": 4,
  "max_nodes": 16,
  "max_per_file": 8,
  "merge_adjacent": true,
  "path_include": [
    "apps/mobile/src/screens/HomeScreen",
    "apps/mobile/src/design",
    "apps/mobile/src/types"
  ]
}
```

Actual result: the only entry point is a generic `dateStr` helper from a TSX test file. `PromoBanner` is absent.

3. Control query:

```json
{
  "task": "Find the React Native PromoBanner component and plan responsive product image layout changes.",
  "mode": "plan",
  "include_code": true,
  "path_include": ["apps/mobile/src/screens"]
}
```

Result: `PromoBanner` and its nearby artwork helpers are returned.

## Expected

`tokensave_context` should treat the spaced natural-language phrase `promo banner` as a strong match for the CamelCase symbol `PromoBanner`, and production component matches should not be displaced by a generic test helper.

## Possible area

[...]

Issue #117, opened 2026-06-16
Context: exact-name matches silently dropped to FTS score (first-seen dedup in find_entry_points)


aovestdipaperino:
## Problem

`find_entry_points` (`src/context/builder.rs:218-249`) dedups search results across channels using first-seen insertion:

```rust
if Self::score_passes(sr.score, options.min_score)
    && seen_ids.insert(sr.node.id.clone())   // <-- first-seen wins
{
    candidates.push(sr);
}
```

FTS runs first and inserts each node with its BM25 score. The **exact-name supplement** that runs afterward tries to add the same node with the high base score (`20.0`):

```rust
for node in exact_nodes {
    if seen_ids.insert(node.id.clone()) {     // returns false for nodes FTS already saw
        candidates.push(SearchResult { node, score: 20.0 });
    }
}
```

When a node was already returned by FTS, `seen_ids.insert()` returns `false`, so the exact-match entry is discarded and the node **keeps its low FTS score**. A perfect name match that also happens to hit FTS gets buried by BM25 noise — exactly the case the exact-name supplement exists to prevent.

## Fix

Merge by **max score across channels** instead of first-seen. Track candidates in a `HashMap<String, SearchResult>` (or index into the `candidates` vec) and, on a duplicate id, upgrade the stored score with `score = score.max(new_score)` rather than dropping the new result.

## Acceptance

- A node returned by both FTS (low score) and exact-name lookup ends up with the exact-match base score.
- Add a regression test in `tests/context_test.rs` covering a query where the exact match also matches FTS.

PR #315, opened 2026-07-26
fix(context): exact-name supplement only for symbols the user wrote (#313)

map588:
Finding 2 of #313. Stacked on #314 (first commit here is #314; review the second).

## Problem

The exact-name supplement consumes every extracted symbol, including single-word segments produced by compound splitting — words the user never typed. Splitting `tokensave_context` yields `tokensave`, which case-insensitively exact-matches this repo's `TokenSave` struct; the API-family supplement then floods the candidate pool with its methods at exact-tier scores. Repro: `tokensave tool context --task "How does tokensave_context build its result: candidate selection, ranking, token budgeting, snippet truncation"` returns `TokenSave`-adjacent entry points instead of `find_entry_points`/`rerank_candidates`/`build_context`.

## Change

Gate the supplement on provenance via `is_authored_symbol`: a symbol qualifies only if it is a multi-word compound (the #202 synthesis path, preserved), appears verbatim as a whitespace-delimited query token (modulo surrounding punctuation), or was passed in `extra_keywords`. Derived single-word segments fall back to ranked FTS — ranked, not anointed. The #117 MAX-merge behavior is untouched.

## Verification

- 5 unit tests on the predicate (compound passes, verbatim passes, punctuation trimmed, derived segment rejected, extra keyword passes).
- Full suite: 2780 passed, 0 failed.
- The repro above flips to rank 1. 12-query eval from #313: hit@3 58% → 66%, MRR .444 → .528, no query regresses.

### DIFF
--- a/src/context/builder.rs
+++ b/src/context/builder.rs
@@ -308,9 +308,19 @@ impl<'a> ContextBuilder<'a> {
 
         // --- Exact name supplement ---
         // Ensures perfect name matches aren't buried by BM25 noise.
+        //
+        // Only symbols the user actually wrote qualify: a verbatim query
+        // token ("tokensave_context", "UserService") or a synthesized
+        // multi-word compound ("PromoBanner" from "promo banner", #202).
+        // Single-word segments derived by compound-splitting do NOT — the
+        // split of "tokensave_context" yields "tokensave", which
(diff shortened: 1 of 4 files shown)


aovestdipaperino:
Merged, thanks.

PR #316, opened 2026-07-26
feat(db): porter stemming + index-time camelCase terms, schema v14 (#313)

map588:
Finding 3 of #313. Stacked on #314 and #315 (review the last commit).

## Problem

FTS matching is exact-token: "candidates" never matches `candidate`, "truncation" never matches `truncate` (#264 noted the missing `generates` → `generate` normalization). Two query-side workarounds compensate — the trailing-`s` strip and `generate_stem_variants`' hand-rolled suffix table — covering the common inflections and missing the rest.

## Changes

- `nodes_fts` rebuilt with `tokenize='porter unicode61'`: inflected prose stems to indexed identifiers at match time.
- **Both workarounds deleted** (net −201 lines). Evidence: porter conflates every pair the suffix table targets (verified in FTS5: authentication ↔ authenticator ↔ authenticate, configuration ↔ configure, serializer ↔ serialization, truncation ↔ truncate, validator ↔ validated; the one jump neither makes is resolution → resolve). With porter active, the 12-query eval returns identical entry points with the workarounds removed.
- New `search_terms` column: camelCase inner words the tokenizer cannot derive (`updateCloudClient` → `Cloud Client`), computed at insert time by `text::search_terms`, weighted 6.0 in `bm25()`. Empty for snake_case names — unicode61 already splits `_`. No-op for this repo; closes the #202 class at the index for TS/Java/Go corpora.
- `split_compound`/`is_camel_case` move to the shared `text` module so the DB layer can reuse them; behavior unchanged.
- Migration v14: add column, Rust-side backfill, recreate `nodes_fts` + triggers, FTS rebuild. Derived data only — worst case is a re-index; no user-authored state is touched. Fresh-install schema mirrors the DDL.

## Verification

- Migration test: v13-shaped DB → v14 with backfilled `search_terms`, camelCase inner-word FTS hit, porter matching an inflected term.
- Full suite: 2780 passed, 0 failed.
- 12-query eval from #313: identical results to #315 (66% hit@3) — zero retrieval cost for a net-negative diff. Porter's headroom is inflections the deleted table never covered, and camelCase corpora this eval (a Rust repo) cannot exercise.

If deleting `generate_stem_variants` feels too aggressive for one PR, it splits out cleanly — say the word.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 
 ## [Unreleased]
 
+### Changed
+- **Symbol FTS now stems at the index (`porter unicode61`), and camelCase inner words are indexed at insert time (schema v14).** Inflected prose in a `tokensave_context` task — "candidates", "truncation", "generates" — previously matched nothing, because the unicode61 tokenizer is exact-token and the identifiers are `candidate`/`truncate`/`generate` (#264 called out the missing `generates` → `generate` normalization). The porter tokenizer makes that class of match work for any suffix at query time instead of patching common cases query-side. A new `search_terms` column carries the words a tokenizer cannot derive on its own — the inner words of a camelCase identifier (`updateCloudClient` → `Cloud Client`), computed once at insert by `text::search_terms` and weighted between `qualified_name` and `name` in `bm25()`; it stays empty for snake_case names, which unicode61 already splits on `_`, so the column adds nothing for codebases like this one and closes the #202 class (spaced natural-language query vs. camelCase TSX symbol) for the TS/Java/Go side of the language table. Migration v14 adds the column, backfills it, and rebuilds `nodes_fts` with the new tokenizer — all derived data, so the worst case is a re-index; nothing user-authored is touched.
+
+### Removed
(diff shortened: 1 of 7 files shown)


aovestdipaperino:
Thanks @map588 — holding this one on you rather than resolving it myself. Now that #314 and #315 are merged, this branch conflicts with them in `src/context/builder.rs`, `src/db/queries/search.rs`, and `tests/db_test.rs`. Since all three are your own changes from the same #313 series, you know which side wins in each hunk; I did not want to guess at ranking semantics and silently land something other than what you intended. Rebase on current `master` and I will merge as soon as CI is green.

Everything else in the queue (#311, #312, #314, #315, #317) is in. Note the Windows lib-test job flaked with a `STATUS_ACCESS_VIOLATION` on two unrelated branches today and passed on rerun, so if you see that, it is not your change.

map588:
Rebased onto v7.8.1 — the branch is now the single v14 commit on current master. The conflicts were as you guessed: my own #314/#315 hunks meeting their merged versions; resolution was taking the merged side everywhere and keeping only the v14 delta (split_compound/is_camel_case move to text.rs, generate_stem_variants + plural-strip deletion, migration). Full suite green locally (2878 passed).

map588:
Conflict resolution verified by comparing the rebased delta against the previously reviewed one: line-identical, except it no longer deletes a duplicated is_authored_symbol block that only existed in the pre-merge stack — nothing else changed. #314/#315 content is untouched on this branch: one is_authored_symbol, EXACT_MATCH_SCORE = 100, score = -rank at both sites, test_symbol_penalty intact. The v14 payload is complete: porter tokenizer in all three DDL sites, five-weight bm25() at all three call sites, plural-strip and generate_stem_variants both removed (the one remaining grep hit is a string literal in #314's unit test).

Full suite locally: 2878 passed, 0 failed. Ready to merge from my side @aovestdipaperino.

map588:
Just added the entry into change log, CI is running again because of the push but no changes to the code in this one.

aovestdipaperino:
Merged into `master` (3cb256c) — its V14 was collapsed with #326's into a single `migrate_v14` so there's one schema bump, not two. Thanks for the porter-stemming work and the net-negative diff! 🙏

PR #317, opened 2026-07-26
fix: prevent cross-process config clobbering

janstol:
## Summary

- Keep stable preferences in `config.toml` untouched during routine state-only persistence.
- Merge concurrent `state.toml` updates by field semantics instead of replacing newer state with a stale snapshot.
- Make partial-write retries idempotent when state succeeds but the subsequent config write fails.

## Motivation

Long-lived tokensave processes could overwrite preferences or machine-local state written by another process. In particular, routine MCP persistence could restore a stale `wildcard_permissions` value, concurrent pending-token updates could be lost, and retrying a partially successful save could apply the same token delta twice.

## Changes

- Extend the existing `config.toml.lock` transaction across config and state reads and writes.
- Track separate config and state baselines; leave `config.toml` byte-for-byte unchanged unless a stable preference changed or legacy migration is pending.
- Merge counters, timestamped caches, version markers, and installed-agent additions/removals according to their domain semantics.
- Advance the state baseline immediately after its durable write while advancing config and legacy-migration bookkeeping only after config succeeds.
- Add regression coverage for concurrent changes, deleted state, failed rereads, partial-write retries, and legacy migration.

## Test plan

- [x] `cargo test` passes

`cargo test --workspace` and `cargo fmt --all -- --check` pass.

## Checklist


aovestdipaperino:
Merged, thanks.

Issue #320, opened 2026-07-27
`serve`: version stamp persists as empty string, forcing a reindex on every start that fails with `database table is locked` and silently serves a truncated index (file_count=1)


someshfengade:
## Summary

Every `tokensave serve` startup detects a version mismatch reported as `indexed by ""` (an **empty string**, not a version number) and triggers a forced full reindex. That reindex then **fails** with `database table is locked`, and the failure is **non-fatal and silent** — the MCP server continues serving whatever is in the DB.

The practical consequence is that `tokensave_status` (and therefore every retrieval tool) intermittently returns a **near-empty graph**. On one run the index reported `file_count: 1 / node_count: 264` for a 259-file project; seconds later the same call returned `file_count: 258 / node_count: 7464`.

For an agent consumer this is invisible — the tool call succeeds, it just returns almost no context. This looks like "the agent can't find relevant code" rather than an index failure.

## Environment

| | |
|---|---|
| tokensave | **7.8.1** (also reproduced on 7.6.0) |
| OS | macOS (Apple Silicon, arm64 sonoma bottle) |
| Install | Homebrew — `aovestdipaperino/tap/tokensave` |
| Project | Python, 259 files, ~3.0 MB source, DB ~29 MB |
| Branch mode | multi-branch enabled, 11 tracked branches |
| Clients | VS Code (stable) + VS Code Insiders + Copilot CLI + Codex CLI, all registered via `tokensave install` |

## Reproduction

With more than one MCP client registered (so multiple `tokensave serve` processes may touch the same SQLite DB), run a minimal handshake from the project root:

```sh
cd /path/to/project && printf '%s\n' \
'{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1"}}}' \
'{"jsonrpc":"2.0","method":"notifications/initialized"}' \
'{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"tokensave_status","arguments":{}}}' \
| tokensave serve 2>&1 | tail -25
```

## Observed output

**Run 1** (7.8.1):

```
[tokensave] previous operation was interrupted — checking database integrity…
[tokensave] tool call: tokensave_status
[tokensave] major upgrade or schema change detected (indexed by "", running 7.8.1) — forcing project reindex…
[...]

PR #321, opened 2026-07-27
fix: stamp index version and order bulk load before clear (#320)

someshfengade:
> **Disclosure: this PR was written by an AI coding agent** (GitHub Copilot), driven by the user who reported #320, and is offered as a **potential** fix rather than a confirmed one. The root-cause analysis below is derived from reading the source and from reproduction logs on a real project, but the maintainer should judge whether this is the intended fix — particularly the ordering change in `index_all`, which touches the destructive path. Please review accordingly.

Potential fix for #320.

## Problem

Two independent defects compound into a destructive loop.

**1. `last_indexed_version` is never written by the CLI.**

The marker is written in exactly two places, both in `src/mcp/server.rs::run_version_reindex`. `tokensave init` / `index_all` never write it, so every CLI-initialised project keeps the `String::new()` default.

`cloud::bump_kind("", running)` cannot parse the empty old version, so it returns `BumpKind::Major`. The result is that the first `tools/call` of **every** session on **every** CLI-indexed project logs:

```
[tokensave] major upgrade or schema change detected (indexed by "", running 7.8.1) — forcing project reindex…
```

and runs a full reindex of a project that is already current.

**2. `index_all` clears the database before it acquires bulk-load mode.**

```rust
self.db.clear().await?;
self.db.begin_bulk_load().await?;
```

`begin_bulk_load` drops 14 indexes and 7 triggers, so it needs the table lock. When a second tokensave process holds it — a common setup, since `tokensave install` registers VS Code, VS Code Insiders and one or more CLI agents against the same project and DB — it fails with:

```
[tokensave] version reindex failed: database error: failed to begin bulk load:
SQLite failure: `database table is locked` (operation: begin_bulk_load)
```

By then `clear()` has already deleted every row. Note `busy_timeout` does not help here: this is `SQLITE_LOCKED`, not `SQLITE_BUSY`.

**The compounding.** `run_version_reindex` treats the failure as non-fatal and `return`s early, so the marker is *never* advanced. The server keeps serving out of the now-empty graph, and because the marker still reads `""`, the next start repeats the whole cycle. Observed on a 259-file project — two consecutive `tokensave_status` calls against the same DB:

| run | `file_count` | `node_count` |
| --- | --- | --- |
| 1   | **1**        | **264**      |
| 2   | 258          | 7,464        |

`tokensave doctor` reports the project fully healthy throughout, so the condition is invisible unless you read stderr.

## Changes

**`src/tokensave/indexing.rs`**

1. Swap the order to `begin_bulk_load()` → `clear()`. A lock failure now aborts with the previous index **intact** instead of destroying it. This is safe because `db::queries::clear::clear()` issues explicit `DELETE`s against all seven tables and does not rely on FK cascade, so it does not depend on the `PRAGMA foreign_keys = OFF` / index-drop state either way. It is also slightly faster: the FTS triggers are already dropped, so the `DELETE FROM nodes` no longer fires a per-row FTS delete trigger.
2. Add `record_indexed_version()`, called after a successful index, so a CLI-indexed project no longer looks pre-7.0 and the redundant forced reindex never starts. It is best-effort — a failure to persist the marker is reported on stderr but does not fail an index that otherwise succeeded — and it skips the write when the value is already current, to avoid rewriting `config.toml` on every index.

Only the first change is needed to stop the data loss; only the second stops the needless reindex. They are kept together because the bug requires both to reproduce.

## Tests

**Modified — please look at this one.** `test_first_tool_call_backfills_last_indexed_version` relied on `index_all` leaving the marker empty, i.e. its premise encoded the bug. Its actual intent — that a **pre-7.0** project gets backfilled on first tool call — is still valuable, so rather than delete it I made the legacy state explicit: it now resets the marker to `""` after indexing before exercising the backfill path.

**Added:**
- `test_index_all_records_running_version` — a full index records the version that produced it.
- `test_cli_indexed_project_needs_no_forced_reindex` — drives a real `tokensave_search` call through `run_server_with_messages` against a CLI-indexed project and asserts the marker is correct after the version-reindex gate settles.

I did **not** add a test for the ordering change. Deterministically inducing `SQLITE_LOCKED` on `begin_bulk_load` from inside the test harness did not look reliable, and I would rather flag the gap than land a flaky test. Suggestions welcome if you know a good way to force it.

## Verification

Toolchain 1.95.0 per `rust-toolchain.toml`, built with `--no-default-features --features lite,test-transport`:

- `cargo fmt --check` — clean
- `cargo clippy --all-targets` — no new warnings
- `cargo test --test mcp_server_test` — **44 passed, 0 failed**
- `cargo test --test config_test` — **34 passed, 0 failed** (incl. `test_last_indexed_version_persists`, `test_legacy_config_without_last_indexed_version_loads_empty`)
- Full suite — one pre-existing failure cluster: 7 doc-related tests in `mcp_handler_test.rs` fail under `lite`. **Confirmed pre-existing**, identical 184 passed / 7 failed with my changes stashed on a clean `797207a`. They appear to need a language feature that `lite` disables; the full feature set was not built locally.

`CHANGELOG.md` updated under `[Unreleased] → Fixed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 
 ## [Unreleased]
 
+### Fixed
+- **A project indexed by the CLI no longer force-reindexes on the first MCP tool call of every session, and a failed forced reindex no longer leaves the project with an empty index.** `last_indexed_version` was written only by the MCP server's `run_version_reindex`, so every project created by `tokensave init` kept the `String::new()` default. `bump_kind("", running)` classifies an unparseable `old` as `BumpKind::Major`, so the first `tools/call` of each session logged `major upgrade or schema change detected (indexed by "", running <version>) — forcing project reindex…` and ran a full reindex — on a project that was already current. `index_all` opened that reindex with `db.clear()` *before* `db.begin_bulk_load()`, so when `begin_bulk_load` lost the table lock to a second tokensave process (a common setup: VS Code, VS Code Insiders, and one or more CLI agents all registered against the same project) the rows were already gone, the reindex returned early, and the marker was never advanced. The server treats that failure as non-fatal and keeps serving, so every subsequent tool call answered out of a gutted graph — observed as `file_count: 1` on a 259-file project — and the same destructive cycle repeated on the next start, since the marker still said `""`. A full index now records the running version, so the redundant reindex never starts; and `begin_bulk_load` runs before `clear`, so a lock failure aborts with the previous index intact instead of destroying it. Clearing after the FTS triggers are dropped also avoids firing a per-row FTS delete trigger for every node.
+
 
 ## [7.8.1] - 2026-07-27
 
--- a/src/tokensave/indexing.rs
+++ b/src/tokensave/indexing.rs
@@ -255,9 +255,12 @@ impl TokenSave {
         write_dirty_sentinel(&self.project_root);
         let start = Instant::now();
 
-        // 1. Clear existing data and enter bulk-load mode
-        self.db.clear().await?;
+        // 1. Enter bulk-load mode, then clear existing data. Order matters:
+        // `begin_bulk_load` can fail while another process holds the table
+        // lock, and clearing first would leave the project with an empty index
+        // that the MCP server then keeps serving (#320).
         self.db.begin_bulk_load().await?;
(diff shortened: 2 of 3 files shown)


aovestdipaperino:
Merged into `master` (edb7678) alongside #326 and #316 — thanks for the fix and the thorough root-cause writeup on the destructive reindex loop! 🙏

PR #326, opened 2026-07-28
fix(resolver): stop the resolver from producing annotates edges at all

janstol:
## Summary

- The resolver no longer produces `annotates` edges at all: `kind_compatible`'s `EdgeKind::Annotates` arm is now `=> false`.
- Schema migration V14 repairs existing databases in place, widened to delete phantom `annotates` edges targeting either `annotation_usage` or `decorator` nodes.
- `Cargo.toml` is deliberately left at `7.8.1` — version bumps land as separate `chore(release)` commits.

## Motivation

`annotates` names exactly one relation to every consumer: attachment of an annotation/decorator usage to the item it decorates. Extractors already emit that edge directly at the usage site. The resolver's `kind_compatible` allowlist (`Annotation | Decorator`) let an `Annotates` ref additionally bind to any same-named usage node, producing phantom self-edges and cross-node edges with no consumer that reads them as attachment.

Measured before this fix (`edges e JOIN nodes t ON t.id = e.target WHERE e.kind='annotates' AND t.kind='annotation_usage'`):

| Corpus | phantom `annotates` | total `annotates` | phantom % |
|---|---|---|---|
| thingviewer (Flutter, 78 files) | 104 | 208 | 50% |
| tokensave itself (Rust, 408 files) | 3,244 | 3,457 | 94% |

`NodeKind::Decorator` was also in the allowlist on the mistaken premise that it is a declaration kind. It is not: `Decorator` is emitted at the *application* site (`@foo(...)`) and is always the *source* of the direct attachment edge, never a legitimate target. Leaving it in the allowlist kept the same phantom class alive for Python/TS, reachable via a cross-language name collision.

## Changes

- `src/resolution/resolver.rs`: `EdgeKind::Annotates => false` in `kind_compatible` — a complete kill switch, since every path that mints a `ResolvedRef` filters through this gate.
- `tests/resolution_test.rs`: kept `test_annotation_ref_does_not_bind_to_sibling_usage`; replaced two prior "positive control" tests with negative assertions that a ref must not resolve to a `NodeKind::Annotation` declaration nor to a `NodeKind::Decorator` usage node.
- `src/db/migrations.rs`: widened `migrate_v14`'s `DELETE` target set from `kind = 'annotation_usage'` to `kind IN ('annotation_usage', 'decorator')`. `annotation` is deliberately not added — legitimate direct edges target it (e.g. `@Retention @interface Foo {}`), and `TokenSave::open` already forces a full reindex on any migration, which overwrites stale rows regardless.
- `tests/migration_test.rs`: extended the fixture with a phantom usage-to-decorator row (must be deleted) and a legitimate decorator-as-source row (must survive), proving the delete keys on target kind.
- `CHANGELOG.md`: corrected the now-false claim that "usage-to-declaration edges are unaffected".

## Test plan

- [x] `cargo test` passes (both `--no-default-features --features lite` and full default features; 7 pre-existing sidecar-documentation failures under `lite` confirmed unrelated, reproduced identically on stock `master`)
- [x] `cargo clippy` has no new warnings (`cargo clippy --workspace --all-targets`, `cargo fmt --all -- --check`)
- [x] Tested manually: mutation checks (reverting the resolver allowlist, and narrowing the migration `DELETE` two different ways) confirmed each new test fails exactly as expected and nothing else. End-to-end verification against scratch copies of two real stale v13 databases (this repo's own root DB and a Dart/Flutter corpus) confirmed `user_version` reaches 14, phantom-edge count drops to 0, and legitimate `annotates` target-kind counts are unchanged. A synthetic Python/Java fixture with an intentional cross-language name collision confirmed no `annotates` edge targets a `decorator` or `annotation_usage` node even in that case.

## Checklist

- [x] Breaking changes documented (if any) — N/A, this is a bug fix; existing databases are repaired in place by migration V14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 ## [Unreleased]
 
 ### Fixed
+- **`annotates` edges no longer bind an annotation usage to another usage of the same annotation, including itself.** The resolver allowed an `Annotates` reference to target a sibling `AnnotationUsage` or `Decorator` node, and every extractor names that reference after the annotation itself, so any `@override`/`#[test]`/`@Override`-style usage could resolve straight to another usage instead of staying unresolved — a phantom edge affecting 11 language extractors. The resolver now never produces `annotates` edges at all; the attachment edge extractors already emit directly at the usage site is the only one, matching what every consumer reads. Schema migration v14 removes any phantom edges already stored in existing projects' databases automatically on next open, no resync required.
 - **Dart calls made inside closures and local functions are no longer dropped.** `extract_call_sites` skipped `function_expression`/`lambda_expression` outright instead of attributing their calls to the enclosing named symbol, so calls inside Flutter callbacks (`onPressed:`, `builder:`, `setState(…)`, `.map(…)`) and local functions were invisible to `tokensave_callers`, `tokensave_impact`, and `tokensave_dead_code` — a helper reached only from a callback looked dead. Same fix as TypeScript #209 for nested arrows.
 
 ## [7.8.1] - 2026-07-27
--- a/src/db/migrations.rs
+++ b/src/db/migrations.rs
@@ -15,7 +15,7 @@ use crate::errors::{Result, TokenSaveError};
 
 /// The highest migration version defined in this file. Bump this and add a
 /// new entry to `run_migration` whenever the schema changes.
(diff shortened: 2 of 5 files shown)


aovestdipaperino:
Merged into `master` (75ae970) — its schema migration was collapsed with #316's into a single `migrate_v14` (phantom-edge repair + porter FTS/search_terms in one version). Thanks for the deep resolver analysis and the phantom-edge measurements! 🙏

Issue #170, opened 2026-07-04
config.json exclude patterns are not enforced when a symlink leads outside them (unlike --skip-folder)


alligatorshoes:
**Summary**

Exclude globs set in `.tokensave/config.json`'s `"exclude"` array do not prevent the indexer from following a symlink that lives *inside* an already-excluded directory and escaping to a location outside the project root. The equivalent `--skip-folder` CLI flag on `sync`/`init` does not have this problem — it appears to prune the directory before descending, while the config-file exclude apparently only filters after the walk (or after symlink resolution), which is too late.

**Reproduction**

1. Project layout:
   ```
   project/
     build-output/                  <- listed in config.json "exclude": ["build-output/**"]
       some-nested-prefix/
         dosdevices/
           z:  -> /                 <- symlink to filesystem root
           w:  -> /mnt              <- symlink to a mount point
   ```
   (This pattern occurs naturally with Wine prefixes, which create `dosdevices/z: -> /` and similar drive-letter symlinks — but any symlink-inside-an-excluded-dir reproduces it.)

2. `.tokensave/config.json` has:
   ```json
   "exclude": ["build-output/**", ...]
   ```

3. Run `tokensave sync` (or `init`) from the project root.

**Expected:** `build-output/` and everything under it, including the symlinks, are skipped — matching the exclude pattern.

**Actual:** The indexer follows `dosdevices/z: -> /` and begins walking the entire filesystem, including slow network-mounted paths (e.g. a FUSE-mounted network drive) far outside the project. In our case this ran for 6+ minutes with the process blocked on `__fuse_simple_request` before dying/timing out, with `sync.lock` left behind referencing a PID that had already exited (see below).

**Workaround:** Passing the same folder explicitly via the CLI flag works correctly and completes almost instantly:
```
tokensave sync --skip-folder build-output
```
This scanned only the intended source files (hundreds, not filesystem-wide) and finished in ~2.8s.

**Secondary issue — stale lock file:** After the runaway process died/exited on its own, `.tokensave/sync.lock` was left behind still containing the dead PID. Subsequent `tokensave sync` invocations refused to run ("another sync is already in progress (PID ...)") even though that PID no longer existed. Might be worth having `sync` verify liveness of the PID in the lock file before refusing to proceed, rather than requiring manual deletion of the lock file.

**Environment:** tokensave 7.0.3, Linux x86_64.

aovestdipaperino:
[...]

PR #328, opened 2026-07-28
fix(indexing): stop the walk at symlinks that re-enter the project root

### LINKED ISSUE #327: File walker follows symlinks out of the project root — `dosdevices/z: -> /` (Wine) makes init/sync unbounded, serve pins a core forever
### Summary

The index/sync file walker follows directory symlinks without bounding or excluding symlinks that escape the project root. A symlink pointing at `/` (which every Wine prefix ships as `dosdevices/z: -> /`) makes indexing effectively unbounded: `tokensave init` never completes, and `tokensave serve`'s startup catch-up sync pins one core at ~100% (95% sys time) indefinitely — until SIGKILL (SIGTERM appears to be ignored while walking).

Real-world impact: my project root is my home directory (agent launched from `~`), which contains a Wine game prefix. Every agent session leaked one spinning `serve`; I found two stale ones that had burned 20.5 h and 14.4 h of CPU, holding an idle laptop at 69 °C / 41 W on battery.

### Environment

- tokensave **7.8.1** (also reproduced on 7.0.2) — release binary `tokensave-v7.8.1-x86_64-linux.tar.gz`
- Pop!_OS 24.04 (Ubuntu-based), kernel 6.18.7, x86_64

### Minimal repro (3 commands, no Wine needed)

```bash
mkdir -p proj/src && cd proj
echo 'fn main() {}' > src/main.rs
ln -s / rootlink        # emulates Wine's dosdevices/z: -> /
tokensave init          # never completes; ~100% CPU until SIGKILL
```

`strace -f` shows the walker crossing into the symlink **683,707 times in 40 s**:

```
statx(AT_FDCWD, ".../proj/rootlink/snap/gtk-common-themes/.../folder-image.svg", ...)
```

Same syscall shape as the long-running `serve` processes: ~95% kernel time on one `tokio-rt-worker`, dominated by `statx` (254k/8s), `getdents64` (37k/8s), `openat`/`close`, walking `<project>/…/dosdevices/z:/usr/…` with per-directory probes for `.git`, `.jj`, `.gitignore`, `.ignore`.

### What I ruled out

- **A plain symlink cycle inside a small project does not reproduce** (steady-state 0% CPU) — ancestor-loop detection seems to work. The problem is specifically a symlink whose target is *outside/above* the project root (worst case `/`), which is followed and walked exhaustively; combined with `/` transitively containing the project again (via `/home/...`), the walk never converges.
- Not the #80 watcher issue — reproduces on versions long after that fix, and with the CLI `init` path alone.
- Not environment-specific — reproduced under a clean empty `$HOME`.

### Suggested direction

Don't follow directory symlinks during the walk (ripgrep/`ignore` default), or at minimum skip symlinks resolving outside the project root. Wine prefixes make `* -> /` symlinks common enough on Linux desktops that this hits ordinary users who index a directory containing one.

Happy to test a patched build.

rNoz:
## Summary

- Prune a directory symlink whose canonical target is the canonical project root itself or one of the root's ancestors. `dosdevices/z: -> /`, which every Wine prefix ships, is the reported case; `-> $HOME` with `$HOME` as the root and `-> .` are the same shape.
- Apply the rule in all three link-following walks: `scan_files_walkdir`, `scan_files_with_gitignore`, and the gitignore fallback `has_source` probe.
- Leave the deliberate external-source-tree feature (#34) untouched: a symlink into a disjoint tree is still followed and indexed.
- Fixes #327

## Motivation

`tokensave init` and `sync` never complete, and `serve`'s freshness scan pins one core at ~100% (95% sys) indefinitely, when the project contains a directory symlink that resolves to `/`. The reporter's root is `$HOME`, which contains a Wine game prefix; two leaked `serve` processes had burned 20.5 h and 14.4 h of CPU. `strace -f` showed the walker crossing into the link 683,707 times in 40 s.

Following a link that resolves to the root or above it can only re-expose paths the walk already covers, plus the entire rest of the filesystem: `/` contains the project, which contains the link, which resolves to `/` again. Both walkers' ancestor-loop detection does fire eventually, but only after each pass has walked everything reachable, including `/proc`, `/sys`, and network mounts. `exclude`/`--skip-folder` pruning (#170, #182) only helps when the link happens to live under an excluded directory.

Two simpler fixes were rejected because both revert a merged feature:

- `follow_links(false)` reverts #34 ("follow symlinked source directories during indexing") wholesale.
- "prune every symlink resolving outside the project root" is precisely what #34 added, since both of its tests symlink an external directory into the project.

So the rule separates "escapes the root" (a supported layout) from "swallows the filesystem" (the defect), and needs no new configuration.

## Changes

- `src/tokensave/indexing.rs`: two new private helpers. `path_contains_root(target, canonical_root)` is the pure relation (`canonical_root.starts_with(target)`), component-wise, so `/home/user2` is not an ancestor of `/home/user`. `reenters_project_root(path, canonical_root)` canonicalizes the link and applies it, failing open when either side cannot be canonicalized (an unresolvable link or root keeps its previous treatment, matching the containment checks in `src/context/builder.rs` and `src/global.rs`).
- `scan_files_walkdir`: the guard runs immediately after the depth-0 exemption and before the hidden-directory, `include`, manifest, and `exclude` rules, so no `include` glob or manifest entry can re-enable an escaping link.
- `scan_files_with_gitignore`: same guard in the existing `filter_entry`, after the `is_excluded_dir` prune so an already-excluded link pays no extra `canonicalize`.
- The `has_source` fallback probe: same guard. It is bounded by `max_depth(2)`, so this is consistency, not a second hang.
- All three sites exempt the walk root (`depth() == 0`), so a project opened through a symlinked path (`/work/proj-link -> /real/proj`) still scans; the root link canonicalizes to the root and would otherwise prune the entire walk.
- Cost is one `canonicalize` per directory symlink encountered plus one per scan for the root. Ordinary directories pay nothing.
- `tests/symlink_walk_scope_test.rs`: new unix-gated suite, 8 tests. Every fixture carries a walker fingerprint (a `.gitignore`d source file that only the plain walker indexes), so each test's exact expected path set also pins which of the two walkers produced it, rather than silently accepting the gitignore-aware walker's plain-walkdir fallback.
- `src/tokensave/indexing.rs`: new `walk_scope_tests` unit module, 8 path-relation and fail-open cases (the verbatim-path cases are Windows-gated because only Windows parses `\\?\C:\…` into components).

## Behavior change to be explicit about

Files that live outside the root but under an ancestor the link pointed at are no longer indexed. With root `/home/u/proj` and `proj/uplink -> /home/u`, the file `/home/u/other/lib.rs` used to be indexed as `proj/uplink/other/lib.rs`. That is the defect, not a feature: it silently pulls the user's home directory into a project index. Rows created that way before the fix are removed by the next full `tokensave sync` (covered by a regression test). The focused MCP freshness path keeps such a row until then, which is the pre-existing behavior for any file that leaves the scan set while remaining on disk, for example after a new `exclude` glob.

## Scope and known limitations

Deliberately out of scope, each pre-existing and unchanged here:

- **A symlink into a large disjoint tree is still followed** (`-> /usr`, `-> /mnt/nas`). It terminates, and it is exactly the layout #34 supports; `exclude` globs and `--skip-folder` remain the tool for it (#170, #182).
- **Non-symlink filesystem cycles are invisible to this rule.** Bind mounts and `/proc` self-references are not symlinks, so `path_is_symlink` never sees them.
- **No new configuration.** A `follow_symlinks` knob is a product decision, not a bug fix, and would not be one logical change with this.
- **Explicit `.tokensave/project.json` external globs** (`project_manifest::expand_external_files`) expand through the `glob` crate, not these walkers. That is an opt-in the user spells out, and it is untouched.
- **Duplicate indexing through a link to a descendant of the root** (`proj/link -> proj/src`) is unchanged.

## Test plan

- [x] `cargo test --test symlink_walk_scope_test`: 8 passed. Five are red on `ea08cce` before the fix: the two ancestor-link cases (one per walker), the nested re-entry case, the `include`-bypass case (both walkers, since the pattern names the hidden directory itself), and the post-upgrade sync-cleanup case. The other three are invariants that must not regress: a disjoint sibling link is still followed under both walkers, a symlinked project root still scans under both walkers, and a link to the root itself stays pruned (walkdir's own ancestor-loop detection already handled that one, so it is green before the fix).
- [x] `cargo test --lib walk_scope`: 7 passed on macOS (the eighth is the Windows-gated verbatim-path case).
- [x] `cargo test`: 2,898 tests across 108 suites, 0 failures.
- [x] `cargo clippy --workspace --all-targets` clean.
- [x] `cargo fmt --all -- --check` clean.
- [x] Real before/after proof with release binaries built from `ea08cce` and from this branch, including the reporter's exact `ln -s /` repro.

### Another last Repro (e2e, for anyone)

```sh
mkdir -p /tmp/ts327wine/src && printf 'fn main() {}\n' > /tmp/ts327wine/src/main.rs
ln -s / /tmp/ts327wine/rootlink
cd /tmp/ts327wine && timeout 25 tokensave init      # before: never finishes
```

<details>

<summary>Release-binary proof (macOS 15, both binaries built from the same tree state as their commit)</summary>

Reported repro, `ln -s /` inside a one-file project:

```text
=== PRE-FIX (ea08cce), timeout 25s ===
Initialized TokenSave at /private/tmp/ts327wine
(no indexing result: killed by the timeout)
real 0m25.016s   user 0m6.987s   sys 0m12.471s

=== POST-FIX ===
✔ indexing done: 1 files, 2 nodes, 1 edges in 5ms
real 0m0.214s    user 0m0.016s   sys 0m0.026s
```

Bounded variant that shows what the walk actually swallows: project at `/tmp/ts327demo/proj`, `uplink -> /tmp/ts327demo`, and 3,000 unrelated `.rs` files in the sibling `/tmp/ts327demo/outside`:

```text
=== PRE-FIX (ea08cce) ===
✔ indexing done: 3002 files, 6004 nodes, 3002 edges in 296ms
   (the project's own file, all 3,000 outside files, plus the project
    re-entered through the link as uplink/proj/src/main.rs)

=== POST-FIX ===
✔ indexing done: 1 files, 2 nodes, 1 edges in 8ms
```

`serve`'s freshness scan is the same code path (`find_stale_files` calls `scan_files`), so it is fixed by the same change.

</details>

## Checklist

- [x] No dependency changes
- [x] Breaking changes documented: none in the API. The one behavior change is the ancestor-side content described above.


melonwer:
Tested `rnoz/symlink-walk-scope` (ac8fa63) on the machine from the report — works. The `ln -s / rootlink` repro now indexes in under a second (strace shows the walk never enters the link), a nested Wine-style `dosdevices/z: -> /` is fine too, and `serve` on the same home project that used to pin a core now idles at 0%. Symlinks to disjoint external trees are still followed (#34), and a symlinked project root still scans. All 8 new tests pass here.

rNoz:
> Tested `rnoz/symlink-walk-scope` ([ac8fa63](https://github.com/aovestdipaperino/tokensave/commit/ac8fa63ec7540ba2e1b5279a2e3d10ef4e6036f0)) on the machine from the report — works. 

Great, thanks for double confirming in your machine! Shipping it! 🚢

PR #330, opened 2026-07-29
feat(context): add max_code_lines to cap snippet size (#324)

### LINKED ISSUE #324: Feature Request: max_code_lines parameter on tokensave_context to cap snippet size
Environment

  - tokensave: 7.6.2
  - Platform: Windows 11, Git Bash (MSYS2)

  Problem

  tokensave_context(include_code=True) returns the full function or class body with no way to bound the size. For large functions (100–500+ lines) this forces a binary
  choice: include_code=False gives bare symbol location only, which is often insufficient to act on without a follow-up Read call that counts against the call budget;
  include_code=True returns the full body, which for a large function can cost thousands of tokens for a single symbol when all that was needed was the signature and
  first 20 lines. There is nothing in between.

  Steps to observe

  # Investigating a large batch-processing function.
  # You need the signature and first ~30 lines to understand
  # its inputs, outputs, and high-level structure.

  tokensave_context("process_payment_batch", include_code=True)
  → Returns: full 300-line function body (~2,000 tokens)
    Call budget consumed: 1 of 4

  # With max_code_lines=30 (proposed):
  tokensave_context(
      "process_payment_batch",
      include_code=True,
      max_code_lines=30
  )
  → Returns: first 30 lines + "// … (270 lines omitted)"
    Call budget consumed: 1 of 4 — same cost, ~10x less output
  Ask

  A max_code_lines (or max_code_chars) parameter on tokensave_context that caps the returned snippet length per symbol and appends a truncation marker.

  Proposed fix

  Schema change in src/mcp/tools/definitions.rs — add alongside include_code:

  max_code_lines: Option<usize>
  // only meaningful when include_code is true
  Handler change wherever code snippets are assembled:

  let lines: Vec<&str> = source_for_symbol.lines().collect();

  let (snippet, truncated) = match request.max_code_lines {
      Some(limit) if lines.len() > limit => {
          (lines[..limit].join("\n"), Some(lines.len() - limit))
      }
      _ => (lines.join("\n"), None),
  };

  let final_snippet = match truncated {
      Some(n) => format!("{}\n// … ({} lines omitted)", snippet, n),
      None => snippet,
  };
  No AST required — line-count truncation is sufficient. The // … (N lines omitted) suffix gives the model enough signal to know the function continues and to use a
  targeted Read on the specific line range if needed.

  Notes

  - Especially valuable against the 4-call budget: a 30-line excerpt across 4 symbols costs the same budget as one full 300-line body, but covers 4 entry points
  instead of 1.
  - Natural companion to a complexity-aware call budget if that's ever considered — both address "how much usable information per call" from different angles.

oscaruiz:
## Summary

- Adds an optional `max_code_lines` parameter to `tokensave_context`, giving callers a predictable structural cap such as "the signature and first 20 lines."
- Keeps the existing approximately 1500-byte limit and `tokensave_body node_id=...` follow-up handle; both limits apply, with the tighter one winning.
- Preserves line and UTF-8 boundaries, leaves shorter snippets unchanged, and clamps explicit values to 1–1000.
- Fixes #324.

## Motivation

`tokensave_context` already prevents unbounded code output with an approximately 1500-byte limit, but a byte budget cannot express a deliberately short structural prefix. A caller asking for enough code to understand a signature and the beginning of a body may receive more lines than needed while the snippet still remains within that byte budget.

`max_code_lines` complements the existing protection instead of replacing it. Callers can request a short, predictable excerpt, while the byte limit remains the absolute size bound and `tokensave_body` remains the explicit path to the complete body.

## Scope and decisions

| Decision | Choice | Why |
| --- | --- | --- |
| Existing byte limit | Preserved | It remains the absolute per-snippet size bound |
| Limit interaction | The tighter limit wins | A requested line count must not bypass the byte budget |
| Default | No line cap | Omitting the new parameter preserves existing behavior |
| Explicit range | Clamp to 1–1000 | Avoid empty snippets and unbounded numeric input |
| Truncation boundary | Whole lines and valid UTF-8 | Never return a split line or invalid string |
| Follow-up behavior | Reuse `tokensave_body node_id=...` | Keeps the existing way to retrieve the full body |
| Omitted-line count | Not calculated | Avoids a second full pass; the follow-up handle is actionable |

## Changes

- `src/mcp/tools/definitions.rs`: exposes `max_code_lines` in the `tokensave_context` input schema and documents its interaction with the byte limit.
- `src/mcp/tools/handlers/graph.rs`: parses the optional value and clamps it to 1–1000.
- `src/types.rs`: carries the optional cap in `BuildContextOptions`; `None` preserves the previous behavior.
- `src/context/builder.rs`: applies the line prefix before the existing UTF-8-safe byte truncation and reuses the existing follow-up marker.
- `src/context/builder.rs`: adds coverage for exact line boundaries, snippets below the cap, byte/line interaction, and UTF-8-safe truncation.
- `CHANGELOG.md`: documents the parameter under `[Unreleased]`.

## Test plan

- [x] Line-cap regression: a long snippet is limited to the requested number of lines and includes its `tokensave_body` handle.
- [x] Short-snippet regression: code below both limits is returned verbatim.
- [x] Limit interaction: the existing byte limit still wins when it is tighter than the requested line cap.
- [x] Boundary coverage: exact line counts, trailing newlines, no trailing newline, and the helper's defensive zero-limit behavior are covered.
- [x] Existing UTF-8 truncation coverage remains green.
- [x] On the final rebased commit (`62b7585`), `cargo test --workspace` completed with 2,920 passed and 0 failed across 108 suites using the default `full` feature set.
- [x] `cargo clippy --workspace --all-targets` completed with no warnings.
- [x] `cargo fmt --all -- --check` completed successfully.

A manual MCP-server round trip was not made a gate: the truncation policy is a pure transformation covered directly at its boundary, while the full workspace suite compiles the schema, handler, options, and context-builder wiring together.

## Checklist

- [x] No dependency changes
- [x] Breaking changes documented: none; omitting `max_code_lines` preserves existing behavior


rNoz:
Thanks, Oscar! Reviewed locally and now merged! Your first contrib! ¡Disfruta del día!

PR #332, opened 2026-07-29
fix(serve): parse MCP root URIs on Windows (#331)

### LINKED ISSUE #331: MCP roots discovery fails for Windows file URIs
### tokensave version

7.8.1

### Operating system

Windows

### Installation method

Built from source

### What happened?

  When `tokensave serve` cannot discover a project from its current working directory or the global
  project lookup, it falls back to the `roots` array from the MCP `initialize` request.

  A Windows project root is normally sent as a file URI such as `file:///D:/Dev/project`. The current
  implementation removes only the `file://` prefix, producing `/D:/Dev/project`. That is not the
  registered Windows path `D:\Dev\project`, so both the exact-project comparison and the directory
  discovery walk fail.

  The same fallback does not decode percent-encoded paths. For example, `file:///C:/My%20Project` is
  treated as a path containing the literal characters `%20`.

  As a result, the MCP server can report that no TokenSave index exists even though the project is
  already initialized.

### Steps to reproduce

1. Use an initialized project located at a Windows drive path, for example `D:\Dev\project`.
2. Launch `tokensave serve` from outside that project, so current-directory discovery does not
  identify it.
3. Make the MCP client send this root in its `initialize` request:

     ```json
     {
       "jsonrpc": "2.0",
       "id": 1,
       "method": "initialize",
       "params": {
         "roots": [
           {
             "uri": "file:///D:/Dev/project",
             "name": "project"
           }
         ]
       }
     }

  existing configuration error:

  no TokenSave index found at '<current directory>' and no projects registered in the global database
  — run 'tokensave init' in your project first.

### Error output

```shell

```

### Component

MCP server

### Additional context

- Architecture: x86_64
- Reproduced against the tokensave 7.8.1 source tree on Windows.

oscaruiz:
## Full Issue

When neither the working directory nor the global database identifies a project, `serve` peeks at the `initialize` request and resolves the workspace from its `roots` array — the path that covers a VS Code multi-folder workspace, or any client launched from outside the project.

It stripped a bare `file://` prefix, which is correct only on Unix. A Windows root arrives as `file:///D:/Dev/app`, leaving `/D:/Dev/app`: a path that matches no registered project and that `discover_project_root` cannot walk. The fallback silently found nothing and the server exited telling the user to run `tokensave init` in a project that was already indexed. Percent-encoding was ignored too, so any path containing a space (`file:///C:/My%20Project`) failed on every platform.

The URI is now decoded — on bytes, so a percent-encoded multi-byte character reassembles correctly — and `file://localhost/x`, the spelled-out form of `file:///x`, is recognized as local in any casing, since host names are case-insensitive. Registered
projects are compared canonically rather than literally, so the same directory spelled `D:\Dev\app` and `D:/Dev/app` is recognized as one. Other schemes and non-local authorities are skipped instead of being turned into relative paths, while a plain path with no scheme still passes through for clients that send one.

Whether a leading `/C:` is a drive letter or a directory named `C:` depends on the host, not the URI — `/C:/x` is a legal POSIX path — so that rule is keyed on `cfg!(windows)`. It lives in a parameterized `root_uri_to_path_on` rather than behind `cfg!` directly, because CI runs on Linux and gating it there would leave the Windows case, the one this commit exists for, untested.

This is the inverse of `agents::kiro::file_resource_uri`, which already produced exactly this URI form; only the decoding direction was missing.

## Summary

- Parse local `file://` URIs from MCP `initialize.roots`, including Windows drive paths, percent-encoded paths, and an explicit `localhost` authority.
- Preserve host-specific path semantics: Windows drive roots lose the URI-only leading slash, while valid POSIX paths such as `/C:/x` remain unchanged.
- Match registered projects canonically when equivalent path spellings differ, while retaining literal matching when canonicalization is unavailable.

## Motivation

When `tokensave serve` cannot identify a project from its working directory or the global project lookup, it falls back to the MCP client's `initialize.roots`.

That fallback previously removed only the `file://` prefix. On Windows, a standard root such as `file:///D:/Dev/project` therefore became `/D:/Dev/project`, which matched neither the registered project nor a discoverable filesystem path. Percent-encoded paths and an explicit `localhost` authority failed for the same reason.

Fixes #331.

## Changes

- `src/serve.rs`: decode supported local file URIs, apply the Windows drive-prefix rule only for Windows hosts, and skip unsupported schemes and non-local authorities.
- `src/serve.rs`: accept both literal and canonical equivalence between registered and requested project paths, returning the registered spelling.
- `src/serve.rs`: add focused coverage for Windows and POSIX path policies, percent encoding, authorities, malformed input, and registered-project matching.
- `CHANGELOG.md`: document the fix under `[Unreleased]`.

## Test plan

- [x] `cargo test --workspace` passes

On the final commit (`ac75f98`), the full workspace suite completed with 2,929 passed and 0 failed across 108 suites. Clippy completed with no warnings or errors, and `cargo fmt --all -- --check` completed without differences.

The Windows URI policy is exercised through the platform-parameterized parser. GitHub CI will run the full workspace suite on Windows, macOS, and Linux.   A native Windows stdio MCP round trip was completed using a registered project whose path contained spaces. The percent-encoded root resolved correctly; `initialize`, `notifications/initialized`, `ping`, and `tokensave_status` completed successfully; the server remained alive after the handshake and exited cleanly with code 0.

## Checklist

- [x] Breaking changes documented: none; existing plain paths and supported Unix file roots retain their behavior


rNoz:
@oscaruiz Thanks for the fix! I've rebased and polished this branch (including adding struct-based matching and hardening the percent-decoder). Moreover, while digging in, I have discovered another bug, so took the time to patch it and prepare another test for it.

So, regarding the original issue, could you verify that this genuinely solves your Windows MCP issue?

The idea is to verify in your own Windows machine if a project path with spaces (e.g., `C:\My Project`) now resolves correctly in your client. I've verified the logic via unit tests, but a real Windows integration test would be the final green light :)

When done, check the remaining "PR task" and ping me.

rNoz (approved):

rNoz (changes requested):
He is evaluating the remaining question and found a blocker. Waiting for his changes and report.

oscaruiz:
Hi @rNoz, thanks for the review and for polishing the original fix!

  I've now completed the manual Windows verification using a real stdio MCP round trip with a registered project whose path contained spaces. The client sent the root as a percent-encoded URI (`file:///D:/Dev/tokensave/.manual%20PR332%20final`),
  while `tokensave serve` was launched from outside the project with multiple projects registered.

  Your follow-up commit `246c52e`, which fixed the authority slice offset for `127.0.0.1` and `[::1]`, is preserved unchanged.

  The root resolved correctly and, while testing the complete handshake, I found and fixed two additional issues:

  - The replayed `initialize` response was not newline-terminated, so a line-framed MCP client waited indefinitely.
  - In debug builds, `run()` treated the replayed `initialize` as evidence that the server loop had already run and panicked before accepting the next request.

  I also hardened percent-decoding so sequences that decode to invalid UTF-8 (`%C3`, `%FF`, or truncated multibyte sequences) reject the root instead of collapsing into a lossy U+FFFD path.

  The additional commits applied on top are:

  - `b5f0bab` — newline-terminate the replayed `initialize` response
  - `870ac54` — allow `run()` after a replayed `initialize`
  - `663ec9d` — reject invalid UTF-8 root URIs

  The final manual round trip completed successfully: `initialize`, `notifications/initialized`, `ping`, and `tokensave_status` all worked; the server remained alive after the handshake, exited cleanly with code 0, and produced no panic.

  CI is now green on Windows, Linux, and macOS, including Format and Clippy. I've also marked the remaining manual-testing task as completed in the PR description.

  Thanks again for the guidance and for catching and improving the original URI handling. Since these three commits were added after your approval, could you please take another look when you have a chance?

rNoz:
Thanks @oscaruiz, the live Windows round trip is exactly the verification this needed, and both handshake bugs you found are real.

I have taken your three follow-up commits, rebased the branch onto current `master` (it now sits on top of #336, no conflicts), and tidied the history before we land it:

- The two `initialize` commits (`b5f0bab` newline framing and `870ac54` the `run()` guard) are folded into one, since they are two symptoms of the same replayed-initialize path and neither stands alone. Full authorship stays yours; I only combined them and wrote a single message covering both.
- `663ec9d` (reject invalid UTF-8) is kept as its own commit with a message body added, since it is a distinct concern from the handshake fix.
- The other three commits (your original URI fix and my two review commits - the refactor optimization improvements + bugfix discovered) are unchanged.

The final tree is byte-for-byte identical to your six-commit branch, so this is purely a history cleanup, not a code change. `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets`, and the full `cargo test` suite are green locally.

I will push the rebased history to the PR branch, merging next.

rNoz (approved):

PR #336, opened 2026-07-30
fix(hooks): Activate PreToolUse guardrails anywhere inside an indexed project

### LINKED ISSUE #335: PreToolUse guardrails are silently inactive in subdirectories of an indexed project
### Type

Bug

### tokensave version

7.7.0

### Operating system

macOS

### Installation method

Built from source

### What happened?

`HookEnv::from_runtime()` decides whether the grep/agent guardrails are active by probing `.tokensave/tokensave.db` in the hook process's own working directory, with no ancestor walk. Everything else in tokensave resolves the project by walking upward: `serve`, `sync`, and `status` go through `config::resolve_path_with_discovery`, the MCP server root resolver falls back to `config::discover_project_root`, and Kiro's `postToolUse` sync hook calls `discover_project_root` on the event `cwd`.

The result is a silent split. An agent started in a subdirectory of an indexed repository (a package inside a monorepo, a `crates/<name>` member, or simply `cd src` before launching) gets working tokensave MCP tools, because the server discovers the enclosing project, and a completely disabled guardrail, because the hook does not. Every symbol-shaped `grep` in that session passes through with no savings, no redirect, and no explanation.

The same probe also ignores the working directory the harness reports. Claude Code, Kiro, and Factory Droid all send a top-level `cwd` in the hook event; only Kiro's `postToolUse` and `userPromptSubmit` handlers read it. The `preToolUse` adapters resolve the project from wherever the hook process happens to have been spawned, so one harness resolves the project two different ways inside the same binary.

### Steps to reproduce

1. `tokensave init` at the root of a repository, then `cd src`.
2. `printf '%s' '{"tool_input":{"command":"grep -rn handle_request hooks.rs"}}' | tokensave hook-pre-tool-use`.
3. Observe `{"permission":"allow"}`. The same call from the repository root is denied.

### Error output

```text
$ cd <root>/src
$ printf '%s' '{"tool_input":{"command":"grep -rn target_looks_like_code hooks.rs"}}' | tokensave hook-pre-tool-use
{"permission":"allow"}

$ tokensave tool search target_looks_like_code   # the MCP tools work fine from here
```

### Component

Agent integration (Claude, Cursor, etc.)

### Proposed behavior

1. Discover the indexed root with `config::discover_project_root`, the same ancestor walk the rest of the codebase uses, so the guardrail is active exactly where the MCP tools are available.
2. Resolve that discovery from the event's top-level `cwd` when the harness reports one, with the hook process's working directory as the fallback, in all three PreToolUse adapters (Claude, Kiro, Droid). The reported value is agent-supplied, so accept it only when it is an absolute path to an existing directory; a relative, stale, or non-string value must fall back rather than move the ancestor walk somewhere nobody meant.

### Non-goals

- Do not change the target classifier, the pattern classifier, or prompt classification.
- Do not change Droid's exact built-in `explorer` task guard from PR #282, or the glob-over-path precedence from PR #280.
- Do not weaken the `TOKENSAVE_DISABLE_GREP_HOOK` escape hatch. It belongs to the hook process and must survive any event-cwd override.
- Do not read the database or contact the MCP server from the hook path. Discovery stays a filesystem walk.
- Do not add a depth cap or a git-root boundary to discovery. Widening activation also arms the guardrail for everyone whose `$HOME` is itself indexed, which is intended: the MCP tools are already served from that same discovered index there, so the two now agree, and a cap here alone would make the hook disagree with `serve`, `sync`, `status`, and the MCP root resolver.

### Acceptance

- A working directory inside an indexed project activates the guardrail and reports the discovered root.
- A working directory outside any indexed project leaves the guardrail inactive.
- Claude, Kiro, and Droid events resolve the project from their own `cwd`; a missing, blank, whitespace, relative, non-existent, non-directory, or non-string `cwd` falls back to the hook process environment.
- An event `cwd` outside any index leaves the guardrail inactive for all three adapters, even when the hook process sits in a project.
- The opt-out still bypasses every redirect, and the Droid `explorer` and non-`explorer` task behaviors are unchanged.
- Focused and full Rust gates pass, followed by a real headless hook proof with the hook process deliberately started outside the project.

### Relationship to the absolute project-root target fix

This is the activation half of the guardrail's project resolution and is intentionally a separate change from the target-classification fix ("Redirect symbol grep when its target is the absolute project root"). That one changes which targets are classified as code; this one changes where the guardrail runs at all. Landing the classifier fix first keeps each diff reviewable, and the two compose: from a subdirectory, an absolute target equal to the discovered ancestor root is redirected.

rNoz:
## Summary

- Fixes #335
- Discover the indexed project in `HookEnv` with `config::discover_project_root`, the ancestor walk `serve`, `sync`, `status`, and the MCP root resolver already use, so the guardrail is active exactly where the tokensave MCP tools are available.
- Resolve that discovery from the event's own top-level `cwd` in all three PreToolUse adapters (Claude, Kiro, Droid), accepting it only when it is an absolute path to an existing directory and keeping the hook process's working directory as the fallback.
- Rename `HookEnv::cwd_has_tokensave_db` to `in_tokensave_project`, because the flag no longer describes the working directory alone.

## Motivation

`HookEnv::from_runtime()` probed `.tokensave/tokensave.db` in the hook process's working directory only. Everything else resolves a project by walking upward: `config::resolve_path_with_discovery` for `serve`/`sync`/`status`, `config::discover_project_root` in the MCP server's root resolver, and the same function in Kiro's `postToolUse` sync hook. An agent started in a subdirectory of an indexed repository (a monorepo package, a `crates/<name>` member, or a plain `cd src` before launch) therefore had working MCP tools and a silently disabled guardrail, so every symbol-shaped grep in that session passed through with no savings and no explanation.

The event `cwd` half is the same inconsistency one layer up. Claude Code, Kiro, and Droid all report the session working directory at the top level of the hook event, and Kiro's own `postToolUse`/`userPromptSubmit` handlers already use it, while the `preToolUse` adapters resolved the project from wherever the hook process happened to be spawned. Reading the field the harness sends makes the decision depend on the project that emitted the tool call instead of on hook process placement.

This is deliberately the activation half only. Target classification is unchanged here.

## Changes

- `src/hooks.rs`: `from_runtime()` delegates to a new `from_runtime_at(Option<&Path>)` that discovers the root through `config::discover_project_root`. A private `HookEnv::for_event()` re-resolves the root from a parsed event, preserving `disable_grep_hook`, and returns the base environment unchanged when there is no usable `cwd`.
- `src/hooks.rs`: `kiro_event_cwd()` is reduced to a thin wrapper over a shared `event_cwd(&Value)` parser, which the Claude, Kiro, and Droid adapters use. The parser trims whitespace and accepts only an absolute path to an existing directory, so an agent-supplied relative, stale, or non-string value cannot move the ancestor walk. Droid's parsed event is reused, so nothing is parsed twice.
- `src/hooks.rs`: `cwd_has_tokensave_db` becomes `in_tokensave_project` at its four decision sites.
- `tests/hooks_test.rs`: fourteen regressions over real temporary indexed projects, covering ancestor discovery from a nested directory, no discovery outside a project, event-cwd activation for Claude/Kiro/Droid, deactivation for all three when the reported directory is outside any index, an event cwd below the root, the opt-out surviving an override that points at an indexed project, the Droid `explorer` and `worker` invariants from #282, an absolute ancestor-root target redirecting from a subdirectory, and one table-driven case set proving the fallback for a missing, blank, whitespace, relative, non-existent, non-directory, null, numeric, and object `cwd`.

## Test plan

- [x] `cargo test --test hooks_test` passes, 147 passed. The 8 core activation regressions fail on the unpatched resolver and pass after it.
- [x] `cargo test` passes, 2809 tests across 107 suites.
- [x] `cargo clippy --workspace --all-targets` is clean.
- [x] Real headless proof with the release binary, including a run whose hook process was deliberately started outside the project.

<details>

<summary>Headless proof (release binary)</summary>

```text
before (7.6.2 build), hook process cwd = <root>/src
  {"tool_input":{"command":"grep -rn target_looks_like_code hooks.rs"}}   -> allow

after, hook process cwd = <root>/src, no event cwd
  same event                                                             -> deny

after, hook process cwd = /tmp/ts-proof/outside (outside every index)
  droid,  event cwd = <root>/src                                         -> deny  (exit 2)
  droid,  event cwd = "src" (relative)                                   -> allow (exit 0)
  droid,  event cwd = <root>/does-not-exist                              -> allow (exit 0)
  droid,  event cwd = <root>/Cargo.toml (a file)                         -> allow (exit 0)
  droid,  event cwd outside any index                                    -> allow (exit 0)
  droid,  Task/explorer, event cwd = <root>                              -> deny  (exit 2)
  droid,  Task/worker,   event cwd = <root>                              -> allow (exit 0)
  kiro,   delegate, event cwd = <root>                                   -> deny  (exit 2)
  kiro,   delegate, event cwd outside any index                          -> allow (exit 0)
  claude, Grep path=src, event cwd = <root>                              -> deny
  claude, Grep path=src, event cwd outside any index                     -> allow
  droid,  event cwd = <root>, TOKENSAVE_DISABLE_GREP_HOOK=1              -> allow (exit 0)
```

</details>

## Scope and known behavior

- **Widening activation widens a hard deny.** Anyone whose `$HOME` (or another broad ancestor) is itself indexed now has the guardrail armed in every descendant directory. That is intended, because the tokensave MCP tools are already served from that same discovered index everywhere below it, so the guardrail and the tools it redirects to now agree; the `TOKENSAVE_DISABLE_GREP_HOOK` escape hatch is unchanged for anyone who wants the old behavior in a given session.
- **Discovery has no depth cap**, matching `serve`, `sync`, `status`, and the MCP root resolver. Capping it here alone would make the hook disagree with the tools again.
- **Target classification is untouched.** The bucket rules, the pattern classifier, and #280/#282 precedence are unchanged.

## Checklist

- [x] Breaking change documented: `HookEnv::cwd_has_tokensave_db` is renamed to `in_tokensave_project`. It is a public field on a public type, so any out-of-tree caller constructing `HookEnv` directly must adjust; in-tree it is a hook-internal type built through `from_runtime()`, and the rename plus the new `from_runtime_at` constructor are the only surface changes.


rNoz (commented):
Adding some notes on the compatibility surface, since the diff renames a public field and widens where the guardrail fires.

The rename (`HookEnv::cwd_has_tokensave_db` to `in_tokensave_project`) is confined to `src/hooks.rs` and `tests/hooks_test.rs`; nothing else in the crate references `HookEnv`. No CLI flag, config key, hook event JSON field, or MCP tool signature changes, so there is no user-facing break. The only affected caller would be out-of-tree code constructing `HookEnv` directly, which the checklist already flags.

The behavior change is the activation widening. I ran the release binary headless with the hook process deliberately started outside every index and confirmed: a symbol grep is denied when the event `cwd` (or the process cwd) resolves to an indexed project and allowed when it does not, for the Claude, Kiro, and Droid adapters. A missing, blank, relative, non-existent, non-directory, or non-string `cwd` falls back to the process environment rather than moving the ancestor walk. `TOKENSAVE_DISABLE_GREP_HOOK=1` still bypasses the redirect from inside an indexed project, so the opt-out is intact.

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets`, and `cargo test` (2809 tests) are green on the rebased branch.

Issue #292, opened 2026-07-25
branch_fallback=true doesn’t populate tokensave_files results


paulvno:
### tokensave version

7.6.2

### Operating system

Windows

### Installation method

GitHub release binary

### What happened?

When the current branch isn’t tracked, the server serves from the nearest tracked ancestor (usually main). tokensave_context and tokensave_search return results from the ancestor’s index, but tokensave_files returns nothing — it doesn’t inherit from the fallback branch’s file list.

Impact: tokensave_files on an untracked branch always shows 0 files, giving no indication of which files are indexed (even though queries work).

Source pointer: handle_files() in src/mcp/tools/handlers/info.rs L104 calls get_all_files() from the branch DB — without fallback inheritance, the untracked branch’s DB has no file entries. The root cause is in how branch fallback populates (or doesn’t) the file list vs. the node/edge graph.

Suggested fix: In handle_files, when get_all_files() returns empty and a branch fallback is in effect, repeat the call against the fallback branch’s DB — same pattern the node/edge graph queries already use for context/search. Minimal diff; no new API surface needed.

### Steps to reproduce

1.
2.
3.

### Error output

```shell

```

### Component

None

aovestdipaperino:
Thanks for the report @paulvno. I dug into this and I could not reproduce it — and I think the diagnosis in the issue rests on a wrong model of how branch fallback works, so I want to lay that out before we chase the wrong fix.

Branch fallback is not a per-query feature. It happens once, at DB-open time, in `TokenSave::resolve_db_for_branch` (`src/tokensave/mod.rs:333-388`): when the current branch is untracked, the whole `TokenSave` instance opens the nearest tracked ancestor's DB file. There is no split where the node/edge graph inherits the fallback DB but the file list does not — `handle_files` calls `get_all_files()` against the very same DB handle that context and search read their nodes from. Relatedly, `branch_fallback` is not an input option anywhere; it exists only as an output field of `tokensave_status`.

I also tried it live: created an untracked branch off master in this repo and drove `tokensave serve` over stdio. `tokensave_status` emitted the expected fallback warning, and `tokensave_files {"path":"src/mcp"}` in the same session returned 39 indexed files with full per-file symbol counts.

Caveat: I tested on macOS against 7.6.0, and you are on Windows 7.6.2. Nothing in this path is platform-specific, but I did not test Windows, and the most plausible real cause I can see is Windows-specific. So, three things that would settle it:

[...]

PR #343, opened 2026-07-30
fix(branch): snapshot the parent DB when tracking a branch (#292)

oscaruiz:
## Summary

  - Tracking a branch (`tokensave branch add` and the transparent auto-track path) now snapshots the parent DB with `VACUUM INTO` instead of copying only the `.db` file, so committed rows still sitting in the parent's uncheckpointed WAL are no longer silently dropped from the copy.
  - The snapshot publishes through a unique temp file + atomic rename (same idiom as the #340 `branch-meta.json` fix), so racing trackers each produce a complete DB and a stale destination from an interrupted attempt is replaced, never inherited.
  - Adds regression tests proving WAL-only rows survive the copy, including under a concurrent reader that blocks checkpointing.

  ## Motivation

  Refs #292 — likely its root cause; verification is pending from the reporter in the issue thread, so this deliberately does not close it.

  A long-lived MCP server keeps the parent DB open in WAL mode, so committed rows can sit exclusively in the `-wal` sidecar at copy time. A bare `fs::copy` of the `.db` then produces a copy that opens clean (`quick_check` passes) but is silently truncated: checkpointed tables keep answering `tokensave_context`/`tokensave_search` while e.g. `files` comes up empty — `tokensave_files` reports 0 files. The old comment in `track_branch_copy` already named the hazard and the remedy ("checkpoint-before-copy applied to BOTH paths together"); this completes that upgrade path with a stronger primary mechanism, since sequencing a checkpoint plus two separate file copies still can't guarantee consistency against a concurrent writer.
  - Adds regression tests proving WAL-only rows survive the copy, including under a concurrent reader that blocks checkpointing.

  ## Motivation

  Refs #292 — likely its root cause; verification is pending from the reporter in the issue thread, so this deliberately does not close it.

  A long-lived MCP server keeps the parent DB open in WAL mode, so committed rows can sit exclusively in the `-wal` sidecar at copy time. A bare `fs::copy` of the `.db` then produces a copy that opens clean (`quick_check` passes) but is silently truncated: checkpointed tables keep answering `tokensave_context`/`tokensave_search` while e.g. `files` comes up empty — `tokensave_files` reports 0 files. The old comment in `track_branch_copy` already named the hazard and the remedy ("checkpoint-before-copy applied to BOTH paths together"); this completes that upgrade path with a stronger primary mechanism, since sequencing a checkpoint plus two separate file copies still can't guarantee consistency against a concurrent writer.

  ## Changes

  - `src/branch.rs`: new shared `copy_branch_db` helper — primary path is `VACUUM INTO` (transactionally consistent snapshot, main file + WAL contents, safe mid-commit of a concurrent writer) into a pid+nanos temp file renamed atomically over the destination; fallback (parent not readable SQLite) is best-effort `wal_checkpoint(TRUNCATE)` + plain file copy, bringing the `-wal` sidecar along when the checkpoint couldn't empty it so SQLite recovers it on first open. `track_branch_copy` is now `async` and uses the helper.
  - `src/commands.rs`: `branch add` uses `copy_branch_db` instead of `std::fs::copy`.
  - `src/tokensave/mod.rs`: awaits the now-async auto-track call.
  - `tests/branch_wal_copy_test.rs`: new regression tests (see below).
  - Non-obvious: a non-UTF-8 destination path skips the snapshot layer — a lossy conversion would make `VACUUM INTO` "succeed" at a mangled replacement-character path — and takes the lossless file-copy fallback instead.

  ## Test plan

  - [x] `cargo test` passes
  - [x] `cargo clippy` has no new warnings (`--all-targets`, pedantic-clean)

  The reported scenario is reproduced directly by the new integration tests: 100 rows committed in WAL mode with the writer connection held open (so they live only in the `-wal` sidecar) survive `copy_branch_db`; a second test holds a concurrent read transaction that pins the WAL and asserts the copy still carries the rows via the snapshot path (no `-wal` sidecar next to the copy). The existing `track_branch_copy` unit test now also exercises the non-SQLite-parent fallback.

  Full suite run locally (WSL): all targets green except two pre-existing environment failures unrelated to this change — `run_affected_tests_dispatches_directly_changed_test_files` (load-dependent 60s timeout; passes in isolation) and `serve_disable_test::server_without_disable_reaches_the_mcp_transport` (fails identically on master in this WSL setup; the exercised flow verified manually against this branch's binary, exits cleanly in 0.4s). CI should confirm on native runners.

  ## Checklist

  - [x] Breaking changes documented (if any) — library-level only: `branch::track_branch_copy` is now `async`; no CLI or MCP surface change.


rNoz (approved):
Thanks @oscaruiz for the contribution, fully reviewed and tested locally during the last hours. Checked the issue discussion (it was longer than expected) and implementation. `VACUUM INTO` correctly preserves committed WAL data, both branch-tracking paths use the shared helper, and the regression tests cover active WAL and pinned-reader cases. 

A personal consideration: comments are long... but explain concurrency and fallback constraints that are not obvious from the code. It could help for maintainability.

Rebased to have latest and approving it! I'll merge as CI passes.

PR #353, opened 2026-07-31
feat(accounting): Import Factory Droid usage into cost and monitor

### LINKED ISSUE #352: Account for Factory Droid local usage in cost and monitor
### Problem

`tokensave cost` and `tokensave monitor` read Claude Code sessions but do not account for Factory Droid sessions. A user running both tools sees Claude USD while Droid token usage and Factory credits are omitted, and there is no per-agent breakdown.

### Local source

- `.factory/sessions/**/*.settings.json` contains one cumulative snapshot per Droid session.
- The exclusive `tokenUsage` object exposes `inputTokens`, `outputTokens`, `cacheCreationTokens`, `cacheReadTokens`, `thinkingTokens`, and nullable `factoryCredits`.
- `inclusiveTokenUsage` includes child sessions whose settings files are imported separately, so using it would double-count.
- `providerLockTimestamp` is stable across copied snapshots and can be used as the session timestamp bucket; file modification time is not stable.
- Active snapshots grow cumulatively, so ingestion must update an existing session only when its counters advance without regressing.

### Proposed behavior

1. Add `agent` and nullable `credits` fields to accounting rows, defaulting existing rows to `agent = 'claude'`.
2. Import exclusive Droid usage from local settings files and upsert by session ID.
3. Report Factory credits as an agent-native unit in `cost --by-agent`; never convert credits to USD or add them to Claude cost.
4. Preserve unavailable credits as null rather than treating them as zero.
5. Add Droid usage to the monitor only when a local Droid source is discovered.
6. Preserve the existing `cost` and monitor output when Droid is not installed or no Droid source exists.

### Scope

This remains local and offline. It does not read browser storage, authenticate to Factory services, call `/limits`, import rotated logs that omit credits, or introduce a plugin framework.

rNoz:
## Summary

- Import exclusive Droid session usage from `.factory/sessions/**/*.settings.json`.
- Add `cost --by-agent` with separate USD and agent-native credit fields.
- Report Factory credits without assigning an exchange rate or adding them to Claude USD.
- Preserve the existing `cost` and monitor output when no Droid source exists.
- Closes #352

## Motivation

TokenSave's accounting commands currently read Claude Code sessions only. On a machine that also runs Factory Droid, the Droid portion is omitted even though local settings snapshots contain token counters and Factory credits. This change adds local Droid accounting without provider authentication or a guessed USD conversion.

## Data handling

| Decision | Reason |
|---|---|
| Read settings snapshots instead of rotated logs | Settings contain the exclusive token counters and `factoryCredits`; rotated logs do not contain credits. |
| Import `tokenUsage`, not `inclusiveTokenUsage` | Inclusive usage contains child sessions that have their own settings files and would be counted twice. |
| Store credits separately from USD | Factory does not publish an exchange rate suitable for this calculation. |
| Upsert monotonically by session ID | Active session snapshots are cumulative and may be copied with an unstable modification time. |
| Use `providerLockTimestamp` | It remains stable when the settings file is copied and provides a consistent range bucket. |
| Keep null credits unavailable | A missing credit field is unknown, not measured zero. |
| Keep no-Droid behavior unchanged | Users without a local Droid source retain the existing two-line monitor and existing `cost` output. |

## Changes

- Add `agent TEXT NOT NULL DEFAULT 'claude'` and nullable `credits` columns to `turns`, including migration for existing databases.
- Add `src/accounting/droid.rs` for discovery, parsing, deduplication, and monotonic upsert of Droid snapshots.
- Add `cost_by_agent_since` and `cost --by-agent` table, JSON, and CSV output.
- Keep Droid credits out of USD totals and model-cost views.
- Add a Droid monitor line only when Droid coverage exists; absent Droid sources do not add a line or change the legacy labels.
- Run the stop hook through Claude-only ingestion so ordinary Claude hook activity does not rescan every Droid snapshot.

## Test plan

- [x] Parse the real settings shape and map every exclusive token counter.
- [x] Ignore `inclusiveTokenUsage` to avoid child-session double counting.
- [x] Preserve null `factoryCredits` while retaining token usage.
- [x] Reject regressing cumulative snapshots and update advancing snapshots.
- [x] Keep repeated ingestion idempotent.
- [x] Migrate an existing `turns` table without `agent` or `credits`.
- [x] Keep Droid credits separate from USD in table, JSON, and CSV output.
- [x] Preserve the exact pre-Droid monitor lines when Droid is absent.
- [x] Preserve existing human and JSON `cost` output when Droid is absent.
- [x] Hide stale Droid rows if a previously discovered source is removed.
- [x] Isolate CLI test homes across Unix and Windows.
- [x] Render Droid-only activity, partial sources, and unavailable credits without a dollar sign.
- [x] Reject malformed timestamp components without panicking.
- [x] Mark Claude coverage partial when source discovery or file I/O fails.
- [x] Run `cargo fmt --all -- --check`.
- [x] Run `cargo clippy --workspace --all-targets`.
- [x] Run the complete test suite.

## Local verification

- Verified against Factory Droid 0.175.1 using a local session with `minimax-m3`.
- The Droid-reported token counters matched the imported database row for input, generated and thinking output, cache writes, and cache reads.
- `tokensave cost --by-agent today` displayed Droid credits and raw tokens with USD unavailable.
- The publication material contains no session IDs, project paths, prompts, balances, browser credentials, or raw logs.

## Checklist

- [x] User-visible behavior documented in `CHANGELOG.md`
- [x] No new dependency
- [x] No browser credentials or provider API calls
- [x] No Factory credits included in USD totals
- [x] No-Droid behavior protected by regression tests

Not merging by myself directly because it contains a lot of work (and yes, plenty of local testing), but better to have another pair of eyes here.


aovestdipaperino:
Merged — thank you @rNoz! Really appreciate the care on the data handling here: reading Droid session snapshots and reporting Factory credits without inventing a USD rate is the right call. `cost --by-agent` is a nice addition. 🙏

PR #354, opened 2026-08-01
feat(ruby): resolve static receiver calls

janstol:
## Summary

- Resolve Ruby calls with statically provable constant or `self` receivers to singleton methods.
- Preserve lexical constant lookup and shadowing rules, and upgrade legacy Ruby index rows during incremental sync.
- Avoid class/module-body call edges from blocks that change `self`.

## Motivation

Ruby call references now preserve receiver evidence, but the resolver could not use that evidence to distinguish singleton methods from instance methods. Existing indexes also stored singleton definitions as ordinary methods, and class-body `self` calls required caller attribution that respected Ruby block semantics.

## Changes

- Persist enclosing singleton definitions with the `singleton_method` kind and include them in existing query surfaces.
- Resolve absolute constants, lexical constants, and supported `self` calls only when ownership is positive and unambiguous.
- Re-extract legacy indexed Ruby files once, recording the repair marker only after all scheduled files complete.
- Attribute direct class/module-body calls through the block-aware traversal so eval and Concern blocks cannot fabricate enclosing-class callers.
- Add extraction, resolver, database-query, incremental-sync, and end-to-end regression coverage.

## Test plan

- [x] `cargo test --workspace` passes
- [x] Tested manually: Ruby 3.4.7 receiver and block-scope probes, predicate mutation checks, and a scratch `tokensave init`/`sync` SQLite query confirming one valid edge and no false edges from self-retargeting blocks

## Checklist

- [x] No breaking changes


aovestdipaperino:
Merged — thank you @janstol! Resolving static constant/`self` receivers to singleton methods (and upgrading legacy index rows on incremental sync) is exactly the missing piece to make the receiver evidence useful. Clean, well-tested work. 🙏

Issue #318, opened 2026-07-26
Bulk indexing can leave the database with duplicate edges and no edge indexes


janstol:
## Summary

A TokenSave project database can end up with hundreds of thousands of exact duplicate edge rows and no indexes on the `edges` table after bulk indexing. Subsequent incremental syncs become extremely slow even when few files changed.

In the affected database:

| Metric | Value |
| --- | ---: |
| Indexed files | 408 |
| Nodes | 15,813 |
| Edges | 943,151 |
| `calls` edges | 777,844 |
| Exact duplicate edge rows | 914,537 |
| Approximate unique edge tuples | 28,614 |
| Database size | 165 MB |
| Last recorded sync duration | 913,081 ms |
| Edge indexes | None |

Two separate, unaffected Ruby-heavy projects completed their last syncs in under two seconds. The larger comparison project had:

| Metric | Value |
| --- | ---: |
| Indexed files | 4,724 |
| Nodes | 24,697 |
| Edges | 27,567 |
| Exact duplicate edge rows | 0 |
| Edge indexes | All four, including `idx_edges_unique` |
| Database size | 60 MB |
| Last recorded sync duration | 455 ms |

That project has more than eleven times as many indexed files and more nodes than the affected project. This strongly suggests the slowdown is driven by edge duplication and missing indexes rather than project size or Rust-versus-Ruby parsing behavior.

## Expected behavior

- An edge tuple `(source, target, kind, COALESCE(line, -1))` is stored at most once.
- A completed or recovered bulk index has all mandatory edge indexes, especially `idx_edges_unique`.
- Repeated incremental syncs do not grow the edge table when the semantic graph is unchanged.
- An interrupted or failed bulk-load finalization is detected and repaired on the next database open.

## Actual behavior

The affected database contains more than 900,000 duplicate edge rows and no edge indexes. Some individual edge tuples occur hundreds of times. Examples from an aggregate query showed duplicate counts of 252 and 378 for a single `(source, target, kind, line)` tuple.

Once the database reaches this state, sync spends minutes processing a graph whose deduplicated edge set is comparatively small.

## Diagnostic queries

Run these against an affected project database:

```sql
PRAGMA user_version;
PRAGMA index_list(edges);

SELECT kind, COUNT(*) AS count
FROM edges
GROUP BY kind
ORDER BY count DESC;

SELECT
  COUNT(*) AS total_edges,
  COUNT(*) - COUNT(
[...]

PR #359, opened 2026-08-02
fix(db): heal DBs with indexes/triggers lost by interrupted bulk load

### LINKED ISSUE #358: sync pins CPU at ~100% when bulk-load indexes were dropped and never recreated (no dirty sentinel → no self-heal)
## Summary

`tokensave sync` pins a core at 95-100% CPU and never completes when the project DB has lost its edge/node secondary indexes. The indexes are dropped by `begin_bulk_load` during `index_all`; if that load is interrupted **and** the `.tokensave/dirty` sentinel is absent, no code path ever recreates them, and every subsequent `sync` does full-table scans of the `edges` table (46M rows / 4.6 GB in the reported case) on every `delete_nodes_by_file` call.

The #318 fix (unreleased) added `needs_bulk_load_finalization()` + `end_bulk_load()` healing, but that path **only runs when `has_dirty_sentinel()` returns true** (`TokenSave::open`, line 244→312). If the sentinel was never written, was already cleaned up by a prior `open()`, or was removed manually, the healing never triggers and the DB stays unindexed forever.

## Environment

- **tokensave version**: 7.8.1
- **OS**: macOS 15 (Darwin 25.2.0, arm64)
- **Installation**: GitHub release binary
- **Component**: Indexing / sync

## What happened

`tokensave sync` spun at 95.7% CPU across 11 tokio workers for 70+ minutes with zero WAL growth. Two `sample` traces 10s apart showed an identical stack:

```
sync_with_progress_verbose
  → delete_file
    → delete_nodes_by_file
      → DELETE FROM edges WHERE source = ?1 OR target = ?1
        → sqlite3VdbeExec → sqlite3BtreeNext (full btree scan, 4.6 GB)
```

`EXPLAIN QUERY PLAN` confirmed all three queries were `SCAN` (no index):
- `SELECT id FROM nodes WHERE file_path = ?` → `SCAN nodes`
- `DELETE FROM edges WHERE source = ?` → `SCAN edges`
- `DELETE FROM unresolved_refs WHERE from_node_id = ?` → `SCAN unresolved_refs`

## Root cause

`begin_bulk_load` (`src/db/connection.rs:350`) drops **all** secondary indexes before a full reindex:

```sql
DROP INDEX IF EXISTS idx_nodes_kind;
DROP INDEX IF EXISTS idx_nodes_name;
DROP INDEX IF EXISTS idx_nodes_qualified_name;
DROP INDEX IF EXISTS idx_nodes_file_path;
DROP INDEX IF EXISTS idx_nodes_file_path_start_line;
DROP INDEX IF EXISTS idx_edges_source;
DROP INDEX IF EXISTS idx_edges_target;
DROP INDEX IF EXISTS idx_edges_kind;
DROP INDEX IF EXISTS idx_edges_source_kind;
DROP INDEX IF EXISTS idx_edges_target_kind;
DROP INDEX IF EXISTS idx_edges_unique;
DROP INDEX IF EXISTS idx_unresolved_refs_from_node_id;
...
```

`end_bulk_load` (`src/db/connection.rs:394`) recreates them — but only if it runs. If the process is killed between `begin_bulk_load` and `end_bulk_load` (during `index_all`), the indexes are gone.

The #318 fix added detection via `needs_bulk_load_finalization()` (checks for missing `idx_edges_unique`) and healing via `end_bulk_load()`, but this is gated on the dirty sentinel:

```rust
// src/tokensave/mod.rs:244
let crashed = has_dirty_sentinel(project_root);
if crashed {
    // ...
    if db.needs_bulk_load_finalization().await.unwrap_or(false) {
        db.end_bulk_load().await?;  // heals: recreates indexes
    }
}
```

**The gap**: if the sentinel is absent, the healing never runs. The reported DB had no `.tokensave/dirty` file and `PRAGMA user_version = 13`. No migration (v1 through v14) checks for or restores missing indexes. The DB was permanently unindexed.

## Steps to reproduce

1. Start `tokensave init` on a large project (enough to make `index_all` take >a few seconds).
2. Kill the process (Ctrl-C or `kill`) during the indexing phase — after `begin_bulk_load` drops indexes but before `end_bulk_load` recreates them.
3. Remove `.tokensave/dirty` (or have a prior `open()` clear it).
4. Run `tokensave sync` — it will spin at ~100% CPU doing full-table scans on every file deletion.

## Impact

- `sync` becomes a permanent CPU hog (95-100% on one core, never completes).
- The MCP server (`tokensave serve`) still works but serves an unindexed graph — all edge/node queries degrade to full-table scans.
- `PRAGMA quick_check` passes (B-tree is intact), so there's no obvious error message — just silent slowness.
- Users who hit #320 (the `database table is locked` failure that interrupted the bulk load) and then upgraded may be in this state without knowing.

## Suggested fix

A new migration (e.g. `migrate_v15`) that checks for and recreates any missing secondary indexes using `CREATE INDEX IF NOT EXISTS`. This complements the #318 sentinel-based healing by repairing already-damaged DBs that the sentinel path can't reach. It's idempotent and cheap (SQLite skips existing indexes instantly).

The index set should match `create_schema` (the authoritative schema for new DBs), not `migrate_v1`/`migrate_v5` (which created the single-column `idx_edges_source`/`idx_edges_target` that `create_schema` intentionally omits in favor of the composite `idx_edges_source_kind`/`idx_edges_target_kind`).

## Workaround

Manual `CREATE INDEX` on the affected DB:

```sql
CREATE INDEX IF NOT EXISTS idx_nodes_file_path ON nodes(file_path);
CREATE INDEX IF NOT EXISTS idx_edges_source_kind ON edges(source, kind);
CREATE INDEX IF NOT EXISTS idx_edges_target_kind ON edges(target, kind);
CREATE INDEX IF NOT EXISTS idx_edges_kind ON edges(kind);
CREATE UNIQUE INDEX IF NOT EXISTS idx_edges_unique ON edges(source, target, kind, COALESCE(line, -1));
CREATE INDEX IF NOT EXISTS idx_unresolved_refs_from_node_id ON unresolved_refs(from_node_id);
CREATE INDEX IF NOT EXISTS idx_unresolved_refs_reference_name ON unresolved_refs(reference_name);
CREATE INDEX IF NOT EXISTS idx_unresolved_refs_file_path ON unresolved_refs(file_path);
CREATE INDEX IF NOT EXISTS idx_nodes_kind ON nodes(kind);
CREATE INDEX IF NOT EXISTS idx_nodes_name ON nodes(name);
CREATE INDEX IF NOT EXISTS idx_nodes_qualified_name ON nodes(qualified_name);
CREATE INDEX IF NOT EXISTS idx_nodes_file_path_start_line ON nodes(file_path, start_line);
```

undeemed:
## Summary

- Adds schema migration v15 that recreates any missing secondary indexes and FTS/trait-dispatch triggers dropped by an interrupted `begin_bulk_load`, healing already-damaged DBs the #318 sentinel path can't reach
- Fixes `end_bulk_load` to recreate the v14-era FTS triggers (with `search_terms`), not the pre-v14 bodies that silently downgrade FTS search after every `index_all`

## Motivation

Fixes #358.

`begin_bulk_load` drops all edge, node, and unresolved-ref indexes plus the FTS and trait-dispatch sync triggers. `end_bulk_load` recreates them - but only if it runs. The #318 fix added sentinel-based healing in `TokenSave::open`, yet that path **only fires when the `.tokensave/dirty` sentinel exists**. If the sentinel was never written, was already cleaned up by a prior `open()`, or was removed manually, the DB stays permanently unindexed:

- `PRAGMA quick_check` passes (B-tree is intact) - no error message
- `sync` pins a core at ~100% CPU doing full-table scans of the `edges` table (46M rows / 4.6 GB in the reported case) on every `delete_nodes_by_file` call
- The MCP server silently serves an unindexed graph

No migration v1-v14 checks for or restores missing indexes.

A second root cause was found during investigation: `end_bulk_load` recreates pre-v14 FTS triggers (without the `search_terms` column) after every `index_all` on a v14 DB, silently downgrading the FTS index so `tokensave_search` can no longer match camelCase inner words or porter-stemmed inflections on rows inserted/updated after a full reindex.

## Changes

- `src/db/migrations.rs`: bump `LATEST_VERSION` to 15, add `migrate_v15` that recreates all secondary indexes and FTS/trait-dispatch triggers with `CREATE ... IF NOT EXISTS` (instant on healthy DBs, full repair on damaged ones), add dispatch arm
- `src/db/connection.rs`: update `end_bulk_load` FTS trigger bodies to include `search_terms`, matching `create_schema` and `migrate_v14`
- `tests/migration_test.rs`: add `trigger_exists` helper, two new tests (interrupted-bulk-load heal + idempotency on healthy DB)
- `CHANGELOG.md`: two entries under `### Fixed`

## Test plan

- [x] `cargo test --test migration_test` passes (23/23, including 2 new v15 tests)
- [x] `cargo test --lib` passes (656/656)
- [x] `test_migrate_v15_restores_missing_indexes_and_triggers` - drops all indexes + FTS + trait-dispatch triggers (simulating interrupted bulk load), runs v15, asserts all are recreated
- [x] `test_migrate_v15_idempotent_on_healthy_db` - runs v15 on an already-healthy DB, asserts no error and all indexes still present

## Checklist

- [x] Breaking changes: none (migration is additive, `IF NOT EXISTS` is idempotent)


aovestdipaperino:
Thanks for this @undeemed — the self-heal approach is right, and it's the direct fix for #358. I approved the workflow so CI could run, and it surfaced two real failures on the old-schema upgrade path (Linux + macOS, `Test`):

```
test_migrate_v14_backfills_search_terms_and_rebuilds_fts ... FAILED
test_migrate_v7_adds_and_backfills_attrs_start_line ... FAILED

v15: failed to recreate indexes: SQLite failure: `no such table: main.edges`
  operation: "migrate_v15"
```

The v15 step recreates the secondary indexes/triggers unconditionally, but on a full `v6 → v15` (and `v14 → v15`) upgrade the `edges` (and presumably `nodes`/unresolved-ref) tables don't exist at the point v15 runs, so `CREATE INDEX ... ON edges` fails. A couple of options:

- Guard each recreate on the table existing (`SELECT ... FROM sqlite_master WHERE type='table' AND name=?`), skipping the index/trigger when its table isn't present yet, or
- Only run the repair when the DB is already at the schema version where those tables exist (i.e. gate v15 behind "tables present"), so a fresh migration from an older version builds the tables through the normal path first.

Clippy + Format are green; just the two migration tests to sort. Happy to re-run CI once it's pushed. 🙏

aovestdipaperino:
Pushed a fix to the branch (`maintainerCanModify`) so we can land this — the approach is right, it just needed to survive the old-schema upgrade path CI exercises. Two changes:

**1. Gate each object group on its base table existing.** The migration chain also runs `migrate_v15` against older/minimal schemas (the `db_test` v7/v14 fixtures) where `edges`/`unresolved_refs`/`nodes_fts` don't exist yet — hence `no such table: main.edges`. Added a small `table_exists` helper (reusing the `sqlite_master` idiom already in v9/v13) and guarded each group. On a healthy DB (the #358 target) everything is present and behavior is unchanged; on a partial schema it's now a safe no-op.

**2. Recreate exactly the bulk-load drop set, not all of `create_schema`.** `begin_bulk_load` (connection.rs:354–367) does **not** drop `idx_nodes_lower_name` or `idx_nodes_parent_id`, and `end_bulk_load` doesn't recreate them — they're owned by their own migrations. Recreating `idx_nodes_parent_id` here was the source of `no such column: parent_id` on pre-v10 node tables. Removed both so v15 mirrors the bulk-load cycle precisely.

Added `test_migrate_v15_partial_schema_does_not_error`; your `test_migrate_v15_restores_missing_indexes_and_triggers` and `_idempotent_on_healthy_db` still pass, and the previously-red `db_test` v7/v14 chain tests are green. Locally: `migration_test` 24 pass, `db_test` 17 pass, fmt + clippy clean. Re-approving CI now; will merge once it's green — you'll keep authorship on the squash. Thanks again for chasing down the root cause on #358.

undeemed (commented):
Thanks for catching and fixing that. The table guards and exact bulk-load drop set make sense.

Issue #346, opened 2026-07-31
tokensave_circular fabricates cross-language cycles; tokensave_dead_code has 97%+ false-positive rate (Go + TypeScript)


drewjocham:
## Bug: tokensave's call-graph misses real edges across multiple languages/patterns — causes both false-positive dead-code flags and a fabricated circular-dependency report

**tokensave version:** confirmed present in both v7.6.0 and v7.8.1 (checked again 2026-07-31 after upgrade — not fixed by that release)

### 1. `tokensave_circular` reports a single SCC spanning unrelated files across languages

`tokensave_circular` (and the `acyclicity` dimension of `tokensave_health`) reports one giant strongly-connected component spanning 260+ files across three unrelated subsystems in this polyglot monorepo: a Python service (`anomaly/`), and two **separate Go modules** (`desktop/lufis-terminal/`, `web/backend/`). These subsystems cannot import each other — a Python file cannot import a Go package, and the two Go modules have no shared import path between them. The cycle is not architecturally real.

Example members of the single reported cycle (abridged):
```
anomaly/anomaly/alerts/slack.py              (Python)
desktop/lufis-terminal/internal/pty/pty.go   (Go, module A)
web/backend/api/pkg/app.go                   (Go, module B)
```

Re-verified 2026-07-31 against tokensave 7.8.1 — same false cycle still present, unchanged in shape from the 7.6.0 report.

Suspected root cause: a symbol/import-name collision in the graph builder — generic names (`config`, `client`, `context`, `main`) likely resolve to the same graph node across languages/modules instead of being scoped per-package/per-module.

### 2. `tokensave_dead_code` has a very high false-positive rate — misses real edges, doesn't fabricate them

Ran on a Go package tree (`web/backend`, one directory): 33 symbols flagged, 12 non-`init` candidates manually verified against source:
- 21/33 were Go `func init()` — implicitly runtime-dispatched, never have direct call edges by design. (The tool already excludes Rust trait-impl methods for the equivalent reason; `init()` needs the same treatment.)
- 3/33 (`newPlatformStore` in `store_linux.go`/`store_other.go`/`store_windows.go`) — real callers exist, gated by Go build tags (`//go:build linux` etc). The tool doesn't resolve build-tag-conditional compilation.
[...]

Issue #269, opened 2026-07-24
godot-cpp `_bind_methods` bindings aren't modeled: exposed C++ methods show 0 callers/impact, `dead_code` flags all the boilerplate


starfoxcom:
Problem
---
On a Godot GDExtension project the C++ engine exposes methods to GDScript through _bind_methods() (godot-cpp's ClassDB::bind_method / bind_static_method), and GDScript calls them as normal methods. tokensave doesn't create an edge across that binding boundary, so every exposed C++ method looks like it has no callers, and the binding boilerplate looks like dead code.

Environment
---
- tokensave 7.6.0 (stable), Windows 11
- Godot 4 GDExtension project: C++20 engine compiled with godot-cpp, driven from GDScript. ~311 files (82 GDScript, 57 C++, 52 headers).

Steps to reproduce (minimal synthetic)
---
1. tokensave init a directory with a C++ class that binds one method:
```C++
void MyRegistry::_bind_methods() {
    ClassDB::bind_method(D_METHOD("register_item", "cfg"), &MyRegistry::register_item);
}
```
1. and a GDScript file that calls it:
```gdscript
MyRegistry.get_singleton().register_item({ "id": "example" })
```
2. tokensave tool callers --query register_item
3. tokensave tool impact on register_item
4. tokensave tool dead_code

Actual
---
- callers register_item returns []
- impact on register_item returns node_count 1, edge_count 0
- dead_code lists register_item, _bind_methods, and the singleton accessor

Expected
---
register_item resolves back to its GDScript caller, and the _bind_methods binding boilerplate isn't reported as dead.

Notes
---
On a mod/plugin-style engine the exposed C++ API is basically the whole surface that matters, and right now callers, impact, dead_code, and test coverage all under-report it. The dead_code false positives are also predictable: every _bind_methods and singleton accessor. A name-based link (bound method name matched to a GDScript call of the same name) would clear most of it without full type resolution; parsing the bind_method / bind_static_method calls in _bind_methods gives you the export list to match against.

Issue #137, opened 2026-06-17
Dead-code analysis reports trait implementations as false positives


aovestdipaperino:
## Summary

`tokensave_dead_code` (and the redundancy dimension of `tokensave_health`) report **trait implementation methods** as dead code, even though they are reachable via implicit/trait-dispatch invocation. These are false positives that inflate the dead-code count and depress the health score.

## Reproduction

Running `tokensave_dead_code` on a typical Rust project flags methods such as:

- `Deref::deref`
- `Drop::drop`
- `From::from`
- `AsRef::as_ref`
- `Display`/`Debug`::`fmt`

These have **no explicit incoming call edges** in the graph because they are invoked implicitly by the compiler (auto-deref, drop glue, `.into()`/`?` conversions, `{}`/`{:?}` formatting, trait-object dispatch). Deleting any of them breaks compilation, so they are clearly *not* dead.

### Concrete example

In a recent run, 23 symbols were flagged as dead. ~12 of them were trait-impl methods (`deref`, `drop`, `from`, `as_ref`, `fmt`) and were false positives. The remaining inherent/free functions were the only genuine candidates.

## Expected behavior

Methods that implement a trait should not be reported as dead based solely on the absence of explicit incoming call edges. Options:

1. Exclude `impl Trait for Type` methods from dead-code analysis by default (analogous to the existing `main`/`test*`/`pub` exclusions), with an opt-in flag to include them.
2. Model implicit invocation edges for common compiler-inserted dispatch (Drop glue, Deref coercion, `From`/`Into` via `?` and `.into()`, `Display`/`Debug` via format macros).
3. At minimum, tag trait-impl methods in the result so callers can filter them out.

## Impact

- Inflates `dead_code_count` and lowers the `redundancy` dimension in `tokensave_health`.
- Risks an agent deleting trait impls when asked to "prune dead code," which breaks the build.

## Environment

- Detected on a single-target build (macOS); `cfg`-gated and feature-gated callers are a related but separate source of false positives worth noting.

aovestdipaperino:
Reopening: the fix is not taking effect when querying the live graph. A default `tokensave_dead_code` call (no `include_trait_impls`, no `include_public`) on the `mex` crate still returns 20 symbols, and the trait-impl false positives this issue was filed about are still present.

## Still-flagged trait impls (the original #137 case)

[...]

PR #361, opened 2026-08-02
fix(dead-code): exempt Go `init` from dead-code false positives

aovestdipaperino:
## Summary

Partial fix for #346. Go package-level `func init()` is invoked implicitly by the runtime at package load and can never be called explicitly, so it never carries an incoming `calls` edge — and `find_dead_code` flagged every one as dead. **21 of the 33 flags** in #346's Go sample were `init`.

This is the same class already handled for `main`, `test*`, Godot `_bind_methods` (#269), and Rust trait-impl methods (#137). The fix just adds `init` to that existing exclusion list.

## Change

One clause in the dead-code query:

```sql
AND NOT (name = 'init' AND file_path LIKE '%.go')
```

Scoped to `.go` so a callable function named `init` in another language is still checked, and a genuinely unreferenced Go function is still reported dead.

## Testing

`tests/go_init_dead_code_test.rs` — indexes a Go module with an implicit `init`, a live function reached from `init`, and a genuinely dead helper. Asserts `init`/`seedValue` are spared **and** `unusedHelper` is still flagged (proving the exemption stays narrow). RED before the change, GREEN after. `go_bug148`, `godot_bind_methods`, and `graph` suites still pass; fmt + clippy clean.

## Not covered here (still open under #346)

These are separate edge-resolution gaps, not entry-point exclusions, and need deeper work (and ideally the reporter's polyglot repo to verify):
- the fabricated cross-language strongly-connected component in `tokensave_circular`
- build-tag-conditional callers (`//go:build …`)
- function-value assignments (`var F = fn`) counted as a use
- same-file / same-closure calls being missed (Go test-file same-package, TS Pinia setup closures)

https://claude.ai/code/session_01G1EVaNnJTQbj6zSnGXE1KB

### DIFF
--- a/src/graph/queries.rs
+++ b/src/graph/queries.rs
@@ -267,6 +267,12 @@ impl<'a> GraphQueryManager<'a> {
              WHERE name != 'main'
              AND name != '_bind_methods'
              AND name NOT LIKE 'test%'
+             -- Go package-level `init` is invoked implicitly by the runtime at
+             -- package load and can never be called explicitly, so it carries
+             -- no incoming `calls` edge — the same reason `main`, `test*`, and
+             -- Rust trait-impl methods are exempt. Scoped to `.go` so a callable
+             -- function named `init` in another language is still checked (#346).
+             AND NOT (name = 'init' AND file_path LIKE '%.go')
              {visibility_filter}
              {kind_filter}
              {trait_impl_filter}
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Thanks — the 21-of-33 `init` flags were exactly the noise. Merged.

Issue #363, opened 2026-08-02
Allow explicit read-only MCP queries against another initialized project


rNoz:
### Problem

Tokensave's MCP server serves one project graph. During integration work, a query for a symbol implemented in a sibling repository therefore returns an empty result even when that sibling already has a local Tokensave index. The fallback is broad filesystem discovery and direct reads, losing semantic search, traversal, and impact information.

The server must not silently search every local repository. That would mix unrelated code, make provenance ambiguous, and expand both latency and privacy scope.

### Proposed behavior

Add an optional, explicit selector to the side-effect-free semantic/source MCP query tools:

```json
{
  "query": "twd",
  "graph_root": "/Users/example/projects/typewhisper"
}
```

`graph_root` must resolve to the exact root of an already-initialized project. An optional `graph_branch` selects an explicitly tracked branch; otherwise the target checkout's current branch and existing fallback rules apply.

Each selected-project response identifies the canonical root and served branch. Node IDs are namespaced by project so an ID from one repository cannot silently resolve in another. Traversal calls remain explicit and stateless: the caller repeats `graph_root` and passes the namespaced ID.

### Safety constraints

- Open the foreign graph with SQLite read-only flags through a constructor that does not run migrations or write-oriented pragmas. The reader participates in normal SQLite WAL/SHM coordination so concurrent writers and checkpoints remain safe; coordination sidecar bytes may change.
- Never initialize, sync, auto-track, reindex, repair, or edit a selected project from a read query.
- Bypass persistent handler caches, and confine selected source reads to canonical paths inside the selected root.
- Do not expose selectors on edit, VCS, dependency, subprocess (including diagnostics), session-memory, persistent-cache, workflow, or server-administration tools.
- Reject unsupported selector/tool combinations even when a client bypasses JSON Schema validation.
- Reject raw, empty, malformed, wrong-root, and wrong-branch node IDs before selected dispatch. Preserve source/body bytes while qualifying structured references.
- Do not count foreign calls in the served project's savings counters, monitor, or global ledger.
- Keep the current project as the unchanged default when selectors are absent.

### Scope

This is explicit single-project routing only. It does not add aliases, project discovery, fuzzy matching, multi-root fan-out, cross-project edits, automatic freshness, or a cache of foreign database connections. It is separate from the multiproject-within-one-database proposal in `docs/MULTIPROJECT.md`.

### Acceptance

[...]

Issue #260, opened 2026-07-22
tokensave_context tool description embeds live node count, poisoning MCP client prompt caches


aovestdipaperino:
## Summary

The `tokensave_context` tool advertises a **live graph node count** inside its MCP tool *description*:

> CALL BUDGET: 10 calls maximum for this project (**520445 nodes**). Stop after 10 calls. If the question is not fully answered, synthesise from what you have — do not exceed the budget.

The `(NNNNNN nodes)` value changes whenever the graph is re-indexed. Because MCP tool descriptions are part of the JSON schema that clients splice into their **system prompt**, any client that fingerprints/caches the system prompt sees the bytes change on nearly every launch and is forced to rebuild the cache.

## Impact

Observed against the `plank` agent, which keeps a fingerprinted KV-cache snapshot of the system prompt (`sysprompt.kv`, ~130 MB, hashed over the prompt text incl. MCP schemas). Diagnosis across two launches:

```
run 1: fingerprint MATCHES stored snapshot; system prompt byte-identical
run 2: fingerprint MISMATCH -> rebuild
       system prompt CHANGED (prev_len=30516, cur_len=30516, first diff at byte 12811)
       …(520445 nodes)…   <- only the digits differ
```

Same length, single mid-prompt diff = the node count ticking. Result: the expensive system-prompt cache is discarded on almost every start. Any Anthropic/OpenAI prompt-cache client will be hit the same way (the schema block sits inside the cached prefix), silently losing cache reads and inflating token cost/latency.

## Root cause

Volatile, session-specific data (current node count) is placed in a **static** surface (the tool schema/description) that clients treat as cacheable and stable.

## Suggested fix

Keep the tool description static. Surface the node count where it belongs — in **tool results** (or a dedicated `tokensave_status` field), not in the schema. The call-budget guidance can stay in the description; just drop the `(NNNNNN nodes)` parenthetical (or replace it with static text).

Same guidance applies to any other tool whose description/schema/`instructions` interpolates live counts, timestamps, versions, or paths — those all break client-side prompt caching.

## Repro

1. Point any MCP client that caches the system prompt at tokensave.
2. Index/launch, note the cached prompt.
3. Trigger a re-index so the node count changes.
4. Relaunch — the client rebuilds its system-prompt cache though nothing meaningful changed.

aovestdipaperino:
[...]

PR #369, opened 2026-08-03
fix(mcp): make graph selector and untracked-branch errors actionable

### LINKED ISSUE #368: MCP graph selector and untracked-branch errors name the problem but not the recovery
## Summary

Follow-up to #363. Three error messages on the graph selector path state what was rejected but not what to do instead, which the feature's own design standard requires ("errors name the rejected selector and give one actionable next step"), and the `graph_root`/`graph_branch` schema descriptions never tell the caller that omitting the selectors is the common case.

## Observed

Real MCP agent session (Copilot CLI, model gpt-5.6-terra) working in a private project, 2026-08-03: 41 tokensave tool calls, 14 of them with selectors, 15 failures, all of them guidance gaps rather than feature defects. The 26 selector-free calls all succeeded.

- 3x `tokensave_context` with `graph_root` set to the served project root: `-32602 invalid graph selector: graph_root selects the same project already served by this MCP server`. The agent retried the identical call in later turns because nothing in the message said to omit the selector.
- 2x `graph_branch` without `graph_root`, the agent trying to reach its working branch of the served project after the rejections above: `-32602 invalid graph selector: graph_branch requires a matching graph_root`. No next step, and no statement that selecting a different branch of the served project is not supported.
- 7x `graph_root` naming a git worktree of the served project: `-32602 ... is not an initialized TokenSave project root; run \`tokensave init\` there`. Initializing a linked worktree is never the right action; this one is kept as a separate follow-up and is not part of the expected fixes below.
- 1x untracked branch via a branch tool: `branch 'upstream/main' is not tracked` with no remedy, while the sibling fallback warnings in the same file already say `Run \`tokensave branch add <branch>\` to track it.`

## Expected

- Same-root rejection says to omit `graph_root` (and `graph_branch`, when present) to query the currently served graph, and states that selecting a different branch of the served project is not supported, so the remedy does not silently discard the caller's branch intent.
- `graph_branch` without `graph_root` carries the same guidance.
- Untracked explicit branch errors name the remedy (`tokensave branch add '<branch>'`, branch name shell-quoted) in both the selected-graph and served-graph open paths.
- The selector schema descriptions say to omit the selectors for the served project and that `graph_root` must name a different project.

Strict same-root rejection stays: a selector makes the call semantically selected (graph-qualified node IDs, read-only open), so silently degrading to a local call would hide routing mistakes.

## Environment

tokensave 7.8.1 (master at the #364 merge), macOS, MCP client Copilot CLI. The messages are platform- and client-independent.

rNoz:
Resolves #368.

## Summary

- Graph selector errors (selecting the served root, `graph_branch` without `graph_root`) and untracked-branch errors now tell the caller exactly how to recover, instead of only naming what was rejected.
- The `graph_root`/`graph_branch` schema descriptions now say to omit the selectors for the served project and that `graph_root` must name a different project.
- Message and description text only: routing, validation order, and the strict same-root rejection are unchanged.

## Motivation

Follow-up to #363. Although we tested e2e with different agents, now using it in field observation of more real agent sessions not focusing on testing tokensave specifically (Copilot CLI, gpt-5.6-terra) showed 15 failed tokensave calls in one session, all guidance gaps: the agent retried the identical same-root call 3 times because the error never said to omit the selector, then tried `graph_branch` alone twice, and hit an untracked-branch error with no remedy. The feature's design standard requires every selector error to give one actionable next step; three messages did not, and the schema descriptions never said the common case is to omit the selectors. The 26 selector-free calls in the same session all succeeded, so this is purely about making the failure paths self-recovering.

### Design decision

Why the error instead of silently treating `graph_root` = served root as omitted:

   1. A selected call is not the same call. With a selector, the graph opens read-only through the selection path and node IDs come back graph-qualified. Silently dropping the
      selector would return local IDs from the local path — a different response shape than the caller asked for. Agents carry node IDs across calls (tokensave_context →
      tokensave_node), so switching the ID format underneath them breaks downstream lookups in ways nothing ever reports.

   2. It hides routing mistakes. The common case is benign, but the case that matters is the agent that thinks it pointed at project B and misresolved the path onto the served
      project A. Silent fallback returns A's data while the agent believes it has B's. Wrong data presented as right data is the worst failure class for an agent, because nothing
      downstream signals it. The error surfaces the mismatch in one round trip.

   3. Silent acceptance teaches nothing. LLM callers reinforce whatever pattern "worked". Silently accepting the selector means the agent sends it forever, paying the selected-path
       cost for a no-op. The explicit error with the remedy corrects the behavior in one retry, and the new schema descriptions prevent the selector from being sent at all.

   4. Consistency. Every other invalid selector errors (uninitialized root, untracked branch, branch without root). One silent case would make the validation surface unpredictable.

This was a deliberate design choice by contributor: keep the strictness, fix the guidance. The field evidence backs that reading — the agent's problem was never the rejection itself, it was retrying verbatim because the message didn't say what to do. Strictness now costs  exactly one self-correcting round trip, once.

## Changes

- `src/mcp/graph_scope.rs`: the same-root rejection gains `; omit graph_root to query it`, or, when `graph_branch` was also supplied, `; omit graph_root and graph_branch to query the currently served graph (selecting a different branch of the served project is not supported)`. The branch-aware variant keeps the remedy truthful instead of silently discarding the caller's branch intent. `graph_branch` without `graph_root` gets the parallel guidance.
- `src/tokensave/mod.rs`, `src/tokensave/util.rs`: untracked explicit branch errors in `open_read_only` and `open_branch` now name the remedy, `tokensave branch add '<branch>'`, with the branch name shell-quoted by a new `shell_quote` helper so a name containing shell metacharacters cannot turn the suggestion into a different command. The selected-graph variant also says to run it in the selected project.
- `src/mcp/tools/definitions.rs`: selector schema descriptions updated; they remain compile-time constants, so prompt-cache stability (#260) is unaffected.

## Test plan

- [x] `cargo test` passes (3114 tests, 113 suites)

New tests, each failing before the change and passing after: unit tests for both same-root variants and the branch-only case (`src/mcp/graph_scope.rs`); integration tests for the untracked-branch remedy in both open paths, including a branch name with a single quote to prove the shell quoting (`tests/tokensave_test.rs`); JSON-RPC cases asserting the new guidance text and the `-32602` class end to end (`tests/mcp_server_test.rs`, run with `--features test-transport`); the canonical schema test pins the new descriptions for all 50 graph-scoped tools.

Real executions, before and after this change:

Before (shipped behavior, excerpts from the observed session):

```text
-32602 invalid graph selector: graph_root selects the same project already served by this MCP server
-32602 invalid graph selector: config error: graph_branch requires a matching graph_root
-32603 tool execution failed: config error: branch 'upstream/main' is not tracked
```

After (live replay of the same call shapes against a build of this branch; the same-root-with-`graph_branch` case used the session's original arguments against the same private project):

```text
-32602 invalid graph selector: graph_root selects the same project already served by this MCP server; omit graph_root to query it
-32602 invalid graph selector: graph_root selects the same project already served by this MCP server; omit graph_root and graph_branch to query the currently served graph (selecting a different branch of the served project is not supported)
-32602 invalid graph selector: config error: graph_branch requires a matching graph_root; omit graph_branch to query the currently served graph (selecting a different branch of the served project is not supported)
-32602 invalid graph selector: branch 'untracked' is not tracked; run `tokensave branch add 'untracked'` in the selected project to track it
```

The fourth line is the untracked-branch remedy shown for the selected-graph open path; the served-graph path (`open_branch`, used by the branch tools) now carries the same remedy, while its `-32603` error class is unchanged and tracked as a separate follow-up.

Self-contained repro (bash, any directory):

```bash
cd /tmp && rm -rf ts-demo && mkdir -p ts-demo/src && cd ts-demo
printf 'fn main() {}\n' > src/main.rs
tokensave init >/dev/null
{ printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"repro","version":"0"}}}'; \
  printf '%s\n' '{"jsonrpc":"2.0","method":"notifications/initialized"}'; \
  printf '%s\n' "{\"jsonrpc\":\"2.0\",\"id\":2,\"method\":\"tools/call\",\"params\":{\"name\":\"tokensave_search\",\"arguments\":{\"query\":\"main\",\"graph_root\":\"$PWD\"}}}"; \
  sleep 2; } | tokensave serve 2>/dev/null | grep -o 'invalid graph selector[^"]*'
```

Expected output: `invalid graph selector: graph_root selects the same project already served by this MCP server; omit graph_root to query it`.

## Checklist

- [x] Breaking changes documented (if any): none, message text only

Let's have another pair of eyes when reviewing this PR and if the direction is correct.


aovestdipaperino:
Thanks — the field observation (15 failed calls, all guidance gaps) made the case clear. Merged.

Issue #308, opened 2026-07-26
Flaky: Windows CI test binaries intermittently crash with STATUS_ACCESS_VIOLATION (0xc0000005)


aovestdipaperino:
Observed once on Windows CI during #307 ([failing job](https://github.com/aovestdipaperino/tokensave/actions/runs/30197552213/job/89781680098)):

```
process didn't exit successfully: resolution_test-b2dd6bea594aa188.exe
(exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

The process died partway through the 20 tests — no assertion failure, no panic message, just an access violation. Re-running the identical commit passed cleanly, and Linux/macOS passed both times, so it is intermittent and Windows-only.

Worth tracking rather than dismissing: an access violation in safe Rust is not a normal test failure. Two candidate explanations, neither verified:

- **Stack overflow in a worker thread.** `0xc0000005` is how a stack overflow commonly surfaces on Windows, and Windows thread stacks are smaller by default. `resolve_all` runs under rayon, so a deep or wide resolution could plausibly hit it.
- **Something genuinely unsound or environmental** in the test binary or a dependency.

Note the resolver was reworked in #305 to borrow rather than clone (#253). That is all safe code and the borrow checker enforces the lifetimes, so it should not be capable of producing this — but the timing is close enough that it should not be ruled out without evidence either.

Next time it appears, the useful data is a `--no-capture` run to see which test was executing when the process died. If it recurs with any regularity, running the Windows job with `RUST_MIN_STACK` raised would test the stack-overflow theory cheaply.

aovestdipaperino:
Two more occurrences today, both on 2026-07-27 — and importantly **not** in `resolution_test`. This time the crash was in the **lib unittest binary**:

```
Running unittests src\lib.rs (target\debug\deps\tokensave-cf6effbb3a7fe661.exe)
running 578 tests
...
error: test failed, to rerun pass `--lib`
Caused by:
  process didn't exit successfully: target\debug\deps\tokensave-cf6effbb3a7fe661.exe
  (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

- PR #312 (`feat/ruby-block-bodies`) — [failed attempt 1](https://github.com/aovestdipaperino/tokensave/actions/runs/30244942756/attempts/1), job `89909819499`
[...]

PR #181, opened 2026-07-05
ci: raise RUST_MIN_STACK to 8 MB on Windows to fix STATUS_ACCESS_VIOLATION

rNoz:
## Summary

- Raise `RUST_MIN_STACK` to 8 MB on the CI test step so the Windows runner matches macOS and Linux behavior.
- Fix `STATUS_ACCESS_VIOLATION (0xc0000005)` mid-run on `tests/context_test.rs` and other async-heavy tests in the Windows job.
- Infra-only change: no source or test modifications.
- Found here https://github.com/aovestdipaperino/tokensave/actions/runs/28743907763/job/85231401020 from the other PR https://github.com/aovestdipaperino/tokensave/pull/180

image

## Motivation

A subset of async-heavy tests fit comfortably on macOS and Linux's 8 MB thread stack but overflow Windows MSVC's 1 MB default, surfacing as `STATUS_ACCESS_VIOLATION` after a few tests report `ok`, before the binary exits. Upstream CI matrix shows a Windows-only failure pattern (Linux and macOS in the same matrix are `success`) replicated across some recent runs. 

Hopefully the fix is one env var on the test step. Because it is infra-only, it can ship independently of any feature work (for example, PR #180 on `rnoz/auto-track-branch`) that also surfaces this fault (where I found the CI issue).

| Run | Branch / SHA | Test Windows | Test macOS | Test Linux |
|---|---|---|---|---|
| 28743907763 | rnoz/auto-track-branch @ `3c9bf72` (PR #180) | failure | success | success |
| 28601631125 | master @ `3f801f6` (2026-07-02) | failure | success | success |
| 27870423856 | master @ `42f6937` (2026-06-20) | failure | success | success |
| 27630308516 | master @ `60a5220` (2026-06-16) | failure | success | success |

Failing job signature: `STATUS_ACCESS_VIOLATION (0xc0000005)` in `tests/context_test.rs` after several `ok` lines print but before the test binary exits. Local macOS repro of the same env setting was inconclusive on its own (the macOS pthread stack stays at 8 MB regardless of `RUST_MIN_STACK`), so empirical confirmation rests on the upstream matrix above.

## Changes

- `.github/workflows/ci.yml`: add `env: { RUST_MIN_STACK: "8388608" }` (8 MiB) on the existing `Run tests` step so the Windows runner reserves an 8 MB thread stack before launching `cargo test --workspace`. 8 MiB matches the macOS and Linux default, normalizing the matrix. Single-step change; no YAML restructure.

[...]

PR #370, opened 2026-08-04
fix(deps): stop libsql closing every connection twice (#367, #308)

### LINKED ISSUE #367: Windows CI: libsql-backed test binaries crash with STATUS_ACCESS_VIOLATION at process exit, after every test passes (follow-up to #308)
## Summary

`STATUS_ACCESS_VIOLATION` is back on Windows CI, with a signature different from the one #308 described and #337 fixed. Every test in the binary passes first, and the crash happens as the process exits. It is not confined to one suite: it has now been seen in two different test binaries, both of which open libsql databases.

Filing separately rather than as a duplicate because the failure mode does not match the one that was fixed. Happy to have this folded into #308 if you read it as the same root cause.

## Observed

First occurrence, `tests/accounting_test.rs`:

```
running 12 tests
test cost_by_agent_groups_correctly ... ok
... all 12 report ok ...
test upsert_droid_turn_accepts_larger_snapshot ... ok

error: test failed, to rerun pass `--test accounting_test`
Caused by:
  process didn't exit successfully: accounting_test-*.exe
  (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

Second occurrence on an unrelated branch, `tests/db_query_test.rs`, same shape:

```
test test_insert_edges_empty ... ok

error: test failed, to rerun pass `--test db_query_test`
Caused by:
  process didn't exit successfully: db_query_test-*.exe
  (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

Both binaries run every test to completion, report `ok` for each, and then die on exit. The common factor is libsql, not the accounting code.

## Why this looks distinct from #308

#308 reported a binary that "died partway through the 20 tests". Here the harness prints `ok` for all 12 tests and the crash lands after the last one. That is teardown, not a test body: something torn down after the harness finishes, rather than a stack exhausted while running. The `/STACK:8388608` change in #337 would not affect it.

## Evidence that it is a flake, not a code regression

Two CI runs of the same branch:

- Run `30798902576` on commit `7011a84`: passed.
- Run `30802591320` on commit `5becba1`: failed as above.

`5becba1` differs from `7011a84` only by deleting three markdown files. `src`, `tests`, `Cargo.toml`, `Cargo.lock`, and `build.rs` are byte-identical between the two commits, verified by comparing tree object IDs. A re-run of the same source then passed all five checks.

So the same bytes pass and fail depending on the run, which is the intermittency #308 was opened for.

## Hypotheses, unverified

I could not reproduce this on macOS or Linux and have no Windows machine, so the following is reasoning from the code rather than a diagnosis. Ranked by how well each explains "all tests pass, then the process dies".

1. **A libsql or SQLite handle torn down at process exit.** `accounting_test.rs` has 12 `#[tokio::test]` cases, each building a separate Tokio runtime and each opening a local libsql database in WAL mode through `GlobalDb::open_at`. `db_query_test.rs` follows the same pattern against the project database. That is many runtime create/destroy cycles and many database opens per process. If any FFI-owned state outlives the runtime that created it, the teardown order at exit is where it would surface, and Windows is far less forgiving than POSIX about unmapping a file view that is still referenced. That both affected binaries are libsql-heavy, while suites that touch no database have not been seen to crash, is the strongest signal available without a debugger.

2. **`GlobalDb` has no explicit close.** `Database` in `src/db/connection.rs` provides `close()`, which drops the connection deliberately. `GlobalDb` exposes `checkpoint()` but no equivalent, and the tests call neither, so every connection is closed implicitly by drop order while a `TempDir` is being removed underneath it. That is benign on POSIX and is exactly the pattern Windows tends to punish.

3. **Test-harness threading.** The crash is reported by the harness process itself rather than by a single test, so a detached or background thread still touching state after `main` returns would fit.

Hypothesis 1 is the one I would test first.

## Suggested diagnostics for someone with a Windows runner

- Run the suite with `--test-threads=1` and see whether the crash survives serialization. If it disappears, that points at concurrent runtime teardown rather than a single handle.
- Run under Application Verifier, or with a debugger attached, to get the faulting module. Knowing whether the fault is inside `libsql-ffi` or the CRT would settle hypothesis 1 immediately.
- Add an explicit close, or an explicit `checkpoint()` followed by dropping the `GlobalDb` before its `TempDir`, in one test and see whether that test's contribution to the crash disappears.

## Impact

Low correctness risk, real process cost. It fails PRs that changed nothing relevant, and it trains contributors to re-run red Windows jobs without reading them, which is how a genuine Windows regression eventually gets waved through.

## Environment

Observed on GitHub-hosted `windows-latest`, `libsql 0.9.30`. Intermittent; the same commit passes on re-run.

oscaruiz:
## Summary

- Fixes #367.
- Also explains #308, which was closed on 2026-07-30 on the stack-exhaustion theory. The crash came back, which is why #367 exists; the suite #308 was filed against (`resolution_test`) reproduces here and is clean after this change. Whether to reopen it or leave it closed as superseded is your call.
- libsql closes every local connection twice. The second `sqlite3_close_v2` reads an already-freed `sqlite3` object, which is a use-after-free on every platform and an intermittent `STATUS_ACCESS_VIOLATION` on Windows.
- Fixed by vendoring the published libsql 0.9.30 crate with a one-line change and routing to it through `[patch.crates-io]`, because the bug is still present upstream.

## Motivation

`libsql::local::Connection::disconnect` is not idempotent:

```
LibsqlConnection::drop  -> self.conn.disconnect()   // local/impls.rs:110  first close
Connection::drop        -> self.disconnect()        // local/connection.rs:40 second close
```

The `Arc::get_mut(drop_ref)` guard only proves unique ownership, which is true on both passes, and `raw` is left dangling after the first close.

This was counted, not inferred. Instrumenting the driver with a print around the FFI call:

| suite | `sqlite3_open_v2` | `sqlite3_close_v2` | `SQLITE_OK` | `SQLITE_MISUSE` |
|---|---|---|---|---|
| `db_query_test` | 76 | **152** | 76 | 76 |
| `discover_test` + `gain_test` (`GlobalDb` path) | 5 | **10** | 5 | 5 |
| with this fix | 81 | **81** | 81 | 0 |

Exactly two closes per handle, the second always failing its state check because the handle is gone.

On glibc and macOS the freed page stays mapped, so the second close returns `SQLITE_MISUSE` harmlessly — that part is measured. That the same read faults on Windows once the block has been recycled or decommitted is the explanation for the platform asymmetry, not a second measurement. It also explains the timing #367 reports: heap corruption from a double free detonates later, at process exit, after every test has already printed `ok`.

Neither #181's `RUST_MIN_STACK` nor #337's `/STACK:8388608` could have addressed this. Both target stack exhaustion; this is a heap lifetime bug. That is also why the crash survived #308 being closed: the fix that closed it was aimed at a different mechanism.

## Changes

- `vendor/libsql/` — a copy of the published libsql 0.9.30 crate. Of its 44 source files, only `src/local/connection.rs` differs from the published crate, by two lines plus the comment explaining why. Upstream's MIT licence is included. Examples, tests, benches, test assets, lockfile and contributor notes are dropped; `[lints]` and `[workspace]` tables are added to its manifest.
- `Cargo.toml` — `[patch.crates-io]` pointing libsql at the vendored copy, with the rationale and the condition for removing it.
- `Cargo.lock` — libsql resolves through the path patch instead of the registry.
- `tests/vendored_libsql_test.rs` — four guards, so the patch cannot be lost silently: the patch entry and the vendored directory must exist together; a source checkout must carry the patch; the vendored version must match the requested one; and the fix must both be present and close the handle *before* nulling it, since the inverse order would pass a naive string check while closing nothing.
- `scripts/win-stress.ps1` — the batch runner used for the measurements below. It counts access violations separately from other non-zero exits, and takes a `-TempDir` so a long batch does not fill the system drive.
- `.gitignore` — ignore the separate Windows target directory that script uses.

## Test plan

- [x] `cargo test --workspace` passes, with one pre-existing exception noted below
- [x] Tested manually (see below)

Run against this branch's head: 67 suites green, clippy clean, `cargo fmt --all -- --check` clean.

The single failure is `mcp_handler_test::run_affected_tests_dispatches_directly_changed_test_files`. It fails identically on `master` without this branch — 190 passed, 1 failed in both — so it is pre-existing and unrelated. It passes when run on its own, so it looks like a timing flake under suite load rather than a real regression.

The meaningful verification is on Windows, since the crash cannot be reproduced on Linux or macOS. 500 iterations per suite, per build:

| suite | unpatched | patched |
|---|---|---|
| `db_query_test` (named in #367) | **29/500 (5.8%)** | **0/500** |
| `accounting_test` (named in #367) | **10/500 (2.0%)** | **0/500** |
| `resolution_test` (named in #308) | **13/500 (2.6%)** | **0/500** |
| `db_test` | — | 0/500 |
| `graph_test` | — | 0/100 |
| `extraction_test` (opens no database, control) | — | 0/500 |

Every failure is exit code `0xC0000005`, and every one lands after the last test has printed `ok`. Had the rate been unchanged, 500 clean runs would carry probability around 10⁻¹³.

Two controls, both on the unpatched build, retire the standing theories:

- `--test-threads=1` does not hide the crash but roughly triples it (**47/300**), so parallel test execution is not what drives it.
- `RUST_MIN_STACK` at 32 MB leaves the rate unchanged (**12/300**), so no amount of stack is relevant.

All 111 test binaries in the workspace were also built for Windows and run twice with this fix and #367's companion drop-order fix applied together: 0 access violations, 0 other failures across 222 runs. That sweep bounds the two changes jointly, not this one alone.

These numbers come from the `x86_64-pc-windows-gnu` toolchain; this machine has no Visual Studio Build Tools. The crash reproduces there anyway, which shows it does not depend on the MSVC CRT or on #337's link flag, but equivalence with the MSVC toolchain CI uses has not been demonstrated directly.

## Note on scope

`cargo package` strips `[patch.crates-io]` and leaves `vendor/libsql` out of the published archive — verified against the generated `.crate`. So this covers CI and anyone building from a checkout, while binaries produced by `cargo install tokensave` keep the upstream double free until libsql itself ships the fix.

The bug is reported upstream as [libsql#2251](https://github.com/tursodatabase/libsql/issues/2251), with the equivalent fix waiting in [libsql#2261](https://github.com/tursodatabase/libsql/pull/2261) since July. Both `0.9.30` and `0.10.0-pre.4` still carry it. The vendored copy and the `[patch]` entry should be deleted together the moment an upstream release lands; `tests/vendored_libsql_test.rs` is written so that removing all three at once is the only clean way to do it.

## Checklist

- [x] Breaking changes documented: none. No change to tokensave's own source, public API, or behaviour on any platform; the only runtime difference is that a connection is closed once instead of twice.
- [x] Third-party licence: upstream libsql's MIT notice is included at `vendor/libsql/LICENSE.md`.


aovestdipaperino:
Thanks for this — the double-close was a nasty use-after-free and the 500-run batching made the fix airtight. Merged.

PR #371, opened 2026-08-04
fix(test): close the database before removing its TempDir on Windows 

oscaruiz:
## Summary

- Found while investigating #367; a separate defect from the access violation, and it does not fix that crash.
- Database test suites removed their `TempDir` while the connection still held `test.db` open. On Windows the removal fails and the directory is left behind — measured at 75 of 76 per run of `db_query_test`.
- Fixed by returning the `TempDir` first from every setup helper, so it is the last binding declared at each call site and therefore the last dropped.

## Motivation

The helpers returned `(Database, TempDir)` — or `(TokenSave, TempDir)`, or `(Arc<McpServer>, TempDir)` — and every test bound the pair as:

```rust
let (db, _dir) = setup_db().await;
```

Rust drops the bindings of a `let` in reverse order of declaration, so `_dir` goes first and `remove_dir_all` runs on the directory while the database is still open.

POSIX unlinks an open file happily, which is why this is invisible on Linux and macOS. Windows refuses, the removal fails, and `TempDir`'s destructor discards the error by design, so nothing is reported.

Repeated runs accumulate. The batches used to measure the crash in the companion PR stranded 48,245 directories totalling 20.25 GB on the system drive before the leak was noticed, and filled it — at which point tests began failing with exit 101, which reads like a crash until you look. A CI runner already tight on disk is exactly where this surfaces as an unrelated "no space left on device"; this repository has already had to add a disk-cleanup step to the Linux job.

## Changes

Mechanical, and the diff contains no logic: 23 helpers now return the `TempDir` first, and 381 call sites bind it first. Applied by script across 14 test files, then compiled and audited.

`accounting_test`, `resolution_test`, `integration_test` and `memory_test` already ordered it correctly and are untouched. `corruption_test` and `migration_test` needed the same treatment for their three-element helpers, where the `TempDir` sat in the middle and last respectively.

Helper doc comments now state why the order is what it is, since a comment describing the old order is how this would quietly get undone.

## Test plan

- [x] `cargo test --workspace` passes, with one pre-existing exception noted below
- [x] Tested manually (see below)

Run against this branch's head: 67 suites green, clippy clean, `cargo fmt --all -- --check` clean.

The single failure is `mcp_handler_test::run_affected_tests_dispatches_directly_changed_test_files`. It fails identically on `master` without this branch — 190 passed, 1 failed in both — so it is pre-existing and unrelated. It passes when run on its own, so it looks like a timing flake under suite load rather than a real regression.

`mcp_server_test` is behind `#![cfg(feature = "test-transport")]`, so it is not compiled by a plain `cargo test`; run separately with that feature it is 66/66 green on Linux.

Verification is on Windows, since the leak does not occur elsewhere. Every suite re-run there after the change left **0** orphaned directories, against 75 of 76 before: `db_query_test`, `db_test`, `graph_test`, `tokensave_test`, `mcp_server_test`, `walk_up_test`, `project_manifest_test`, `migration_test`, `corruption_test`, `go_bug148_test`.

A sweep of all 111 test binaries, run twice, also came back with 0 leaked directories — that sweep had #367's libsql fix applied as well, so it bounds the two changes jointly.

## Unrelated observation

While verifying on Windows, `mcp_server_test` reports 63 passed and 3 failed: `selected_calls_skip_all_accounting_and_preserve_local_schema_charge`, `selected_search_is_stateless_and_preserves_local_default`, and `selected_warnings_use_canonical_selected_root_remedies`. The same suite built from `master` without this change fails identically — same tests, same line — so this PR does not introduce it.

They go unnoticed because CI runs `cargo test --workspace` without `--features test-transport`, so those 66 tests never execute on any platform. They arrived with #364. Not touched here; worth its own issue.

## Checklist

- [x] Breaking changes documented: none. Test-only change; no production source, public API, or runtime behaviour is touched.


aovestdipaperino:
Thanks for the cleanup — a resource lifetime only POSIX forgives is a great catch. Merged.

Issue #355, opened 2026-08-01
Indexer silently skips hidden directories — repos with production code under .github/ get a blind graph


dangeReis:
## What happened

On a Node.js repo whose production code largely lives under `.github/scripts/` (~87 files, the majority of the codebase's logic), `tokensave init`/`sync` indexed 346 files but **zero** files under `.github/`. There is no warning or diagnostic that a hidden directory was skipped.

The result: graph tools quietly return misleading answers — `tokensave_dead_code` reported 8 symbols repo-wide; after re-including the dotdir it reports 49. `tokensave_redundancy` went from 44 candidate functions / 0 pairs to 732 candidates / 13 duplicate pairs (5 AST-identical). A user who doesn't know the walker skips dotdirs has no way to notice the graph is missing the hottest code.

## Repro

1. Repo with source files under `.github/scripts/` (any GitHub-App/Actions-heavy project qualifies).
2. `tokensave init` then `tokensave tool files '{"path":".github"}'` → empty.
3. `config.json` `exclude` does not list `.github`; `git_ignore: true` but the files are tracked, not gitignored.

## Workaround

Adding `".github/**"` to `include` in `.tokensave/config.json` (per #194) + `tokensave sync` fixes it (87 files added, 472ms).

## Suggestion

Either index tracked hidden directories by default when `git_ignore: true` (git already knows they're source), or emit a sync-time notice like `skipped N tracked files in hidden directories (.github: 87) — add to include[] to index`. The silent part is the bug more than the default itself.

- tokensave 7.8.1, macOS (darwin 25.5.0), JavaScript repo

aovestdipaperino:
Thanks for raising this. There's a supported way to index a hidden directory today — it's the `#194` mechanism. Add the directory to `include` in `.tokensave/config.json`, naming **both** the directory and its contents:

```json
{
  "include": [".github", ".github/**"]
}
```

then re-index:

```bash
tokensave sync -f
```

Both entries matter: the walker prunes at the directory level, so the bare `.github` is what lets it descend into the hidden dir, and `.github/**` matches the files inside. This works whether or not `git_ignore` is set (both scan paths honor `include`), and it's covered by a regression test (`symlink_walk_scope_test.rs`). Only the top-level dot-dir needs the opt-in; nested paths under it are picked up by the normal filters.

[...]

PR #382, opened 2026-08-11
feat(indexing): warn when tracked files in hidden directories are skipped (#355)

dangeReis:
Follow-up to #355. You kept skip-hidden-by-default (which this PR does not touch) — this addresses only the other half of that report: the *silence*. Feel free to close if a sync-time notice isn't something you want; no hard feelings.

## Change
After `init`/`sync`, one batched `git ls-files -z` is compared against the walker's own pruning predicate, and if git-tracked, extractor-indexable files were skipped solely by the hidden-directory filter, a single aggregated line is printed:

```
warning: skipped 87 tracked files in hidden directories (.github: 87) — add ".github" and ".github/**" to include[] in .tokensave/config.json, then run `tokensave sync -f` to index them (or add ".github/**" to exclude[] to silence this warning)
```

- **No change to what gets indexed** — the `include[]` opt-in you described stays the only remedy, and the warning teaches exactly it (both entries + `sync -f`).
- **No new dependencies**; non-git projects and git failures are silently ignored.

## Design notes
- The suppression check mirrors the walker's per-ancestor directory predicate rather than glob-matching file paths, so a partial config (`.github/**` without the bare `.github`) keeps warning until the config actually works — the exact trap you flagged in #355.
- Only files with extensions a registered extractor supports count, so `.github/workflows/*.yml` alone never warns.
- `exclude[]`-covered dirs/files are deliberate opt-outs (no warning); index entries missing on disk (sparse checkouts, staged deletions) are ignored.

## Test
`tests/warn_skipped_hidden_dirs_test.rs` — 9 cases: tracked/untracked, partial-include trap, full include, dir- and file-level excludes, unsupported extensions, nested hidden dirs, sparse/missing-on-disk, non-git. Verified locally against master (1f11c05) on Rust 1.95.0: 9/9 pass, fmt/clippy clean.

### DIFF
--- a/docs/USER-GUIDE.md
+++ b/docs/USER-GUIDE.md
@@ -184,6 +184,21 @@ rules, set `exclude` globs in `.tokensave/config.json`:
 
 `tokensave init --skip-folder <dir>` writes to that same list.
 
+#### Indexing hidden directories
+
+By default, tokensave prunes all hidden directories (dot-prefixed) during the file walk to avoid indexing massive dependency and config folders (like `.venv`, `.cargo`, or `.next`).
+
+If your project contains code in hidden directories (e.g., `.github/scripts/`), you must explicitly opt-in via the `include` array in `.tokensave/config.json`. **Crucially, you must include both the directory and its contents**, because the walker prunes at the directory level before glob matching happens:
+
+```json
+{
+  "include": [
+    ".github",
+    ".github/**"
+  ]
+}
+```
+
 ---
 
 ## Connecting to Your Agent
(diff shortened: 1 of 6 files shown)


aovestdipaperino:
Merging. Thank you @dangeReis — and the part that makes this worth having is the detail in `detect_skipped_hidden_dirs`'s doc comment: walking each *ancestor prefix* the way the walker does, rather than testing the file path.

That's the trap. A user who reads the existing docs, adds `".github/**"` to `include`, and re-runs would still get nothing indexed, because the walker prunes at the directory level before any glob matching happens — and a naive warning that matched on the file path would go silent at exactly that moment, telling them the problem was solved when it wasn't. Getting the check to mirror the pruning rule instead of approximating it is what stops the warning from being actively misleading in its most important case.

Cross-checking against git-tracked files is the right signal too: a hidden directory that git ignores is one the user almost certainly doesn't want indexed, so this only fires where there's real code being silently dropped. Documenting the both-entries requirement (`".github"` *and* `".github/**"`) alongside the warning closes the loop.

PR #383, opened 2026-08-11
fix(tests): isolate fixtures from global git hooks

### LINKED ISSUE #381: The global git hooks installed by `tokensave install` fire inside tokensave's own test fixtures, failing 3 to 7 tests per run
### tokensave version

7.9.0

### Operating system

macOS

### Installation method

Homebrew (brew)

### What happened?

Environment: tokensave 7.9.0, macOS 26.5.2 (Darwin 25.5.0), arm64, installed via Homebrew.

**Expected:** `cargo test` passes on a clean checkout of `master`.

**Actual:** on `3aee482`, `cargo test --no-fail-fast` fails for anyone who accepted the global git hooks during `tokensave install`, and the count moves between runs because the hooks race the tests. Three consecutive full runs gave **7**, **3**, and **4** failures (3,213 / 3,217 / 3,216 passed). Eight distinct tests failed at least once; two of them failed in all three runs, so a run that reproduces nothing at all would be surprising. Plain `cargo test` hides most of it by stopping at the first failing binary: my run stopped in `mcp_handler_test` after 74 test binaries, without ever reaching `tokensave_test`.

`tokensave install` sets a global `core.hooksPath` and writes two hooks into it:

```sh
# ~/.config/git/hooks/post-checkout
elif [ "$3" = "1" ]; then
	/opt/homebrew/bin/tokensave branch add >/dev/null 2>&1 &
fi

# ~/.config/git/hooks/post-commit
/opt/homebrew/bin/tokensave sync >/dev/null 2>&1 &
```

A global `core.hooksPath` applies to every repository, including the `TempDir` ones the fixtures build with `git init`, `git commit`, and `git checkout -b`. Both hooks therefore fire inside the fixtures, in the background, against the fixture's own `.tokensave` directory.

The eight tests. `runs` marks which of the three runs each one failed in, and where two sites are listed that single test broke in different places on different runs:

```
runs
123   open_read_only_omitted_branch_does_not_auto_track              tokensave_test.rs:124  and :296
123   open_read_only_omitted_branch_uses_existing_fallback           tokensave_test.rs:124  and :78
1.3   open_read_only_rejects_untracked_explicit_branch               tokensave_test.rs:124
1.3   test_changelog_with_real_git                                   mcp_handler_test.rs:1504
1..   open_branch_rejects_untracked_branch_with_remedy               tokensave_test.rs:124
1..   open_read_only_rejects_explicit_branch_with_missing_database   tokensave_test.rs:124
1..   test_status_mid_rebuild_reports_rebuild_in_progress            mcp_handler_test.rs:6055
.2.   doc_staleness_flags_code_committed_after_the_doc               mcp_handler_test.rs:6415
```

Two hooks, three consequences.

**1. `post-commit` takes the sync lock.** `tokensave_test.rs:124` is `graph.index_all().await.unwrap()` in `setup_tracked_branch_project`, three lines after that fixture runs `git commit`. `post-commit` backgrounds `tokensave sync`, which takes the fixture's sync lock, and `index_all` loses to it. Seven tests share that fixture, and which of them lose the race varies: five of the seven failed at least once across the three runs, two of those every time, and two never failed at all. `mcp_handler_test.rs:1504` and `:6415` are the same call in other fixtures.

```
called `Result::unwrap()` on an `Err` value: SyncLock { message: "another sync is already in
progress (PID 56656). If this is stale, remove /var/folders/.../T/.tmp3jApiM/.tokensave/sync.lock" }
```

**2. `post-commit` mutates `.tokensave/` mid-walk.** The background sync creates and removes files under `.tokensave/` while the test runs `git add .`, and git aborts on a path that disappeared underneath it:

```
git ["add", "."] failed: fatal: unable to stat '.tokensave/dirty': No such file or directory
```

**3. `post-checkout` writes into the fixture.** When the sync lock is not contended, the `branch add` side surfaces instead. It shows up two ways. The branch DB appears where the test asserts it must not:

```
thread 'open_read_only_omitted_branch_does_not_auto_track' panicked at tests/tokensave_test.rs:296:5:
assertion failed: !fixture.tokensave_dir().join("branches/work.db").exists()
```

and `branch_meta.json` gains a `work` entry, which trips `assert_metadata_unchanged` at `tokensave_test.rs:78`. That test writes `auto_track: true` into its fixture config (line 277), passes `TOKENSAVE_AUTO_TRACK=true` to the helper it spawns, and asserts that a read-only open with an omitted branch still does not auto-track. It never gets to find out. The fixture runs `git checkout -b work` (line 274), `$3 = 1` is git's branch-checkout flag, `post-checkout` backgrounds `tokensave branch add`, and that copies the parent DB to `branches/work.db`.

That last step does not depend on the test's `auto_track: true`. `branch add` never consults the setting: `auto_track` appears only in `config.rs` and in `TokenSave::open` (`src/tokensave/mod.rs:208`), nowhere in `commands.rs` where `BranchAction::Add` lives. The hook tracks the branch whenever it fires. The auto-track code is in `TokenSave::open`, not in the `open_read_only` the test exercises, so nothing on the path under test could have written that file. It came from the other of the two paths #342 names, the `post-checkout` hook, which that issue calls the primary.

**This is why CI does not see it.** The CI job never runs `tokensave install`, so it has no global `core.hooksPath`. `cargo test --workspace` is green on `3aee482`. The failures land only on machines that have the tool set up the way the docs recommend, which is to say on contributors.

I isolated it by bisecting `HOME`, using `open_read_only_omitted_branch_does_not_auto_track` as the probe:

| Condition | Result |
|---|---|
| my real `HOME` | fails, 25 runs out of 25 |
| empty `HOME` | passes, 5 runs out of 5 |
| empty `HOME` + the whole of `~/.tokensave` copied in | passes, 5 runs out of 5 |
| empty `HOME` + `~/.gitconfig` alone | **fails, 5 runs out of 5** |

`~/.gitconfig` contains exactly one setting relevant here, and `tokensave install` is what put it there:

```
core.hookspath=/Users/viko/.config/git/hooks
```

This is not #164 or #177. Those are about *the user's* hooks being silently disabled by the global path, and the chaining preamble fixed them. Here tokensave's own hooks run correctly, exactly as designed, and reach into test fixtures that assumed nothing external would touch them.

### Steps to reproduce

```
$ tokensave install          # accept the global git hooks prompt
$ git config --global core.hooksPath
/Users/you/.config/git/hooks

$ cd tokensave              # v7.9.0, 3aee482
$ cargo test
...
test result: FAILED. 190 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
error: test failed, to rerun pass `--test mcp_handler_test`
```

That stops after 74 test binaries, so it hides most of the damage. Summing the per-binary lines from three full runs, back to back:

```
$ cargo test --no-fail-fast     # run 1
3,213 passed; 7 failed
$ cargo test --no-fail-fast     # run 2
3,217 passed; 3 failed
$ cargo test --no-fail-fast     # run 3
3,216 passed; 4 failed
```

The set is not stable either: only 2 of the 8 failed in all three runs, and both of those broke at a different assertion depending on the run.

The single-test view, which is the clearest one:

```
$ cargo test --test tokensave_test open_read_only_omitted_branch_does_not_auto_track
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 51 filtered out
```

It failed 25 runs out of 25 for me there. It is a race rather than a certainty, though: the hooks are backgrounded, and on one cold first invocation after a rebuild the assertion beat `branch add` to the file and the test passed. If your first run passes, run it again before concluding it does not reproduce.

To confirm the cause without uninstalling anything (5 runs each, unanimous both ways):

```
$ T=$(mktemp -d); HOME="$T" CARGO_HOME=~/.cargo RUSTUP_HOME=~/.rustup \
    cargo test --test tokensave_test open_read_only_omitted_branch_does_not_auto_track
test result: ok. 1 passed; 0 failed

$ T=$(mktemp -d); cp ~/.gitconfig "$T/"; HOME="$T" CARGO_HOME=~/.cargo RUSTUP_HOME=~/.rustup \
    cargo test --test tokensave_test open_read_only_omitted_branch_does_not_auto_track
test result: FAILED. 0 passed; 1 failed
```

### Error output

```shell
# 1. tests lose the sync lock to the backgrounded post-commit hook

thread 'test_changelog_with_real_git' panicked at tests/mcp_handler_test.rs:1504:26:
called `Result::unwrap()` on an `Err` value: SyncLock { message: "another sync is already in progress (PID 56656). If this is stale, remove /var/folders/pl/20kt_3fs21vgzsncgv4k81cm0000gn/T/.tmp3jApiM/.tokensave/sync.lock" }

# 2. git add trips over a file the background sync removed mid-walk

thread 'test_status_mid_rebuild_reports_rebuild_in_progress' panicked at tests/mcp_handler_test.rs:6055:5:
git ["add", "."] failed: fatal: unable to stat '.tokensave/dirty': No such file or directory

# 3. with the sync lock uncontended, post-checkout surfaces instead. the branch DB appears:

thread 'open_read_only_omitted_branch_does_not_auto_track' panicked at tests/tokensave_test.rs:296:5:
assertion failed: !fixture.tokensave_dir().join("branches/work.db").exists()

# 4. and branch_meta.json gains a "work" entry. decoding the two byte arrays:
#    left  branches = ["feature", "master", "work"]
#    right branches = ["feature", "master"]

thread 'open_read_only_omitted_branch_uses_existing_fallback' panicked at tests/tokensave_test.rs:78:9:
assertion `left == right` failed
  left: Some([123, 10, 32, 32, 34, 100, ...])
 right: Some([123, 10, 32, 32, 34, 100, ...])
```

### Component

CLI

### Additional context

`run_git` already isolates the fixture from the developer's git identity by passing author and committer through the environment. Extending the same idea to hooks looks like it would be enough:

```rust
let output = Command::new("git")
    .args(["-c", "core.hooksPath="])
    .args(args)
    ...
```

An empty `core.hooksPath` makes git fall back to the repository's own `.git/hooks`, which in a `TempDir` fixture holds only the inert `*.sample` files `git init` writes, so no hook fires.

One line in `run_git` would not be enough, though. Four test files run `git commit` or `git checkout` across 26 call sites, through six separate helper definitions plus inline `Command::new("git")`: `tokensave_test.rs`, `mcp_handler_test.rs`, `mcp_server_test.rs`, and `graph_test.rs`. Two of those four already fail today, and the count varies run to run. A shared helper looks worth more than six copies of the same line.

There is precedent in the suite for exactly this kind of isolation: `config_test.rs:234` runs git with `GIT_CONFIG_GLOBAL` pointed at an empty file so the developer's global excludes cannot reach the assertion, and `env_var_namespace_test.rs:22` files that variable under "Git subprocess isolation". An empty `GIT_CONFIG_GLOBAL` neutralises `core.hooksPath` as well, so it may be the more consistent of the two fixes. I have no strong preference.

Worth deciding separately: whether `tokensave branch add` and `tokensave sync` should refuse to act on a repository under a temp directory at all, or whether test isolation is the whole answer. I have no view on that, it is your call.

For context on how I got here rather than as a claim about them: #164 and #177 cover the global `core.hooksPath` disabling the user's own hooks, and #179 and #342 cover the auto-track behaviour these hooks implement. None of them mentions the test suite.

vianbas:
## Summary

- Prevent Cargo-driven test fixtures from inheriting developer-level global Git hooks.
- Add a self-validating regression test using a real `post-commit` hook.
- Document the fix under `[Unreleased]`.

## Motivation

Fixes #381.

`tokensave install` can configure a global `core.hooksPath`. Because that setting
applies to every repository, TokenSave's background `post-commit` and
`post-checkout` hooks also ran inside temporary repositories created by the test
suite. They raced the fixtures, intermittently taking their sync locks or modifying
their `.tokensave/` directories.

Three full runs in the reported environment produced 7, 3, and 4 failures.
@rscorer independently reproduced the failure by enabling the installed global
hooks path and confirmed that disabling it made the affected test pass.

## Changes

The existing `[env]` block in `.cargo/config.toml` now supplies one
command-scope Git setting to processes launched by Cargo:

```toml
GIT_CONFIG_COUNT = { value = "1", force = false }
GIT_CONFIG_KEY_0 = { value = "core.hooksPath", force = false }
GIT_CONFIG_VALUE_0 = { value = "", force = false }
```

An empty `core.hooksPath` makes Git use the fixture repository's own
`.git/hooks`, which contains only the inert sample files created by `git init`.

The issue proposed passing the same setting with `-c core.hooksPath=` at each
Git invocation. It also identified 26 call sites across four test files, through
six helper definitions plus inline commands. Supplying it once through Cargo
covers those existing paths and future Git child processes without modifying
each call site. The setting is unchanged; only its delivery moves from individual
arguments to the existing Cargo environment layer.

This environment also reaches binaries started with `cargo run`. A binary invoked
directly from `target/` does not inherit it.

Each entry uses `force = false`, so caller-provided indexed `GIT_CONFIG_*` values
take precedence. A caller-provided configuration can therefore occupy slot `0`
and leave this isolation inactive. If the controlled global hook remains active,
the regression test fails instead of silently passing.

The regression test commits twice:

1. It sets `GIT_CONFIG_COUNT=0`, confirms that the controlled global
   `post-commit` hook runs, and removes its marker.
2. It commits with the Cargo-provided isolation and confirms that the marker is
   not recreated.

Only `post-commit` is exercised. The `core.hooksPath` setting itself is
hook-type agnostic.

This PR does not change whether `tokensave branch add` or `tokensave sync`
should operate inside temporary directories.

## Test plan

- [x] `cargo test --no-fail-fast`: 3,221 passed, 0 failed locally on macOS
      across 124 test binaries plus doc-tests.
- [x] Neutralizing the isolation with `GIT_CONFIG_COUNT=0` fails at the
      negative assertion.
- [x] Removing the hook's executable permission fails at the positive-control
      assertion.
- [x] Fork CI test matrix passed on commit `95fe4bb`:

| Runner | Passed | Failed |
| --- | ---: | ---: |
| `ubuntu-latest` | 3,221 | 0 |
| `macos-14` | 3,221 | 0 |
| `windows-latest` | 3,169 | 0 |

The different Windows total comes from existing platform-gated tests. The new
regression test ran and passed on all three runners. The separate Format and
Clippy jobs also passed.

- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo fmt --check`
- [x] `git diff --check`
- [x] Tested manually through both regression-test mutation cases above.

## Checklist

- [x] `CHANGELOG.md` updated under `[Unreleased]`.
- [x] Breaking changes documented: none.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,6 +6,16 @@
 CFLAGS = { value = "-DNDEBUG", force = false }
 CXXFLAGS = { value = "-DNDEBUG", force = false }
 
+# Test fixtures create real Git repositories and must not inherit developer
+# hooks such as the global hooks installed by `tokensave install` (#381).
+# This command-scope Git setting is inherited by every Git child process while
+# preserving an explicitly supplied GIT_CONFIG_* environment (`force = false`).
+# A caller-provided indexed Git config can occupy slot 0 and disable this
+# isolation; the regression test then exposes the inherited hook.
+GIT_CONFIG_COUNT = { value = "1", force = false }
+GIT_CONFIG_KEY_0 = { value = "core.hooksPath", force = false }
+GIT_CONFIG_VALUE_0 = { value = "", force = false }
+
 # Raise the main-thread stack reserve to 8 MB on Windows MSVC.  RUST_MIN_STACK
 # (set in CI) only resizes threads Rust spawns; the main thread's stack is fixed
 # at link time and defaults to ~1 MB on MSVC, versus ~8 MB on Linux/macOS.  A
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 
 ## [Unreleased]
 
+### Fixed
+- **Cargo-driven tests no longer inherit developer-level Git hooks (#381).** A developer who had run `tokensave install` also had a global `core.hooksPath`, so real Git operations inside temporary test repositories executed TokenSave's background `post-commit` and `post-checkout` hooks. Those hooks raced the fixture itself, intermittently taking its sync lock or modifying `.tokensave/` while assertions were running. Cargo now supplies an empty command-scope `core.hooksPath` to the processes Cargo launches, while leaving the developer's global Git configuration unchanged.
+
 ### Documentation
 - **SECURITY.md described the database as structural metadata only, which understated what it retains (#377).** The policy named `read_cache` as the sole place source text is kept, but `executable_body_fts` is declared without `content=''`, so FTS5 retains the original text of every column in `executable_body_fts_content` — including `body`, which holds complete function bodies. The enumeration above the sentence said "FTS5 search index", which does not tell a reader that full source is retained. Corrected in three places: the storage list, the sentence that follows it, and the Best Practices note, which likewise mentioned only symbol names and `read_cache`. No storage change is involved: the bodies are load-bearing for `tokensave_context`, and a contentless table would return NULL for `SELECT node_id, body`. The "Supported Versions" table, still listing 6.x as current, is updated to 7.x in the same pass.
 
--- a/tests/git_hook_isolation_test.rs
+++ b/tests/git_hook_isolation_test.rs
@@ -1,0 +1,102 @@
+use std::fs;
+use std::path::Path;
+use std::process::Command;
+
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+
+fn run_git(root: &Path, global_config: &Path, args: &[&str]) {
+    let output = Command::new("git")
+        .args(args)
+        .current_dir(root)
+        // The fixture must exercise a controlled global Git configuration,
+        // independently of the developer's real home directory.
+        .env("GIT_CONFIG_GLOBAL", global_config)
+        .env("GIT_AUTHOR_NAME", "TokenSave Test")
+        .env("GIT_AUTHOR_EMAIL", "tokensave@example.com")
+        .env("GIT_COMMITTER_NAME", "TokenSave Test")
+        .env("GIT_COMMITTER_EMAIL", "tokensave@example.com")
+        .output()
+        .expect("run git");
+    assert!(
+        output.status.success(),
+        "git {} failed: {}",
+        args.join(" "),
+        String::from_utf8_lossy(&output.stderr)
+    );
+}
+
+fn commit(
+    project: &Path,
+    global_config: &Path,
+    marker: &Path,
+    message: &str,
+    restore_global_hooks: bool,
+) {
+    let mut command = Command::new("git");
+    command
+        .args(["commit", "--allow-empty", "-m", message])
+        .current_dir(project)
+        .env("GIT_CONFIG_GLOBAL", global_config)
+        .env("GIT_AUTHOR_NAME", "TokenSave Test")
+        .env("GIT_AUTHOR_EMAIL", "tokensave@example.com")
+        .env("GIT_COMMITTER_NAME", "TokenSave Test")
+        .env("GIT_COMMITTER_EMAIL", "tokensave@example.com")
+        .env("TOKENSAVE_TEST_HOOK_MARKER", marker);
+    if restore_global_hooks {
+        // Cancel the command-scope core.hooksPath supplied by Cargo so the
+        // controlled global hooks path applies for the positive control.
+        command.env("GIT_CONFIG_COUNT", "0");
+    }
+    let output = command.output().expect("commit fixture");
+    assert!(
+        output.status.success(),
+        "git commit failed: {}",
+        String::from_utf8_lossy(&output.stderr)
+    );
+}
+
+#[test]
+fn cargo_test_disables_global_git_hooks_in_fixtures() {
+    let sandbox = tempfile::tempdir().expect("tempdir");
+    let project = sandbox.path().join("project");
+    let hooks = sandbox.path().join("hooks");
+    let global_config = sandbox.path().join("gitconfig");
+    let marker = sandbox.path().join("hook-fired");
+    fs::create_dir_all(&project).expect("create project");
+    fs::create_dir_all(&hooks).expect("create hooks");
+
+    let hook = hooks.join("post-commit");
+    fs::write(
+        &hook,
+        "#!/bin/sh\nprintf fired > \"$TOKENSAVE_TEST_HOOK_MARKER\"\n",
+    )
+    .expect("write hook");
+    #[cfg(unix)]
+    fs::set_permissions(&hook, fs::Permissions::from_mode(0o755)).expect("chmod hook");
+
+    let config_output = Command::new("git")
+        .args(["config", "--file"])
+        .arg(&global_config)
+        .arg("core.hooksPath")
+        .arg(&hooks)
+        .output()
+        .expect("write controlled global config");
+    assert!(config_output.status.success());
+
+    run_git(&project, &global_config, &["init", "-b", "master"]);
+
+    commit(&project, &global_config, &marker, "positive control", true);
+    assert!(
+        marker.exists(),
+        "controlled post-commit hook did not run; this test cannot verify Cargo hook isolation"
+    );
+    fs::remove_file(&marker).expect("remove positive-control marker");
+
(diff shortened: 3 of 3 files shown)


aovestdipaperino:
Merging. Thank you @vianbas — the positive control is what makes this test worth having: asserting only that the hook *didn't* fire would pass just as happily if the fixture were broken and no hook could ever run. Driving the controlled hook first, asserting the marker exists, then clearing it and re-running under Cargo's isolation proves the test can actually detect the failure it's guarding against. Nice catch that #381's flakiness was developer-environment leakage rather than a race in our own code.

Issue #323, opened 2026-07-27
Feature Request: Index non-code artifacts (.feature, fixtures, schemas) so tokensave_files can discover them


paulvno:
Environment

  - tokensave: 7.6.2
  - Platform: Windows 11, Git Bash (MSYS2)

  Problem

  tokensave_files and tokensave_context work well for code symbols. Workflows that need to locate non-symbol artifacts — BDD .feature files, step definitions, test
  fixtures, JSON/YAML schemas — have no tokensave alternative. These files have no symbols; they're located by path pattern. The hook blocks the only shell alternative
  (find -name), leaving no way to answer "where are the .feature files for the login flow?"

  This is also the gap that prevented Glob blocking from being accepted in #294. The deferral cited tokensave_files being lossy (only indexed files) as the reason a
  Glob redirect would produce wrong answers. Indexing non-code artifacts would fix that lossiness for these file types and is a prerequisite for Glob blocking to ever
  be safe for non-code discovery.

  Steps to reproduce

  # 1. In a tokensave-indexed project, create a .feature file
  mkdir -p features
  cat > features/login.feature <<'EOF'
  Feature: Login
    Scenario: Valid credentials
      Given the login form is open
      When I submit valid credentials
      Then I am redirected to the dashboard
  EOF

  # 2. Sync the index
  tokensave sync
  In the MCP session:

  tokensave_files(pattern="**/*.feature")
  → [] — 0 results; pattern is valid but entries are absent from the `files` table

  tokensave_context("login flow feature file")
  → 0 results — no symbols to index
  Shell fallback is blocked:

  # find . -name "*.feature"
  # → TOKENSAVE POLICY BLOCK: Bash discovery command blocked
  # (and once evaluate_find_command lands from #294, upstream will block it too)
  Root cause

  tokensave_files already accepts a pattern glob filter (defined in src/mcp/tools/definitions.rs L588-591). However, handle_files queries the files database table
  (src/mcp/tools/handlers/info.rs L104), which only contains entries produced by the symbol-extraction indexing pass. Files with no extractable symbols (.feature, .md,
  .json, etc.) are never inserted, so the pattern matching has nothing to match against.

  Proposed fix

  Option A — extend indexing (preferred)

  Add a "non-symbol artifact" pass to the indexing pipeline that runs after the symbol extractor:

  // Pseudocode — runs after symbol extraction
  for entry in walk_dir(project_root) {
[...]

PR #384, opened 2026-08-12
fix(search): recalibrate the bounded symbol fetch now that #359 activated search_terms

map588:
# Summary

- Set the search_terms FTS weight to 3.0 at all three bm25() sites. The old 6.0 sat above qualified_name and let short camelCase inner-word fields outrank multi-term symbol matches.
- Exclude file and doc nodes from search_nodes_bounded, so they no longer consume slots in the symbol candidate pool of the context builder.

# Motivation

Follow-up to #316, prompted by #358 and #359. Until #359, the stale triggers in end_bulk_load rebuilt nodes_fts without the search_terms column after every index_all. The column was empty on every real index, including the index that the #316 eval was calibrated against. After #359 made the column live, BM25 length normalization gave the one-or-two-token field at weight 6.0 an outsized score. A PascalCase type that matched one query word could outrank a snake_case function that matched three. A bisect pins the rank flip to 0e3e613, and the index confirms it: FTS content for search_terms: goes from 0 rows to 160 rows across that commit. On the 12-query retrieval eval from #313, hit@3 dropped from 66% to 58%.

# Changes

- src/db/queries/search.rs — weight list 10.0, 5.0, 1.0, 2.0, 3.0 at all three sites, and AND n.kind NOT IN ('file', 'doc') in the bounded fetch. The non-obvious part is the second change: rerank-time down-weighting cannot repair a pool that was polluted at fetch time, and #323 now tracks many more non-code rows.
- tests/db_test.rs — regression test: a file node and a doc node that match a fetched term are not returned, and the code symbol is.

# Test plan

- [x] cargo test passes (3221 passed, 0 failed)
- [x] cargo clippy has no new warnings (error set is identical to master under the same toolchain)

Manual: the 12-query eval from #313 on a fixed corpus goes from 58% to 66% hit@3 (MRR .486 to .528), with per-query ranks identical to the ladder that #316 was accepted against, and no query regresses. Inner-word matching keeps working: search_terms:Result still sur QueryResult. Ran tokensave sync --doctor with this binary on a real indexedproject: sync clean, no doctor findings.

# Checklist

- [x] CHANGELOG.md updated (under [Unreleased] if no version bump)
- [x] No secrets, credentials, or .env files included
- [x] Breaking changes documented (if any) — none: no schema or API change, ranking weights only


aovestdipaperino:
Merging once CI clears. Thank you @map588 — and thanks especially for the archaeology on *why* the 6.0 weight never bit until now. A weight described as sitting between `qualified_name` and `name` that is in fact above `qualified_name`'s 5.0 is straightforwardly a bug, but it took #359 repairing the stale triggers from #358 for any real index to exercise the column at all. That's a subtle enough chain that 'the fix caused a regression' would have been the easy and wrong conclusion.

The BM25 length-normalization argument is the part that convinces me: `search_terms` holds only the inner words of a camelCase name, so it's a one-or-two-token field, and short fields at high weight score out of proportion. The concrete failure — a PascalCase type matching one query word outranking a snake_case function matching three, because snake_case names contribute nothing to the column — is a clean statement of the asymmetry.

On the `file`/`doc` exclusion: agreed, and 'rerank-time down-weighting cannot repair a pool that was polluted at fetch time' is the right framing. Now that #323 tracks many more non-code rows, that pool pressure only grows. Both kinds have their own surfacing channels, so nothing becomes unreachable.

Issue #372, opened 2026-08-05
Project-path resolution: worktree mismatch answered silently, plus `.` and case-variant drive letters registered as distinct projects


xaerogonzo:
> Four related asks, all rooted in how a project path is resolved and then
> recorded in the global DB. Sections 1–2 are the original correctness
> problem; 3–4 are global-DB hygiene bugs found while working on it. They
> could be filed separately, but they share a root cause (paths are used
> as-provided rather than canonicalised) so they are grouped here.
>
> Paths below are anonymised; sizes and token counts are verbatim.

## Summary

A Claude Code session started inside a git worktree gets confident,
well-formed `tokensave_*` answers about a **different checkout** — with no
error. tokensave already detects this precisely (every call carries a
`worktree_mismatch` block, and prints a warning naming both `worktree_root`
and `index_root`), but the tools answer anyway.

Reproduced with tokensave 7.8.1 on Windows 11, inside a project's own
`.claude/worktrees/<name>` worktree (Claude Code's default worktree
location — physically nested inside the main repo, `.claude/` gitignored):

```
$ tokensave list
Found 1 tokensave project(s):
  D:\Projects\MyProject     29.1 MB    210.2k tokens
```

That's the **main checkout's** project — `tokensave_search` for a class
written minutes earlier in the worktree returned `[]`; the graph returned the
main checkout's version of files that had been rewritten, at line numbers
that no longer existed in the worktree's tree.

## Why it happens

tokensave resolves its project by searching **upward** from cwd for
`.tokensave/`. A fresh worktree has none, so the search walks past the
worktree boundary and finds a sibling/ancestor project's index — silently,
successfully, with no error.

**Confirmed as the specific trigger**: passing an **explicit** path (absolute
or a literal `.`) does NOT search upward — `tokensave status <path-with-no-index>`
correctly reports `No TokenSave index found at '<path>'. Create one now?`
rather than falling back. The upward search only fires when no path is given
at all, i.e. when the invoking process's cwd is trusted implicitly (exactly
the shape of a bare `tokensave.exe serve` MCP registration with no `-p`).

### 1. `init`'s no-op behavior on an existing index is inconsistent and easy to trust wrongly

Observed **exit 1** running `tokensave init <path>` non-interactively
(explicit path, closed stdin) against an already-initialized project:

```
error: TokenSave is already initialized at '<path>'.
Use tokensave sync to update the index, or tokensave sync --force to rebuild it.
```

[...]

PR #385, opened 2026-08-12
feat(context): retrieval diagnostics footer and test-file collapse

map588:
# Summary

- Add a ### Retrieval footer to tokensave_context output. It reports per-term hit counts, the best post-boost score, a match tier (exact, strong, fts-only), and a warning that lists the query terms that matched nothing.
- Collapse test files in the Related Symbols section to one count line per section, with a pointer to tokensave_callers for detail.

# Motivation

A retrieval miss is silent today. When tokensave_context returns weak entry points, the agent cannot tell a sparse graph from a failed query. It burns follow-up calls on rephrased guesses, or worse, it trusts the wrong entry points. The footer turns that failure into a one-call recovery. Zero-hit terms tell the agent which vocabulary missed the index. The match tier tells it how much to trust the result before it builds on it.

The footer has already paid for itself once. During the investigation behind #384, it showed every term saturated with hits and no zero-hit terms. That observation separated a pool-displacement problem from a vocabulary problem in one call.

The test-file collapse cuts output that costs tokens without informing the agent. A large test file can dominate Related Symbols with dozens of test_* entries. One line keeps the signal that tests exist and drops the roll call.

# Changes

- src/types.rs — RetrievalDiagnostics (term hits, best score, match tier) on TaskContext, with #[serde(default)] so stored or cached contexts without the field still deserialize.
- src/context/builder.rs — find_entry_points returns the diagnostics next to the entry points. The strong tier floor is the new STRONG_MATCH_SCORE = 10.0, calibrated to the negated-BM25 band that #314 introduced.
- src/context/formatter.rs — footer emission and the test-file collapse. The zero-hit list caps at 6 terms with a +N more tail.
- src/context/ranking.rs — is_test_path, a small public wrapper over the existing path_boost classification, so the formatter and ranking agree on what a test file is.
- src/mcp/tools/definitions.rs — the tokensave_context description tells the agent to read the footer and to re-query with keywords synonyms when terms report zero hits.

Non-obvious parts. The tier thresholds are named constants beside EXACT_MATCH_SCORE, not magic numbers, because #314 already moved this scale once and the next recalibration should be one edit. An earlier draft also capped the symbols listed per file, but apply_per_file_cap already bounds per-file contribution at the candidate level with a configurable option. A second hard-coded cap at the display level would duplicate the concept, so it was dropped.

# Test plan

- [x] cargo test passes (3225 passed, 0 failed)
- [x] cargo clippy has no new warnings (error set is identical to master under the same toolchain)

Manual: ran the 12-query eval from #313 before and after. Ranks are identical, which is expected, because the footer observes and does not rank. Verified the footer on live queries against this repository: a healthy query reports match: strong with per-term counts, and a vocabulary miss reports the zero-hit terms and the recovery hint. Formatter tests cover the footer lines, the collapse, and the empty-diagnostics case.

# Checklist

- [x] CHANGELOG.md updated (under [Unreleased] if no version bump)
- [x] No secrets, credentials, or .env files included
- [x] Breaking changes documented (if any) — none. The field is additive and #[serde(default)]. Output gains one section, and existing sections keep their format.


aovestdipaperino:
Merging once CI clears. Thank you @map588 — a silent retrieval miss is exactly the failure class I've been arguing about elsewhere this week (see #372, where tokensave answers confidently from the wrong git worktree): the damage isn't the bad result, it's that the agent has no way to tell a sparse graph from a failed query and so builds on it. A match tier plus zero-hit terms converts that into a one-call recovery, and the vocabulary miss is the case where the agent genuinely cannot guess what went wrong without help.

That the footer already paid for itself during the #384 investigation — showing every term saturated with hits and no zero-hit terms, which separated pool displacement from a vocabulary problem in one call — is a better argument for it than anything in the design rationale.

I did weigh the token cost of adding a section to every `tokensave_context` response, given that the tool's whole purpose is token economy, and considered asking for it behind a flag. Two things decided it: the footer is small and bounded (zero-hit list capped at 6 with a `+N more` tail), and the test-file collapse in the same PR plausibly pays for it outright on any project with substantial test coverage. A diagnostic that's off by default also tends not to be there in the moment you need it.

Good judgement dropping the per-file display cap once you found `apply_per_file_cap` already bounds it at the candidate level — two caps for one concept is how thresholds drift apart. Same for putting the tier floors next to `EXACT_MATCH_SCORE` as named constants rather than magic numbers, given #314 already moved that scale once.

PR #386, opened 2026-08-12
fix(redundancy): validate cached fingerprints

### LINKED ISSUE #380: redundancy/similar: stale node_fingerprints rows produce phantom 1.0-similarity pairs (whole-file body)
## Summary

`redundancy` (and `similar`) can report large numbers of phantom `similarity: 1.0` pairs between
functions that share nothing but a file. Root cause is a stale `node_fingerprints` row: the cached
fingerprint holds the **whole file** as the node's body, so every function in that file collides.

Version: `tokensave 7.9.0`, Windows 11, C++ project (~260 files / ~3.3k nodes).

## Symptom

`tokensave tool redundancy --project . --path <subdir>` returned 20 pairs, all at `1.0`, pairing
obviously-unrelated functions:

```
1.0  ClassA::MethodOne  /  ClassA::MethodTwo    body_tokens [2259, 2259]
1.0  ClassA::MethodOne  /  ClassA::MethodThree  body_tokens [2259, 2259]
...
```

`2259` is the token count of the *entire file*, not of either function. Every signal agreed
(`ast_match: true`, `cfg_match: true`, `call_seq_match: true`, `shingle_jaccard: 1.0`), so the
output is indistinguishable from a genuine duplicate unless you already know the code.

Notably the reported `body_tokens` changed with `--path` scope for the *same* pair (2259 under one
scope, 464 under another), which is what first suggested a cache rather than a live computation.

## Evidence

The `nodes` rows are correct - distinct, accurate `start_line`/`end_line` for every function. The
`node_fingerprints` rows are not:

```sql
select n.name, n.start_line, n.end_line, f.body_tokens, f.ast_hash, f.source_hash
from nodes n join node_fingerprints f on f.node_id = n.id
where n.file_path like '%FileA.cpp' order by n.start_line;
```

```
MethodK          228-253  tok=137   ast=5581ab09  src=dcb31773
MethodL          255-277  tok=123   ast=d39c212f  src=753423f6
MethodM          292-321  tok=129   ast=b9ad2bf4  src=cc98cc02
MethodN          353-367  tok=2259  ast=7209c6ae  src=b9128f57   <-- from here down,
MethodO          376-394  tok=2259  ast=7209c6ae  src=b9128f57       11 nodes share one
MethodP          396-405  tok=2259  ast=7209c6ae  src=b9128f57       identical fingerprint
...
```

Identical `source_hash` across 11 different functions means the fingerprint pass was fed the same
source text (the whole file) for all of them. Three files in the project were affected (11, 5 and 4
nodes); the rest were correct.

## Diagnosis

Fingerprints are **built lazily**, not at index time. Confirmed on a fresh throwaway project:

```
$ tokensave init                     # 4 files, 20 nodes
$ sqlite3 .tokensave/tokensave.db "select count(*) from node_fingerprints;"
0
$ tokensave tool redundancy --project .
$ sqlite3 .tokensave/tokensave.db "select count(*) from node_fingerprints;"
4                                    # written by the first tool run, correct token counts
```

Invalidation on edit works: editing a function body and running `sync` refreshed both `body_tokens`
and `source_hash`, and a planted cross-file duplicate was then found correctly at `0.98`.

So the defect is not the computation, it is **lifetime**. Rows written by an older extractor survive
indefinitely, because plain `sync` only revisits files whose content changed - and these files had
not changed since the bad rows were written. My project had previously gone through the 7.4.0 ->
7.9.0 upgrade and its v13 -> v16 schema migration/re-index, which is the most likely origin.

## Workaround

```
tokensave sync --force     # drops the poisoned rows
tokensave tool redundancy --project .   # rebuilds them correctly
```

Before: 20 pairs, all 1.0, all fabricated. After: 4 pairs with plausible per-node token counts, the
top one a genuine copy-pasted helper duplicated across two files that I already knew about.

## Suggested fixes

1. **Version-stamp the fingerprint rows.** Store the extractor/schema version on each
   `node_fingerprints` row and recompute when it does not match the running binary. A schema
   migration should either invalidate the table or be covered by this check.
2. **Cheap sanity check at write or read time.** A fingerprint whose `body_tokens` equals its file's
   total token count is almost certainly a failed body extraction that fell back to whole-file text.
   Rejecting (or recomputing) those would have turned this into a no-op instead of 20 confident
   false positives.
3. **Surface the cache.** `redundancy`/`similar` output gives no hint that any signal came from a
   cache. A `fingerprints: {cached: N, computed: M}` line in the JSON would make a stale-cache
   result visible without a SQL query.

Happy to test a patch - the detection query above reproduces the bad state reliably on my end.

## Detection query, for anyone hitting this

```sql
select n.file_path, count(*) cnt, group_concat(n.name)
from nodes n join node_fingerprints f on f.node_id = n.id
group by n.file_path, f.ast_hash having cnt > 2;
```

Genuine two-member groups (near-identical two-line helpers) are expected. An 11-member group is not.

vianbas:
## Summary

- Align fingerprint cache validation with the exact byte range of each indexed node.
- Reject cached fingerprints when the stored body-token count differs from the current node body, then recompute them.
- Resolve cache misses by the stored line-and-column range so same-line functions and end-of-file nodes map to the intended tree-sitter node.

## Motivation

Stale `node_fingerprints` rows can make different functions appear as a phantom 1.0-similarity pair. The cache reader previously hashed line-inclusive text, while the fingerprint writer hashes the exact tree-sitter node text, so they were not validating the same bytes. This fixes that mismatch and adds validation for cached token counts.

Closes #380.

## Changes

- Extract and hash source using the node's exact start/end line and column coordinates.
- Add deepest exact-range AST lookup for fingerprint recomputation.
- Recompute rows whose cached `body_tokens` do not equal the live node body's token count.
- Add regression coverage for cache hits, poisoned-row recovery, phantom-pair removal, same-line functions, CRLF, indentation, trailing comments, and files without a trailing newline.
- Document the fix under `[Unreleased]` in `CHANGELOG.md`.

## Test plan

- [x] `cargo test` passes
- [x] Tested manually through the public CLI against an isolated temporary project

Commands run locally:

```text
cargo fmt --check
cargo clippy --workspace --all-targets --all-features -- -D warnings
cargo test redundancy --lib
GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.hooksPath GIT_CONFIG_VALUE_0= cargo test --no-fail-fast
git diff --check
```

Results: 24 redundancy tests passed; the full suite completed with exit code 0.

Manual E2E verification:

- Indexed a temporary Rust project containing two intentionally different functions.
- Called `tokensave tool redundancy` once to populate `node_fingerprints` through the production handler.
- Replaced both cached rows' AST/CFG/call-sequence hashes and shingles with identical values and set `body_tokens` to `0`, while preserving their original source hashes.
- Called the same public CLI command again. Both rows were recomputed: their original per-function token counts returned (`20` and `23`), all planted hash values were replaced, and the reported pair remained at `0.2194` rather than becoming a phantom `1.0` match.
- A final SQL assertion confirmed both rows were present, every poisoned signal was replaced, and both original source hashes were preserved.

## Checklist


aovestdipaperino:
Merging once CI clears. Thank you @vianbas — the root cause here is the kind that's invisible until someone lines the two code paths up side by side: the writer hashing exact tree-sitter node text while the reader hashed line-inclusive text means the cache was validating *different bytes* than it stored, so a row could look fresh while describing something else entirely. Phantom 1.0-similarity pairs are about the worst failure mode `tokensave_redundancy` has, because a false 'these are identical' reads as an actionable finding.

The regression coverage is what sells it — same-line functions, end-of-file nodes, CRLF, indentation, trailing comments, and no-trailing-newline are exactly the cases where a line-based and a range-based extraction diverge, and they're each a real bug someone would otherwise have found one at a time. Adding `body_tokens` validation on top of the source hash is a good belt-and-braces call.

The manual E2E deserves a mention too: deliberately poisoning the cached hashes and shingles while *preserving* the original source hashes is the right way to prove recomputation actually fires, rather than that the row happened to be re-derived for some other reason.

I resolved a CHANGELOG conflict against master on your branch — content unchanged, both entries kept.

PR #388, opened 2026-08-13
fix(extraction): extract symbols inside preprocessor conditionals

cezkid:
## Problem

The C, C++, and Objective-C extractors dispatch `preproc_def` and `preproc_include`, but no conditional node kind. `preproc_if`, `preproc_ifdef`, `preproc_else`, `preproc_elif` and `preproc_elifdef` therefore fall through `visit_node`'s catch-all arm, and every symbol declared inside a conditional block is silently dropped.

The failure is quiet and reads as a valid answer: the affected file still indexes as a file node with zero symbols, so `search` returns `[]` rather than an error.

It costs whole files rather than isolated symbols, because two extremely common idioms wrap an entire file in a conditional:

- `#ifndef FOO_H / #define FOO_H ... #endif` include guards -- most C and C++ headers that do not use `#pragma once`
- build-config guards (`#if !NDEBUG`, `#if !SOME_SHIPPING_FLAG`, ...) wrapping a whole translation unit

Minimal repro:

```cpp
// extracts: OutsideGuard
int OutsideGuard(int a) { return a + 1; }
```

```cpp
// extracts nothing
#if !NDEBUG
int InsideGuard(int a) { return a + 1; }
struct InsideStruct { int x; };
#endif
```

```cpp
// extracts nothing -- the classic include guard
#ifndef WIDGET_H
#define WIDGET_H
class Widget { public: void Draw(); };
#endif
```

## Fix

Unwrap the conditional kinds into `visit_children`, alongside the existing `linkage_specification` / `declaration_list` arm. A conditional block carries no symbol itself; its body does.

Both arms of an `#if/#else` are extracted. Deciding which one compiles needs a real preprocessor with the project's define set, and reporting a symbol that some configuration excludes is strictly better than dropping both -- it matches how tag generators handle the same ambiguity.

## Impact

Measured on one ~200-file C++ tree that uses build-config guards heavily:

| | before | after |
|---|---|---|
| nodes | 3592 | 4179 |
| edges | 2438 | 6717 |

16 files went from zero symbols to fully indexed. The edge count nearly tripling is the notable part -- call edges into and out of guarded code were absent from the graph entirely.

## Tests

Six new tests across the three languages, covering the guarded-block, include-guard, and both-branches cases. Each fails on `main` and passes with the fix.

Full suite: `cargo test` passes. Two pre-existing failures in `global::gather_tests` (`empty_dir_yields_empty_result`, `finds_project_at_cwd`) reproduce identically on an unpatched checkout on this machine and are unrelated to extraction.

### DIFF
--- a/src/extraction/c_extractor.rs
+++ b/src/extraction/c_extractor.rs
@@ -113,6 +113,14 @@ impl CExtractor {
             "enum_specifier" => Self::visit_standalone_enum(state, node),
             "preproc_def" => Self::visit_preproc_def(state, node),
             "preproc_include" => Self::visit_preproc_include(state, node),
+            // A conditional block carries no symbol itself, but its body does.
+            // Without this every declaration inside an `#if` / `#ifdef` is
+            // invisible -- which silently drops whole files: the `#ifndef FOO_H`
+            // include-guard idiom wraps an entire header. Both arms of an
+            // `#if/#else` are extracted: deciding which one compiles needs a
+            // preprocessor, and dropping both is strictly worse.
+            "preproc_if" | "preproc_ifdef" | "preproc_else" | "preproc_elif"
+            | "preproc_elifdef" => Self::visit_children(state, node),
             _ => {
                 // For other node types, skip. Comments are picked up as docstrings.
             }
--- a/src/extraction/cpp_extractor.rs
+++ b/src/extraction/cpp_extractor.rs
@@ -186,6 +186,16 @@ impl CppExtractor {
             // `declaration_list` before dispatch, so this can't double-visit
             // them.
             "linkage_specification" | "declaration_list" => Self::visit_children(state, node),
+            // A conditional block carries no symbol itself, but its body does.
+            // Without this every declaration inside an `#if` / `#ifdef` is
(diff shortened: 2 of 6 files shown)


aovestdipaperino:
Merging once CI clears — I've approved the workflow run, since fork PRs sit at `action_required` here until a maintainer does.

Thank you @cezkid. This one is much bigger than its diff suggests: `#ifndef FOO_H` is the include-guard idiom, so it wraps the *entire* contents of essentially every C and C++ header ever written. Not extracting `preproc_ifdef` children didn't degrade those files, it dropped them whole — and silently, since a header that yields no symbols is indistinguishable from one that legitimately has none. Add build-config guards wrapping whole `.cpp` files and the blast radius covers a large fraction of any real C/C++ project.

Extracting both arms of an `#if/#else` is the right call and I'm glad you said why in the comment rather than leaving it implicit. Picking the live arm needs a real preprocessor with the project's actual macro definitions, which is out of scope for a tree-sitter extractor; reporting a symbol that some build configuration excludes is a far cheaper error than dropping both. Covering the Objective-C extractor in the same pass was thorough — it shares the idiom and would otherwise have been left as the one place the bug survived.

Issue #342, opened 2026-07-30
Design discussion and ergonomics: making tokensave "just work" (auto index management follow-up to #179)


rNoz:
I have a new set of proposals after working for a month on this project, and I would like to know the owner's vision, to fully align :)

### Background and scope

This is a design-decision thread for the "auto index management" / "tokensave just works" direction that I opened in #179 and started addressing in #180 (v7.3.0; partially delivered, first steps). It is not a bug report; it collects the calls that need your input before further code lands. Each question is independent, so partial answers are useful.

Recap of where we are. #180 shipped transparent auto-track of the current branch, off by default, opt in via `config.auto_track` or `TOKENSAVE_AUTO_TRACK`, through two paths: a `post-checkout` git hook (primary) and a best-effort auto-track inside `TokenSave::open` (safety net for entry points that never fire git hooks). Two storage-hardening fixes from this follow-up already merged: atomic branch-meta write (#338, PR #340) and branch DB-file collision avoidance (#339, PR #341).

What #179's reporter actually asked for was broader: on every tokensave call, ensure the repo is initialised, ensure the current branch is tracked, and keep the index fresh (including uncommitted changes) with zero manual setup. It helps to split that vision into three layers, because they raise different questions:

1. Freshness: the served graph reflects the working tree without the operator running `sync` or `branch add`. Note the MCP serve path already does this lazily (`McpServer` runs a startup catch-up sync and, on each `tools/call` past a ~30s cooldown, `maybe_sync_if_stale` + `find_stale_files` re-index mtime-changed files). So the freshness gap is really only the non-MCP entry points.
2. Control surface: one predictable knob governs the above across every entry point (hook, CLI, MCP), togglable at runtime without reinstalling a hook.
3. Continuity: a working branch carries its own context (issue/PR link and session memory) so switching branches restores the right thread.

The questions below map to those layers. Q1 is control-surface, Q2/Q3/Q5 are freshness, Q4 is continuity.

### Q1: How should we migrate already-installed post-checkout hooks?

[...]

PR #392, opened 2026-08-13
fix(hooks): track the branch of a new worktree or clone (#391)

### LINKED ISSUE #391: post-checkout hook does not register the initial branch in new worktrees
### tokensave version

7.9.0

### Operating system

macOS

### Installation method

Homebrew (brew)

### What happened?

The global `post-checkout` hook installed by Tokensave initializes a newly created worktree but does not register that worktree's checked-out branch.

The resulting index contains only the repository's default branch. MCP queries from the worktree then warn that its branch is untracked and serve the default branch as a fallback.

  ### Cause

  The generated hook uses mutually exclusive branches:

  ```sh
  if [ "$1" = "0000000000000000000000000000000000000000" ]; then
      tokensave init >/dev/null 2>&1 &
  elif [ "$3" = "1" ]; then
      tokensave branch add >/dev/null 2>&1 &
  fi
```

git worktree add -b feature /path/to/worktree triggers the all-zero previous-HEAD case. The hook runs tokensave init, but the elif prevents tokensave branch add from registering feature.
We need to also run branch add in the first branch.

However, Simply starting both commands as separate background jobs would introduce a race, because branch add could run before init finishes. They must run sequentially within the same background job.

### Suggested fix

For the all-zero case, initialize and then register the current branch sequentially inside one background job:

```sh
  if [ "$1" = "0000000000000000000000000000000000000000" ]; then
      (
          tokensave init >/dev/null 2>&1 || exit 0
          tokensave branch add >/dev/null 2>&1
      ) &
  elif [ "$3" = "1" ]; then
      tokensave branch add >/dev/null 2>&1 &
  fi
```
  I verified this with a real linked worktree. Its branch metadata contained both main and feature.

### Steps to reproduce

```sh
# Create the smallest possible repository with an initial commit.
root=$(mktemp -d)
git init -b main "$root/repo"
git -C "$root/repo" commit --allow-empty -m "initial commit"

# Create a linked worktree on a new branch; this invokes post-checkout.
git -C "$root/repo" worktree add -b feature "$root/worktree"

# Allow the hook's background initialization to finish.
sleep 5

# Inspect which branches TokenSave registered for the worktree.
tokensave branch list --path "$root/worktree"
```
With TokenSave’s original v7.9.0 hook, the result is:
```
  Default branch: main

    main — ...
```

### Error output

```shell
WARNING: branch 'feature' is not tracked — serving from 'main'. Run `tokensave branch add feature` to track it.
```

### Component

Indexing / sync

### Additional context

_No response_

vianbas:
## Summary

- A new `git worktree add -b feature`, and a `git clone -b feature`, were indexed without their branch being tracked (#391). The `post-checkout` hook ran `tokensave init` on git's all-zeros previous-HEAD sentinel and `tokensave branch add` in an `elif`, so the two could never both run.
- The sentinel arm now runs `init` and then `branch add`, sequentially inside a single background job. **The snippet is @polarene's own suggestion from the issue, token for token**; he is on the commit as co-author.
- `scripts/post-checkout`, the reference hook for manual installs, gets the same behaviour. It had never carried the `branch add` arm added in 7.3.0 at all.

## Motivation

Fixes #391.

git reports the initial checkout of a fresh clone, and of every new `git worktree add`, with the all-zeros sentinel. That checkout is also a branch checkout (git passes flag `$3 == 1`, confirmed for `-b`, `<existing-branch>` and `--detach`) and it can land on a branch that is not the default one. The index then recorded only the default branch, and every MCP query from the new checkout carried `WARNING: branch 'feature' is not tracked — serving from 'main'` (`src/tokensave/mod.rs:388`, inserted into each tool-call result at `src/mcp/server.rs:1793`).

Sequencing matters twice over: `branch add` copies the index that `init` creates, so it must run second, and both must sit in one background job because two independent jobs would race.

## Changes

- **`src/agents/hooks.rs`** — `post_checkout_snippet` gains the subshell for the sentinel arm. The `elif` arm, ordinary branch switches and file checkouts are untouched. The section now closes with a `# tokensave: end auto-init` fence (see *Deliberately out of scope* below). The doc comment is rewritten: it previously described the sentinel case as "fresh clone" only, which is the mental model that produced the bug.
- **`scripts/post-checkout`** — brought in line with the generated hook.
- **`README.md`** — the branch-tracking section now covers the fresh clone/worktree case, and states plainly that a hook written by an earlier version keeps its old body.
- **`CHANGELOG.md`** — entry under `[Unreleased]`.

## Test plan

- [x] `cargo test` passes — 3221 tests across 124 binaries, 0 failed, four clean runs
- [x] `cargo clippy --all-targets --all-features` — no warnings
- [x] `cargo fmt --check` — clean
- [x] Tested manually

The new regression test (`post_checkout_snippet_tracks_the_branch_of_a_fresh_worktree_or_clone`) **executes** the generated snippet under `sh` against a stub binary and asserts the call sequence for all three argument shapes. It is written that way on purpose: the existing test asserted substrings, which is exactly why two correct-looking commands sitting in mutually exclusive arms sailed through it. Reverting `post_checkout_snippet` to the 7.9.0 body makes the new test fail with `left: ["init"], right: ["init", "branch add"]`, while the old test still passes.

Manual verification, each run twice on macOS with git 2.50.1:

| scenario | before | after |
|---|---|---|
| `git worktree add -b feature` | `main` only | `main` + `feature` |
| `git clone -b feature` | `main` only | `main` + `feature` |
| `git clone` (default branch) | `main` | `main`, unchanged |
| ordinary branch switch | `branch add` only | unchanged |
| `git worktree add --detach` | `branch add` never runs; the sentinel arm is `init` only | `branch add` runs, exits 1 on detached HEAD, output discarded; only `main` tracked, same end state |

Two more checks made with the fix in place, rather than as a before/after pair. A worktree created *inside* an already-indexed checkout still gets its own `.tokensave/`, and the main checkout's `branch-meta.json` comes out byte-identical, so nothing leaks into the parent index. And `tokensave init` exits 1 on an already-initialized project, which is why the `|| exit 0` in the sentinel arm is safe: neither `git worktree add` nor `git clone` will accept a non-empty destination, so that arm cannot land on a directory that already holds a `.tokensave/`.

End to end, the patched binary generated its own hook via `tokensave install --git-hook yes` into a throwaway `HOME`, and that generated hook drove real `git worktree add -b` and `git clone -b`. Re-running `install` against a hook that already carries the new fence leaves the file byte-identical, with one of each marker and no duplicated chaining preamble.

**Not covered:** Windows. The new test is `#[cfg(unix)]`, matching the existing convention in that file, so the Windows job will skip it. The snippet is POSIX `sh` and Git for Windows runs hooks under its bundled shell, but I have no Windows machine to confirm it on.

## Deliberately out of scope

**This reaches new installs only.** `offer_git_post_commit_hook` skips the post-checkout hook whenever the marker is already present (`src/agents/hooks.rs:370-372` on both the v7.9.0 tag and master), and the writer never replaces existing content. Verified: with a throwaway `HOME` seeded with a 7.9.0-shape hook, `tokensave install --git-hook yes` created the post-commit hook and announced it, then left post-checkout byte-identical. `install` is also the only route to those files at all: every `write_global_hook` call site lives in `agents/hooks.rs` and is reachable only from the `Install` arm, and `doctor`/`upgrade` contain no hook code.

Migrating existing hooks is the open policy question in **#342 Q1** from @rNoz, which weighs an ownership-aware in-place rewrite against a prompted migration. That call belongs to the maintainer and to that thread, so nothing here decides it. The only down payment is the closing fence, which serves either option: without it a migrator has to pattern-match the two shapes that shipped in 6.4.3 and 7.3.0, both of which end in a bare `fi`.

Worth flagging while that question is open: the same rule has already stranded one feature. The hook shipped in 6.4.3 and the `branch add` arm only in 7.3.0, so anyone who accepted the hooks between those two versions is still running the init-only hook today, with no command that would replace it.

Until it is settled, an existing install can close the gap with `TOKENSAVE_AUTO_TRACK=1`, or `"auto_track": true` in that project's `.tokensave/config.json`, on the next command that *opens* the project. A bare `tokensave branch list` does not open it and will not track. Simplest of all is a plain `tokensave branch add` in the new worktree.

## Checklist

- [x] Breaking changes: none. Ordinary branch switches, file checkouts and default-branch clones behave exactly as before.

@polarene the fix is yours; say the word and I will drop the co-author trailer, or close this so you can send your own version.


aovestdipaperino:
Merging. Thank you @vianbas — this is a good catch and a better test. The bug is the kind that survives review precisely because both arms *look* right in isolation; the `elif` only fails for the one case where git raises the all-zeros sentinel and sets `$3 == 1` at once. Writing the regression by executing the snippet under `sh` with a stub binary, rather than asserting on substrings, is the right call for exactly the reason you gave: every `contains` assertion in the existing test passes over a mutually-exclusive-arms bug without noticing.

Also appreciated: sequencing `init` before `branch add` inside a single background job rather than two, and being explicit in the changelog that this reaches new installs only. On that last point — I've now answered #342 Q1 in favour of an ownership-aware in-place rewrite, so your closing `# tokensave: end auto-init` marker is exactly the fence that work will build on. Good instinct to add it here.

Issue #87, opened 2026-05-25
Windows: same file indexed under both src/file.py and src\file.py path variants, doubling all tool results


xaerogonzo:
**OS:** Windows 11 Pro

**Symptom:** The `files` table contains duplicate rows for the same physical file — once with forward-slash paths (`src/file.py`) and once with backslash paths (`src\file.py`). Confirmed via:

```sql
SELECT path FROM files WHERE path LIKE 'src/%' OR path LIKE 'src\%'
```

**Impact:** `god_class`, `complexity`, `dead_code`, `redundancy`, and `unused_imports` tools all return doubled results. The `health` tool's `redundancy` dimension score is artificially halved (shows ~0.39 instead of ~0.50+). Dead-code counts inflate by ~2×.

**Repro:** Run `tokensave sync` from PowerShell, then run it again from a Git Bash or subprocess context (or vice versa). Check the `files` table — both path separator variants appear for the same physical files.

**Suggested fix:** Normalize all stored paths to forward slashes (or `os.sep`) before writing to the DB, so the same file always maps to a single row regardless of which shell context triggered the sync.

aovestdipaperino:
Fixed by #89. Backslash-form paths (Windows `src\foo.py`) are now normalized at the four sync/staleness entry points (`sync_if_stale`, `sync_if_stale_silent`, `sync_single_files`, `check_file_staleness`) before they reach the DB, so the lookup hits the canonical row stored under `src/foo.py` and no duplicate is written. Regression tests in `tests/tokensave_test.rs` pin the post-fix behaviour. Thanks @xaerogonzo for the clean repro.

Issue #80, opened 2026-05-22
Very High CPU and Memory usage - Tokensave MCP slow down the entire system.


AGiorgetti:
### tokensave version

5.1.2

### Operating system

Windows

### Installation method

GitHub release binary

### What happened?

Sometimes tokensave MCP server use lots of systems resources for very long time causing a whole system slowdown.

CPU is always in the range 40 - 60%

<img width="2275" height="185" alt="Image" src="https://github.com/user-attachments/assets/5805342b-584c-4834-b2de-46f4ee90940d" />

According to "tokensave status" my database is pretty tiny:

│Files             2,943 │ Nodes            70,118 │ Edges            18,938 │
│ DB Size        353.2 MB │ Source          13.8 MB │ Other             2,940 │
│ Java                  3 │                         │                         │

<img width="1377" height="362" alt="Image" src="https://github.com/user-attachments/assets/46037114-068b-44a8-884e-5c145856dd9e" />

The only thing I can do is kill the process.

### Error output

```shell

```

### Component

None

aovestdipaperino:
I've the impression that this is connected to the daemon.
Are you using it?

AGiorgetti:
No, I'm not using the daemon.

Inviato da Outlook per Android<https://aka.ms/AAb9ysg>
________________________________

aovestdipaperino:
**Status update for v6.0.0 (commit b636b18).** Two changes that target the most likely root causes for this report, plus new telemetry so we can diagnose any residual pressure without guesswork.

### What changed

1. **Watcher rewritten — `target/`, `node_modules/`, `.git/` are no longer watched at the OS level.** The previous implementation called \`watch(project_root, RecursiveMode::Recursive)\` and then filtered ignored paths *after* the event fired. On Windows, \`ReadDirectoryChangesW\` has a small per-watch kernel buffer that gets overwhelmed when a large \`node_modules\` or \`target/\` churns — even though we discarded the events, the kernel still allocated, queued, and woke our callback for every one. The new watcher enumerates top-level entries at startup, registers a recursive watch only for surviving subtrees, and watches the root itself non-recursively to catch new top-level directories. The OS never reports events for ignored directories. ([CHANGELOG](https://github.com/aovestdipaperino/tokensave/blob/master/CHANGELOG.md#600---2026-05-25))

[...]

PR #394, opened 2026-08-14
docs(freshness): replace the removed file watcher with the on-demand model (#390)

### LINKED ISSUE #390: docs: watcher_debounce still describes removed watcher in v7.9.0
### Summary

  TokenSave v7.9.0 uses an on-demand staleness check with a 30-second cooldown, but parts of the documentation and configuration schema still describe an embedded file watcher controlled by
  `watcher_debounce`.

  This makes `watcher_debounce = "2s"` appear to provide two-second synchronization when it currently has no effect on runtime freshness.

  ### Evidence

  - The root README correctly documents:
    - an on-demand staleness check on MCP tool calls;
    - a 30-second cooldown;
    - catch-up synchronization when the MCP server connects;
    - removal of the OS-level watcher in v6.1.0.
  - `src/mcp/server.rs` explicitly states that there is no background watcher task.
  - `src/tokensave/staleness.rs` implements the 30-second notification-free freshness check.
  - `docs/USER-GUIDE.md` still describes an embedded watcher and says `watcher_debounce` configures its debounce interval.
  - `src/user_config.rs` still describes `watcher_debounce` as controlling the embedded MCP watcher, although the runtime does not consume it.

  ### Reproduction

  1. Configure `watcher_debounce = "2s"`.
  2. Start TokenSave through an MCP-connected agent.
  3. Invoke a TokenSave tool to establish the latest staleness-check timestamp.
  4. Create a normal supported source file without committing it.
  5. Wait more than two seconds but less than 30 seconds.
  6. Query TokenSave for the new file: it is not present.
  7. Wait until more than 30 seconds have elapsed and query again: the file is synchronized and returned.

  Observed with TokenSave v7.9.0 on macOS through the Codex MCP integration.

  ### Suggested resolution

  - Update `docs/USER-GUIDE.md` and the `UserConfig::watcher_debounce` field comment to describe the current on-demand model.
  - Remove or explicitly deprecate `watcher_debounce` if it is retained only for configuration compatibility.
  - Document that callers needing freshness inside the 30-second cooldown should run `tokensave sync`.

vianbas:
Fixes #390.

## What was wrong

`docs/USER-GUIDE.md` gave the embedded file watcher a table-of-contents entry
and a section of its own, telling readers that the MCP server "watches the
project directory for file changes and automatically runs incremental syncs in
the background" and offering `watcher_debounce` as the knob that tunes it.
`UserConfig::watcher_debounce`'s doc comment described it as the debounce
interval for that watcher.

Nothing reads that field. Outside the documentation, searching `origin/master`
(`1f11c05`) for `watcher_debounce` matches only two files: `src/user_config.rs`
(the declaration, its default, the `ConfigFile` mirror, the merge logic and unit
tests) and `tests/user_config_test.rs`. No call site consumes the value.
`notify` appears zero times in `Cargo.toml` and zero times in `Cargo.lock`.

The reader disappeared with the watcher itself: at `v6.1.0`,
`src/mcp/server.rs:240` passed `config.watcher_debounce` through
`parse_duration` into `ProjectWatcher::new`; at `v6.1.1` the field has no reader
outside `src/user_config.rs`.

### Measured, not just read

I ran the MCP server against a throwaway project (its own `HOME`, so no real
config was touched) and timed how long a newly created file took to reach the
index. Trigger and observation were kept separate: the trigger is an MCP
`tools/call`, the observation is a direct read of the `files` table from a
private copy of the SQLite file, so observing cannot itself cause a sync.

| `watcher_debounce` | Run 1 | Run 2 | Rebuilt from `origin/master` |
|---|---|---|---|
| `"2s"` (the shipped default) | 29.413 s | 29.413 s | 29.410 s |
| `"600s"` | 29.433 s | 29.444 s | 29.438 s |
| `"0s"` | 29.403 s | 29.424 s | — |

Every measurement falls within 41 ms of every other, across a 300× change in
the setting and a zeroed value. What is being measured is the cooldown, not the
debounce: the gate is 30 s since the last successful sync, so timed from the
file change it lands between 29.4 s and 30.4 s depending on how recently the
previous sync ran. A modification and a deletion sit on the same gate: 29.435 s
and 29.414 s for the modification, 30.433 s and 30.397 s for the deletion,
which started from a freshly confirmed sync and so lands a second later. This
is exactly what @polarene reported.

Two controls separate "no watcher" from "slow watcher":

- Server up, file created, **zero tool calls for 75 s**: not indexed at 15,
  30, 45, 60 or 75 s. One tool call, and it appears. (Two runs.)
- Server stopped, file created, server restarted, **zero tool calls**: indexed
  45 ms and 65 ms after connect, sampled every 20 ms. (Two runs.)

## What actually keeps the index fresh

| Mechanism | Where |
|---|---|
| Catch-up sync when the MCP server connects | `McpServer::run_startup_catch_up_sync`, `src/mcp/server.rs:621` |
| Staleness check at the top of every `tools/call`, 30 s cooldown | `McpServer::maybe_sync_if_stale`, `src/mcp/server.rs:675` |
| The tree walk that finds stale files | `find_stale_files`, `src/tokensave/staleness.rs:71`, which calls `scan_files()`, the same `scan_files_diagnostics()` that `sync_with_progress_verbose` and `index_all_with_progress_verbose` use |
| Per-project sync lock shared with the CLI and peer servers | `try_acquire_sync_lock`, `src/tokensave/guard.rs:109` |

The 30 seconds is a literal in `maybe_sync_if_stale`, in both the `last_sync`
gate and the `last_staleness_check_at` gate. There is no configuration path to
it, which is why the guide now says so explicitly.

## What I changed

**`docs/USER-GUIDE.md`.** The watcher section is replaced by "How the MCP
Server Refreshes the Index", describing connect-time catch-up, the per-call
staleness check and its cooldown, and `tokensave sync` as the way to be fresher
than the cooldown allows. The `watcher_debounce = "15s"` sample is gone; the
annotated `config.toml` sample marks the key inert. Six other places that
referred to a running watcher are corrected, including the troubleshooting
bullet, which claimed the check "runs in the background so [it] never block[s]
you". The staleness walk runs inline in the tool call.

While writing that section I first wrote that the watcher "shipped in 6.0.0",
then checked. It did not. `src/project_watcher.rs` was added by `553c94f`,
first tagged `v3.5.0`, where `src/daemon.rs` and `src/main.rs` drove it. What
happened in 6.0.0 is that daemon mode went away (`ffb5959`) and the watcher
moved into `src/mcp/server.rs`, so 6.0.0 is when it became embedded in
`tokensave serve` rather than when it first existed. The guide now says that.

The reason for the removal is kept, but sourced: `CHANGELOG.md:440` is where
"severe CPU and memory pressure on large monorepos" comes from, and the
sentence now points at it instead of asserting it in the guide's own voice. I
did try to measure it, out of caution, by building `v6.1.0` and running it
beside 7.9.0 on a 2,420-file tree with nested `node_modules`. Peak RSS was
consistent across four runs (31–32 MB against 28 MB); peak CPU was not
reproducible between runs, so I am not quoting a number for it. That result
neither supports nor contradicts the CHANGELOG: my tree is far smaller than the
ones it describes, both binaries were debug builds, and #80 was filed on
Windows against 5.1.2.

**`src/user_config.rs`.** The field's doc comment now says what the field is,
that nothing reads it, and what replaced it.

**`docs/INDEX-DESIGN.md`, `src/tokensave/indexing.rs`.** Two architecture notes
and two code comments named the watcher as the party that syncs. They now name
the MCP server's connect-time catch-up and per-call staleness check, which are
`sync_if_stale_silent`'s two real callers (`src/mcp/server.rs:624` and `:699`).
Neither describes any behaviour the original did not.

**A wrong release number, in five places across three documents.** `README.md`,
`docs/COMPARABLE-TOOLS.md` (three places) and `docs/TOKENSAVE-VS-TOKENSAVIOR.md`
date the removal to 6.1.0. The commit that removed the watcher is `f7f7c9b`
("fix: remove notify watcher (#80) and normalize Windows paths (#87) (#89)",
2026-05-26):

```
$ git merge-base --is-ancestor f7f7c9b v6.1.0 ; echo $?
1
$ git tag --contains f7f7c9b | head -1
v6.1.1
```

CHANGELOG.md already files the entry under `## [6.1.1] - 2026-05-26`, so those
five prose mentions were the outliers. They now say 6.1.1.

**`.github/ISSUE_TEMPLATE/bug_report.yml`.** The "Which part of tokensave is
affected?" dropdown still offered `Daemon / file watcher`. Both are gone: the
standalone daemon in 6.0.0, the watcher in 6.1.1, so choosing it labels an
issue with a subsystem that no longer exists. Removed rather than renamed, and
the form keeps its `Other` option. The field has no `default:` and no
`validations: required`, so nothing else about it changes. Happy to drop this
commit if you would rather keep the option.

## What I deliberately did not change

**The field itself.** The issue suggests removing or deprecating
`watcher_debounce`. That is your call, so I only documented it. One thing worth
knowing before you decide, and I tested this rather than reasoned about it:
`UserConfig` does not set `deny_unknown_fields`, so an old `config.toml` still
*loads* with an unrecognised key, but the next config-mutating command
rewrites the file without it. Adding `unknown_legacy_key = "keep-me"` to a
`config.toml` and then running `tokensave disable-upload-counter` dropped the
key both times I tried, while `watcher_debounce` survived. Deleting the field
would therefore strip it from users' files silently.

**`docs/BRANCHING-USER-GUIDE.md`.** It has a "How the watcher handles branches"
section and four more mentions. Renaming the actor would not make it correct,
so I left it alone rather than guess at replacement prose. Briefly, tested
against a running server, twice each: `last_synced_at` is not updated by an MCP
sync or by `tokensave sync` (`BranchMeta::touch_synced` has one caller in the
tree, `src/commands.rs:107`, on the `tokensave branch` path); and a server
started on `main` that then sees `git checkout feature`, a *tracked* branch,
keeps writing into `main`'s database, still reports `active_branch: main`, and
warns about none of it until it is restarted. That contradicts "You don't need
to restart the MCP server after adding a branch". The detached-HEAD paragraph,
by contrast, is accurate: the fallback happens and `tokensave_status` does
prefix `WARNING: detached HEAD — using default branch index`.

**`docs/LSP-INTEGRATION.md` and `docs/MULTI-BRANCH-DESIGN.md`.** Both are
forward-looking proposals that plan to hook into the watcher ("Phase 3: keep
LSP servers alive between syncs via the embedded MCP watcher"; "Watcher
awareness" as implementation step 6). Their premise is stale, but rewriting
them means deciding what those designs hook into now, which is a design
decision rather than a documentation fix.

**`tests/mcp_handler_test.rs`.** The test
`mcp_server_owns_watcher_and_refreshes_token_map_on_change` still carries the
old name, but its body comment already explains that the watcher is gone, and
the name is cited by the #134 CHANGELOG entry. Renaming it is churn.

**Historical records.** `CHANGELOG.md`, `docs/TOKEN-SAVE-WHATSNEW.md`, and the
dated daemon-mode plan and spec under `docs/superpowers/` mention the watcher
too. They record what shipped at the time rather than describing current
behaviour, so I left them; I did not audit their contents.

## Verification

The three commands `.github/workflows/ci.yml` runs, on this branch:

| Command | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --workspace --all-targets` | clean, no warnings |
| `cargo test --workspace --no-fail-fast` | 124 binaries, 3220 passed, 0 failed, 0 ignored |

Three full runs, identical counts every time. The suite is a guard against
accidental damage rather than evidence for the fix: the only Rust that moved
is doc comments.

One local caveat, in case you reproduce it: the suite needs `core.hooksPath`
neutralised first, or the global post-commit hook that `tokensave install`
writes runs `tokensave sync` inside the tests' temporary repositories and some
of the tests that build real git repositories fail. That is #381, and it
reproduces identically on unmodified `master`.

The `docs/USER-GUIDE.md` heading rename changes its anchor. Its table of
contents is updated in the same commit, and a check of all 14 internal anchor
links in the files I touched finds none broken; nothing links to the
`docs/COMPARABLE-TOOLS.md` heading whose text I changed.


aovestdipaperino:
Merging. Thank you @vianbas — and the part I value most here is that you documented `watcher_debounce` as inert rather than deleting it. Removing the field would have silently stripped the key from existing `config.toml` files on their next save, which is a real (if quiet) data loss for anyone who'd tuned it, and you were right to leave the deprecation decision to me rather than fold it into a docs fix.

The version-archaeology is appreciated too: verifying that `f7f7c9b` isn't an ancestor of `v6.1.0` and that `v6.1.1` is the first tag containing it is the kind of check that's easy to skip and easy to get wrong, and it had propagated to five places across three documents. Dropping the dead issue-template option in the same pass was the right scope.

Issue #362, opened 2026-08-02
tokensave_context can miss exact string-literal matches in TypeScript


RELnemtzov:
## Summary

`tokensave_context` can omit both the component rendering a UI message and the large localization catalog containing the exact quoted phrase. Semantic status/loading decoys outrank the exact copy evidence.

This is reproducible using only the synthetic fixture below.

## Environment

- TokenSave CLI and MCP server 7.8.1
- TypeScript / TSX
- Fresh full index
- Synthetic graph: 26 files, 672 nodes, 19 edges

## Reproduction

Create a fresh directory and run this generator:

```bash
mkdir tokensave-context-string-repro
cd tokensave-context-string-repro

node --input-type=module <<'JS'
import { mkdirSync, writeFileSync } from "node:fs";

mkdirSync("src", { recursive: true });

writeFileSync(
  "src/status-panel.tsx",
  `import { translate } from "./translations";

type TextProps = { children: string };

function Text({ children }: TextProps): string {
  return children;
}

export function Banner({ busy }: { busy: boolean }): string {
  return Text({ children: translate(busy ? "dashboard.waiting" : "dashboard.ready") });
}
`
);

const catalogEntries = Array.from(
  { length: 800 },
  (_, index) => `  "catalog.entry.${index}": "Synthetic message number ${index}",`
).join("\n");

writeFileSync(
  "src/translations.ts",
  `export const MESSAGES = {
  "dashboard.waiting": "Waiting for status",
  "dashboard.ready": "Ready",
${catalogEntries}
} as const;

export type MessageKey = keyof typeof MESSAGES;

export function translate(key: MessageKey): string {
  return MESSAGES[key];
}
`
);

for (let fileIndex = 0; fileIndex < 20; fileIndex += 1) {
  const functions = Array.from({ length: 30 }, (_, functionIndex) => {
    const number = fileIndex * 30 + functionIndex;
    return `export function loadDashboardStatus${number}(loading: boolean): string {
  return loading ? "Checking dashboard state ${number}" : "Dashboard ready ${number}";
}`;
  }).join("\n\n");

  writeFileSync(
    `src/decoy-${fileIndex}.ts`,
    `export const MODULE_STATUS_${fileIndex} = "ready";

${functions}
`
  );
}
JS

tokensave init .
```

Run the exact context request:

```bash
tokensave tool context \
[...]

PR #407, opened 2026-08-17
perf(resolution): stop loading the node columns resolution never reads

aovestdipaperino:
Part 1 of #306. **Does not close it** — see "What this does not do" below.

## The finding

#306 asks for the graph-sized allocation in the resolution path to be chunked or streamed. It cannot be, and I want to record why before proposing the alternative: `ReferenceResolver`'s caches are all `&'a Node` / `&'a str` borrowed from the slice, and resolution needs a *global* name index (name, qualified-name, and every `::` suffix) to bind a cross-file reference. Chunking the input would mean a reference could not find a target outside its chunk. That is a resolver redesign, not a call-site change.

What *is* available without touching behaviour: `get_all_nodes()` selects every column, and two of them are unbounded TEXT that resolution never reads.

- `signature` holds a `const`'s entire initializer — #362 found one at roughly 43 KB, an 800-entry localization catalog in a single field.
- `docstring` holds full doc comments.

Measured field usage across `resolution/resolver.rs` and `resolution/variants.rs`: only `id`, `kind`, `name`, `qualified_name`, `file_path`, `start_line`, `visibility` and `parent_id` are ever touched. Neither text column appears.

## The change

`get_all_nodes_for_resolution()` substitutes `NULL` for those two columns, keeping the column *positions* identical so `row_to_node` is shared rather than duplicated. Both sync sites (full and incremental) use it.

Measured on this repository:

| | bytes |
|---|---|
| nodes | 17,594 |
| `signature` omitted | 631,854 |
| `docstring` omitted | 617,833 |
| id + name + qualified_name + file_path retained | 2,404,590 |

So roughly a third of the node string payload, on a repo whose largest signature is only 1.5 KB. On a project with the large const tables from #362 the ratio is considerably better — that is where the unbounded upside is.

## What this does not do

Being explicit, because it bears on the freshness ceilings in #342 Q2:

- **It narrows the peak; it does not bound it.** The pass still holds every node at once, and the remaining cost stays proportional to node count. A one-dirty-file sync still triggers a whole-graph load.
- **It does not make the RSS ≤1.25× ceiling achievable.** I still expect that benchmark to fail, and this change does not alter that expectation — it reduces a constant factor, not the proportionality.
- **Item 2 of #306 is untouched**: the analysis handlers that independently call `get_all_nodes()` (health ×3, analysis ×3, redundancy, graph). Those are a separate change, and SQL aggregates or paged reads are the right shape for them.

I'd suggest #306 stays open, with this filed against item 1 only.

## Test plan

- [x] `tests/resolution_slim_nodes_test.rs`, three cases:
  - the slim and full loads resolve to **identical edges** — the contract, so nobody has to trust a comment about which fields matter, and a future resolver change that starts reading `signature` fails loudly rather than silently seeing `None`;
  - the two columns really are absent, asserted directly, since identical-edges would also hold if the slim load quietly went back to selecting everything;
  - every field resolution *does* read survives intact, so the identical-edges test cannot pass vacuously on a fixture whose references are all unresolvable.
  - The fixture is deliberately cross-file (trait, impl, caller in another module, re-export, a const with a table initializer) so resolution has real work to do.
- [x] `cargo test --no-fail-fast` — exit 0, full suite
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all`


aovestdipaperino:
Folded directly into `master` rather than merged as a PR, at the maintainer's request. Commit preserved with `cherry-pick -x`, so the original SHA is recorded in the commit message.

PR #415, opened 2026-08-18
feat(ruby): extract attr_reader/attr_writer/attr_accessor as method nodes

janstol:
## Summary

- Ruby `attr_reader`/`attr_writer`/`attr_accessor` declarations now emit method nodes, matching the kind, visibility, parent, and `Contains` edge an equivalent `def` would produce.
- Fixes `private attr_reader :foo` landing public (visibility directive now recurses into the nested attribute-directive call, same as it already does for `private def foo; end`).

## Motivation

`attr_*` produced no node at all, which is worse than a missing edge: the method is invisible to search, `module_api`, `callers`, `rename_preview`, and `doc_coverage`, and any call to it is permanently unresolvable. Measured against a local Rails checkout, `attr_*` declarations (1,540) outnumber `define_method`/`alias_method`/`module_function` combined 5x, making it the dominant source of missing method nodes in application code.

## Changes

- `src/extraction/ruby_extractor.rs`:
  - New `visit_attribute_directive` handler, registered in the dispatch arm alongside `visit_visibility_directive`/`visit_mixin_directive`, modeled on `visit_mixin_directive`'s guard sequence (explicit receiver -> skip, `class_depth == 0` -> skip, arguments required). Only `simple_symbol` and static `delimited_symbol` arguments resolve to a name; string/splat/dynamic-identifier arguments emit nothing rather than fabricate a node.
  - Node construction mirrors `visit_method`: `SingletonMethod` inside `class << self`, visibility from `visibility_mode`, position from the `attr_*` call node, a synthesized signature (`def x` / `def x=(value)`), and `singleton_method_ids`/`foreign_singleton_method_ids` registration so a later `private_class_method` can find it.
  - `visit_visibility_directive`'s catch-all argument arm now recurses into a nested `attr_*` call with `visibility_mode` temporarily set, fixing `private attr_reader :foo`.
  - Because `attr_*` reuses `visit_method`'s existing `class_depth`/`singleton_scope` state machine, it gets correct behavior for free at `class << self`, Concern `included do`/`class_methods do`, and `Class.new do` — no new state needed there.
  - One gap deliberately left open (documented inline): an `attr_*` call written directly inside a `def` body is extracted unconditionally, mirroring an identical pre-existing gap in `visit_mixin_directive` for `include`/`extend`/`prepend`. Fixing it needs a shared `self_is_instance` gate across every DSL directive handler, not a special case for `attr_*` — tracked as follow-up.
- `tests/ruby_extraction_test.rs`: one test per truth-table row (reader/writer/accessor, multiple symbols, delimited vs. interpolated symbols, splat/dynamic/string arguments as negatives, explicit receiver, top-level, module body, `class << self` + `private_class_method`, `Contains` edge, docstring, `included do`/`class_methods do`, `Class.new do`, and the method-body gap case).

Full probe notes (Ruby 3.4.7 semantics, tree-sitter-ruby grammar shapes, the site x behavior table, and the mutation-check results) are in the commit body.

## Test plan

- [x] `cargo test` passes (774 lib tests, 165 Ruby integration tests)
- [x] Mutation-checked every new predicate (reverted one at a time, confirmed exactly its own test failed)
- [x] Tested manually: built the binary, ran `tokensave init` on a scratch fixture covering every truth-table row, queried `.tokensave/tokensave.db` directly to confirm expected nodes present/absent
- [x] Tested against a local Rails checkout (read-only corpus, synced into a throwaway copy): +2,403 method-kind nodes against 1,540 `attr_*` declarations, and spot-checked `ActiveSupport::TimeZone#name` resolves with the correct qualified name, kind, and visibility

## Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 ## [Unreleased]
 
 ### Added
+- **Ruby `attr_reader`/`attr_writer`/`attr_accessor` now emit method nodes.** A receiverless `attr_*` call in a class or module body is indexed the same way an equivalent `def` would be — same kind, visibility, parent, and `Contains` edge — including inside `class << self` and Concern `included`/`class_methods` blocks. Previously these accessors were invisible to search, callers, and dead-code analysis.
+
+### Added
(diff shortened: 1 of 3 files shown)


aovestdipaperino:
Thank you @janstol !

Issue #422, opened 2026-08-20
`gini --metric lines --scope file` adds every nested node's span to the file node's own, so an 849-line file reports 2,104 and the top ten reorders


vianbas:
## What is wrong

`tokensave_gini --metric lines --scope file` sums `end_line - start_line + 1` over **every** node in a file, not over the file's symbols. The graph holds a node per struct, per field, per enum variant, per impl block and per method, and they nest, so one physical line is counted once for every node that encloses it. It also holds one node of kind `file` whose span is the whole file, so the file's own length is already one of the terms being summed and every other node is added on top of it.

`src/types.rs` in this repository is 849 lines. The tool reports 2,104.

| node kind | count | lines contributed |
|---|---|---|
| `file` | 1 | 849 |
| `impl` | 7 | 303 |
| `struct` | 22 | 284 |
| `method` | 11 | 268 |
| `field` | 161 | 161 |
| `enum` | 6 | 129 |
| `enum_variant` | 89 | 89 |
| everything else | 10 | 21 |

The `file` row alone matches `wc -l`. That holds for 448 of the 452 files indexed here; four differ by exactly one line (`tests/fixtures/sample.fst`, `tests/fixtures/sample.svelte`, `tests/fixtures/sample.astro`, and the empty `scripts/mcp_probe/tools/__init__.py`). I did not chase why, and none of the four are near the rankings below.

## Where

`src/mcp/tools/handlers/health.rs:227`, the `("lines", "file")` arm, iterates `nodes` with no kind filter.

The same term appears at `src/mcp/tools/handlers/health.rs:84` inside `compute_health_snapshot`, where it is added to per-file complexity and from there feeds the `equality` dimension of `tokensave_health`.

I found this while reading the neighbouring `scope: "symbol"` arms for #417. That is a different code path and does not touch this one: a build of `0651c0a` and a build carrying the #417 fix both report 2,104 for `src/types.rs`.

## Why it is not a harmless constant

Across the 452 files this repository indexes at `0651c0a`, the reported value divided by the file node's own span runs from 1.00x to 3.45x, median 1.96x. It cannot fall below 1.00x, since the file node is one of the nodes being summed. Files with no extracted symbols, such as shell scripts, sit at exactly 1.00x; densely structured files inflate most. A Gini coefficient is scale invariant, so a uniform factor would cancel. This one does not.

| | value | where it comes from |
|---|---|---|
[...]

Issue #418, opened 2026-08-19
get_all_edges() materialises the whole edge table in six places, including the incremental sync and four MCP handlers


vianbas:
#410 named eight MCP handlers that materialised the whole node table, six were
fixed in `f52a68e`, and #411 took the two that had no predicate to push down.
Three of those eight load the entire edges table in the same handler, and a
fourth handler that neither issue lists does the same. That half was never
opened. #411's own analysis says it out loud while discussing the node load: "it
walks `all_edges` and an edge can point anywhere in the graph". The edge walk was
the justification for loading every node; the edge load itself was not examined.

`get_all_edges()` is `SELECT source, target, kind, line FROM edges`, with no
predicate. There are six consumers.

## The call sites

Measured on `0651c0a`, indexing tokensave's own tree, exported with `git archive`
rather than checked out. Every count comes from two separately indexed copies and
the two runs agreed exactly: 18,045 nodes, 25,580 edges, of which 19,282 are
`calls`. Each line number below was mapped to its enclosing function rather than
read off by proximity.

| Consumer | File | Line | What it uses |
|---|---|---|---|
| `handle_hotspots` | `handlers/analysis.rs` | 660 | in-degree and out-degree per node, then the top `limit` |
| `test_reached_node_ids`, behind `handle_complexity` | `handlers/analysis.rs` | 1530 | `calls` edges only |
| `handle_gini`, for `fan_in` and `fan_out` | `handlers/health.rs` | 210 | cross-file edge counts per file, and nothing at all under `--scope symbol` |
| `handle_test_risk` | `handlers/health.rs` | 671 | `calls` edges only, scanned three times |
| `sync_if_stale_silent` | `tokensave/indexing.rs` | 771 | build-variant propagation |
| `sync_with_progress_verbose` | `tokensave/indexing.rs` | 1122 | build-variant propagation |

The two sync sites are the ones I described in #409, so I will not repeat them
beyond noting that they hold the edge vector live alongside the node slice.

## What it costs

At rest the table is 1.64 MB of `Edge` structs plus 2.09 MB of endpoint strings,
so 3.72 MB materialised on this repository.

For a runtime figure, `handle_gini` happens to offer a clean isolation.
It loads the edge table whenever the metric is `fan_in` or `fan_out`, without
looking at the scope, but under `--scope symbol` the arm that runs never touches
`all_edges`. So the two invocations below produce output that is byte-identical
[...]

PR #423, opened 2026-08-20
fix(gini): measure lines, fan_in and fan_out at symbol scope (#417)

### LINKED ISSUE #417: tokensave_gini returns complexity for three of its ten metric and scope pairs, labelled as the metric asked for
`tokensave_gini` declares five metrics and two scopes in its own schema. Three of
the ten combinations return the complexity distribution instead of the metric
that was asked for, and the response echoes back the requested metric, so nothing
in the reply says otherwise.

Everything below is on `0651c0a`, indexing tokensave's own tree, exported with
`git archive` rather than checked out. Nothing here rests on a single run: the
database counts come from two separately indexed copies and agreed exactly, and
every tool invocation was run twice with the same result.

## 1. Three of five metrics are discarded at `--scope symbol`

`handle_gini` matches on `(metric, scope)` in `handlers/health.rs`. The arms are
`("complexity", "file")` at 219, `("lines", "file")` at 227, `("fan_in", "file")`
at 235, `("fan_out", "file")` at 256, `("members", _)` at 276, `(_, "symbol")` at
299, and a bare `_` at 315.

Four of those five metric arms are pinned to `"file"`. So `--scope symbol` falls
past them into `(_, "symbol")`, which computes
`branches + loops + returns + max_nesting` per function. That is complexity,
whatever was requested. `members` survives only because `("members", _)` accepts
any scope and sits above the catch-all.

Both argument values come from the tool's own declared enums, so this is reached
by following the schema exactly, not by misusing it:

| `--scope symbol` | gini | items | top outlier |
|---|---|---|---|
| `--metric complexity` | 0.5768 | 7673 | `src/main.rs:run` = 232.0 |
| `--metric lines` | 0.5768 | 7673 | `src/main.rs:run` = 232.0 |
| `--metric fan_in` | 0.5768 | 7673 | `src/main.rs:run` = 232.0 |
| `--metric fan_out` | 0.5768 | 7673 | `src/main.rs:run` = 232.0 |
| `--metric members` | 0.5417 | 342 | `Node` = 28.0 |

`lines`, `fan_in` and `fan_out` are wrong. `complexity` is right by coincidence
and `members` by arm order.

The bodies for the first four are byte-identical apart from the echoed name:
`diff` with the `metric` line removed reports no difference, and the file sizes
differ by exactly the length of the name. That the value really is complexity is
confirmed a third way: `SELECT branches + loops + returns + max_nesting` for that
node returns 232 in both databases.

The response for the third row says `"metric": "fan_in"`.

## 2. An unrecognised metric silently becomes complexity

The bare `_` at 315 answers anything it does not know with file-level complexity,
again labelled with whatever was sent:

| invocation | gini | items | top outlier |
|---|---|---|---|
| `gini --metric bogus_metric --scope file` | 0.6634 | 452 | `handlers/analysis.rs` = 552.0 |
| `gini --metric complexity --scope file` | 0.6634 | 452 | `handlers/analysis.rs` = 552.0 |

This needs an off-schema value to reach, so it is milder than case 1, but it is
the same root cause: the catch-all arms are reached silently, and the `metric`
field in the response repeats the request rather than describing the computation.

## 3. `--path` drops the cross-scope edges the whole table was loaded for

Separate cause, same tool, and only `fan_in` and `fan_out` at file scope.

The comment above the load says path scoping happens in SQL, and that `all_edges`
still spans the graph "because fan-in/fan-out of a scoped node counts edges from
outside it". That is the stated reason the entire edges table is loaded. It does
not hold. `node_to_file` at 236 is built from `nodes`, which is the *scoped*
result of `get_nodes_filtered`, and the edge loop at 245 requires both endpoints
to be present in that map, so an edge whose source lies outside the scope is
dropped before it can be counted. The `fan_out` loop at 265 has the same shape.

Reading the code is one method, running the tool a second, and raw SQL over the
two databases a third. All three agree, and the tool and the SQL each gave the
same answer twice. Taking `src/agents/traits.rs` as the target:

| | fan_in |
|---|---|
| `gini --metric fan_in` with no path | 527 |
| `gini --metric fan_in --path src/agents` | 416 |
| SQL, cross-file edges into that file from anywhere | 527 |
| SQL, the same restricted to sources under `src/agents` | 416 |

111 incoming edges from outside the scope are dropped without a note. A scoped
`fan_in` is therefore not the fan-in of a scoped node, it is the fan-in counted
within the scope, which is a different question and a smaller number.

## Why this is more than a doc fix

The coefficient is the headline: a caller asks whether a codebase is lopsided on
a chosen dimension. Returning the complexity distribution under the name `fan_in`
answers a different question with equal confidence, and the reply carries nothing
that distinguishes the two. An agent picking `--scope symbol` from the schema has
no way to notice.

The shape of a fix I would expect: reject an unknown metric by name, either
implement `lines`, `fan_in` and `fan_out` at symbol scope or refuse that pair
explicitly, and either build `node_to_file` graph-wide or state that a scoped
fan-in is scope-local. I have written none of that, so this is a report and not a
proposal.

## Related

The whole-edge-table load behind case 3 is worth its own issue and I am raising
it separately rather than folding a performance report into this one. Case 1 is
what gives that one a clean measurement: `--scope symbol` with `--metric fan_in`
loads every edge and then never reads one.

## Reproducing

```
git clone https://github.com/aovestdipaperino/tokensave && cd tokensave
git checkout 0651c0a && cargo build --release
TS="$PWD/target/release/tokensave"
mkdir -p /tmp/corp && git archive 0651c0a | tar -x -C /tmp/corp
cd /tmp/corp && "$TS" init
for m in complexity lines fan_in fan_out members; do "$TS" tool gini --metric $m --scope symbol --limit 3; done
"$TS" tool gini --metric bogus_metric --scope file --limit 2
"$TS" tool gini --metric complexity --scope file --limit 2
"$TS" tool gini --metric fan_in --scope file --limit 200 | grep -A1 'agents/traits'
"$TS" tool gini --metric fan_in --scope file --path src/agents --limit 200 | grep -A1 'agents/traits'
```

## What I did not check

Whether the same catch-all shape appears in other handlers, so this is a report
about `tokensave_gini` and not a claim about the tool surface generally. Nothing
at large-repository scale. And I have not looked for callers that depend on the
current scoped behaviour, so I cannot say whether case 3 is a regression or has
always read this way.

vianbas:
## Summary

- `lines`, `fan_in` and `fan_out` at `scope: "symbol"` returned per-symbol complexity under the name of the metric that was asked for. They now measure what they say.
- Two new match arms ahead of the catch-all, plus a small `is_symbol_kind` helper so the node-kind list lives in one place.
- Two tests that fail against the previous handler; the two gini tests already there could not have caught this.

## Motivation

Closes #417.

`handle_gini` matches on `(metric, scope)`. At `scope: "symbol"` the only metric with an arm of its own is `members`; `lines`, `fan_in` and `fan_out` fall through to `(_, "symbol")`, which computes per-symbol complexity. The response still carries the metric name that was asked for, so three of the ten pairs the tool declares in its `metric` and `scope` enums answer with complexity, and nothing in the output says so.

## Changes

`src/mcp/tools/handlers/health.rs`, two arms ahead of the catch-all, covering the three metrics:

- `("lines", "symbol")` is the node's line span, the same `end_line - start_line + 1` the file-scope `lines` arm uses per node.
- `("fan_in" | "fan_out", "symbol")` counts edges by endpoint.

Every edge kind counts, and an edge whose two endpoints are the same symbol does not, so a recursive call is not a dependant of itself. That is the file-scope rule at symbol granularity, with one difference under `path`, which is deliberate and worth your attention: the file arms count an edge only when **both** endpoints are inside the filtered set, while these arms require only the symbol being measured to be inside it. The comment above the query says the fan-in of a scoped node counts edges from outside it, and that is what these arms do. The effect is large. Under `--path src/mcp`, `handle_tool_call` has 221 incoming edges counted this way and 3 counted the file way, because almost all of its callers live outside `src/mcp`. If you would rather the two scopes agree, say so and I will match the file arms instead.

`complexity` and `members` are untouched, and `(_, "symbol")` stays as the fallback for a metric outside the declared enum. The handler reads `metric` as a free string, so the enum is not enforced at runtime.

`all_edges` is already loaded for `fan_in` and `fan_out` regardless of scope, so this adds no query.

### Measurements

Indexing this repository at `0651c0a`, `scope: "symbol"`, taken twice with identical results:

| metric | before | after | items |
|---|---|---|---|
| `complexity` | 0.5768 | 0.5768 | 7,673 |
| `lines` | 0.5768 | 0.4994 | 7,673 |
| `fan_in` | 0.5768 | 0.6718 | 7,673 |
| `fan_out` | 0.5768 | 0.5681 | 7,673 |
| `members` | 0.5417 | 0.5417 | 342 |

`before` is a build of `0651c0a` with no other change, not an older release. Both binaries were built with the same command and read the same index, so the handler is the only variable. Every figure above also reproduces on a second index built from a fresh export of the same revision.

The values themselves were checked a second way, against the index in SQL rather than through the tool. The top five symbols of each of the three fixed metrics agree exactly, in both order and value, 15 of 15. The fan-in query is below; fan-out is the same with `e.source` in place of `e.target`, and `lines` reads `end_line - start_line + 1` from `nodes`:

```bash
# On a checked-out branch the live index is .tokensave/branches/<branch>.db,
# where each `/` in the branch name becomes `_`. .tokensave/tokensave.db
# is the default-branch index and can be many days stale.
sqlite3 .tokensave/tokensave.db "
SELECT n.file_path||':'||n.name, COUNT(*) FROM edges e JOIN nodes n ON e.target=n.id
WHERE n.kind IN ('function','method','singleton_method') AND e.source<>e.target
GROUP BY n.id ORDER BY 2 DESC LIMIT 5;"
```

## Test plan

- [x] `cargo test` passes: 3,367 tests across 139 binaries, run three times with identical counts
- [x] `cargo clippy --all-targets` has no new warnings, and `cargo fmt` is clean
- [x] Tested manually, described below

`test_gini_symbol_scope_distinguishes_metrics` asserts the three metrics no longer return complexity's map. `test_gini_symbol_scope_values` pins the values on the four functions of the existing fixture for `lines`, `fan_in` and `fan_out`, and pins `complexity` too, since this change must leave it alone. Both fail against the previous handler and pass against this one, checked two independent ways: stashing the source in place, and copying the test file onto a clean checkout of `0651c0a`.

The two gini tests already in `mcp_handler_test.rs` only assert that the `gini` and `interpretation` fields exist, never a value, and both call the tool at the default `file` scope, so neither could have caught this.

## Checklist

- [x] Breaking changes documented (none; `complexity` and `members` return exactly what they did before)

## What I did not check

- Only Rust sources were indexed for the numbers above. The change is language-independent, but the coefficients are not.
- I did not change the file-scope arms, so under a `path` filter the two scopes now answer differently, as described above. Everything else about them is the same rule read at a different granularity.
- I did not measure whether the same catch-all pattern hides in the other handlers that match on a pair.
- The file-scope `lines` arm I took the span formula from sums that span over every node in a file, which I have since reported separately as #422. This PR does not touch that arm; only the per-node formula is shared, and the symbol-scope arm here counts each symbol once.
- `members` ignores `scope` entirely, by its `("members", _)` arm. That is unchanged here and may be deliberate.


aovestdipaperino:
Keep what you have — the `path` behaviour in your new arms is the right one, and the file-scope arms are the ones that need fixing.

The comment above the query at `health.rs:203` is the intended contract:

```
// Path scoping happens in SQL (#410); `all_edges` still spans the graph
// because fan-in/fan-out of a scoped node counts edges from outside it.
```

Your arms implement that. The file arms don't: they build `node_to_file` from the already path-filtered `nodes`, so `node_to_file.get(&e.source)` returns `None` for any endpoint outside the prefix and the edge is dropped. Under `--path` that makes them intra-prefix only, which also makes the whole-graph `get_all_edges()` directly above them dead weight — we pay for the full edge table and then discard exactly what the comment says we loaded it for.

So the 221-vs-3 figure on `handle_tool_call` is the size of that defect, not a divergence you introduced. Matching the file arms would propagate it into three more places and make it harder to unwind later. And fan-in that ignores external callers isn't a useful measurement anyway: the reason to scope by path is to ask which symbols in a module are most depended on, and most of the answer lives outside the module by construction.

Your `source == target` skip is the correct symbol-level analogue of the file arms' `src_file != tgt_file`, so you're following them everywhere they're right.

Merging as-is. Two follow-ups, neither blocking this:

1. The file-scope `fan_in`/`fan_out` path filtering. That sits in the same arms as #422, so the two are probably one change — would you take it?
2. The new tests don't exercise a `path`-filtered case, so the semantics we just settled on isn't pinned by anything. One test with `scope_prefix` set and a caller outside the prefix would lock it in.

I'll reword the CHANGELOG line myself — "one difference from the file-scope arms is deliberate" reads as though the file behaviour is correct, and it's the opposite.

Thanks for this one, and for #417, #422, #418 and #424 — that's a careful run of reports, and the SQL cross-checks in this PR made the decision above a five-minute one instead of an afternoon.

aovestdipaperino:
Follow-up landed: `28d42b9` (#426) fixes the file-scope arms, so the divergence I asked you to keep is now closed from the other side — the file arms count edges crossing a `--path` boundary the same way your symbol arms do, and `handlers/mod.rs` under `--path src/mcp` reports 490 against the 271 the old arms gave. Your 221-vs-3 measurement was the thing that made it obvious which side was wrong. Thanks for pushing back instead of just conforming.

PR #427, opened 2026-08-21
fix(extraction): parse C and C++ headers the way a preprocessor would

cezkid:
A `.h` was always extracted with the C grammar, and any macro-decorated declaration was not parsed at all. A C++ header therefore yielded no class, no method and no field node - the shape UE, Chromium, Qt and LLVM trees all spell. This routes a `.h` by its source dialect, blanks declaration macros before parsing, and fills in the class-body members the C++ walk dropped.

### Extraction

- route a `.h` by its source dialect, not by the extension alone
- widen the dialect markers: `::`, 11 more keywords, a default member init
- blank declaration macros space-for-byte before parsing, leaving a body or brace initializer alone, where a `MACRO(...)` is a call carrying edges
- keep a caps-named TYPE: `MYLIB_API const int x` still names a type past the qualifier and is blanked, `HANDLE const h` does not and is not
- lex raw string literals and digit separators, which otherwise desync the scan and leave every macro after them in the file unblanked
- extract globals declared with an array declarator
- extract the class-body members the C++ walk dropped: preprocessor-guarded members, a `friend` definition, a function-pointer member, a nested type, a `using` alias
- keep a signature on one line, single-spaced
- hand fingerprints and blame the bytes the extractor parsed, not the file's

### Resolution

- score C and C++ one language family: a `.cpp` calling into its own `.h` took 0.5 confidence, under the 0.6 floor, and lost the edge
- prefer the definition over the header declaration once both exist

### Redundancy

Candidate selection took only function/method/singleton-method nodes, so a header-only C++ template layer produced no candidate at all and the tool answered `scanned: 0` - indistinguishable from "no duplication found". A `template_declaration` node carries the whole body, function or class, and fingerprints like any other candidate.

### Measured

Over a 230-file UE-style C++ codebase, 112 of them headers:

| | before | after |
| --- | --- | --- |
| header symbols | 886 | 2604 |
| classes | 0 | 76 |
| methods | 0 | 760 |
| fields | 296 | 1105 |
| nodes | 3663 | 5381 |
| edges | 1903 | 2032 |
| signatures carrying a newline, tab or double space | 356 | 0 |

Every name the old index carried and this one does not is a type name the C grammar had recorded as a symbol.

### Tests

`cargo test --release` for `c_extraction_test` (31), `cpp_extraction_test` (44) and `resolution_test` (30) all pass; new fixtures cover each bullet above.


aovestdipaperino:
Thanks for this — and welcome. This is a badly broken case and the fix is well built. I reviewed it by building both this branch and `master` and indexing the same fixture, rather than reading alone; below is what I verified, one thing I'd like added before merge, and a note on where my review stopped.

## Verified

A UE-shaped header — `class MYGAME_API AActor : public UObject`, `UPROPERTY`/`UFUNCTION`-decorated members, a default member initialiser:

```
master:   file, include, preprocessor_def|MYGAME_API, static|MYGAME_API|5
          ...and nothing else. No class, no method, no field.

this PR:  class|AActor|5, field|Health|9, method|TakeDamage|12,
          method|GetName|14, field|speed|16
```

Your remark that every name the old index carried and this one does not is a type name the C grammar had recorded as a symbol checks out precisely: `static|MYGAME_API|5` is that artifact, and it is the *only* thing master got from the class.

## The invariant this rests on, and why it is right

`blank` preserving `\n` and `\r` while spacing over everything else is the load-bearing decision in the whole change, and it is the correct one. Byte offsets survive, which is what keeps `compute_fingerprint` comparable, and line numbers survive too, which is what keeps every stored `start_line` pointing where it did. Blanking a multi-line `MACRO(\n…)` to plain spaces would have silently shifted every symbol below it in the file. `blanks_the_class_macro_and_keeps_every_offset` asserts the length half of that; the newline half is what makes it safe in the multi-line case, and it holds by construction.

Routing fingerprints and blame through `source_for_parse` so they see the same bytes the extractor parsed is the other half of that, and skipping it would have made every C/C++ fingerprint fail to match across the two paths. Good catch.

The resolver change is small and aimed at a real dropped edge: `lang_from_path` maps `.h` to `c`, so a `.cpp` referencing its own header took the cross-language 0.5 and fell under the 0.6 floor. Merging c and cpp into one family is the narrow fix, and it does not weaken the guard for anything else.

## One thing I'd like added before merge

`is_header_path` treats `.inl`, `.ipp` and `.tcc` as headers, but `lang_from_path` does not map them — they still fall to `_ => "unknown"`. That matters more than it looks. As @vianbas documented on #346, an unmapped extension is exempt from **both** cross-language guards: alone it keeps 0.9 and clears the floor, and among several candidates it takes neither the `+50` nor the `-80`, so being unrecognised is a *scoring advantage* over the correct same-language candidate. Template-heavy C++ trees are exactly where `.inl` and `.tcc` files are common.

Adding `"inl" | "ipp" | "tcc" => "cpp"` to `lang_from_path` is a three-word change in a file you are already editing, and it closes the hole for the extensions this PR teaches the rest of the system to care about. Please fold it in and I will merge.

(There is a broader fix for that whole class on #346 — making the fallback the extension itself rather than `"unknown"` — but it is not yours to carry here.)

## A style note, not a blocker

The `.:` notation for "therefore" appears throughout the new comments (`tags differently .: without this…`). It reads as a typo on first pass and does not appear anywhere else in the codebase. If you are touching these files again anyway, spelling it out would help the next reader. Merging either way.

## Where my review stopped

I read and verified the paths where a mistake would be expensive — the blanking invariant, the fingerprint and blame routing, the resolver scoring, the header-dialect routing — and I confirmed the extraction outcome end to end. I did not read all 542 lines of `c_api_macro.rs` line by line, and the raw-string and digit-separator lexing in particular I am trusting your tests on rather than having independently exercised. Saying so plainly since "reviewed" should mean something specific.

The change is also broad: extraction, resolution, redundancy and blame in one PR. I would normally ask for that split, but the parts genuinely depend on each other to reach the outcome you measured, and splitting after the fact would cost you more than it buys anyone. Not asking.

cezkid:
Thank you for reviewing it that way. Building both sides and indexing the same fixture is more work than reading the diff, and it is the only kind of review that would have caught what I found afterwards. Both of your asks are folded in, the branch is rebased onto current `master`, and I then went back over the blanking scanner adversarially and found three defects of my own. Those are the substance of this update.

## Your two asks

`"inl" | "ipp" | "tcc" => "cpp"` is in `lang_from_path`, with a test pinning that a `.cpp` call into an `.inl` definition clears the 0.6 floor rather than scoring as `unknown`. Thank you for the pointer to what @vianbas documented on #346; I had not understood that an unmapped extension is a scoring *advantage* rather than merely a missing signal.

The `.:` notation is gone. You are right that it reads as a typo, and it is a habit from my own notes rather than anything this codebase does. Eighteen sites, all spelled out.

## What the rebase touched

`master` moved 15 commits under me. Two conflicts, both mechanical, but one is worth naming: in `tests/resolution_test.rs` git factored out the trailing `);\n}` that both sides shared, so a naive resolution left your `propagate_variant_edges_ignores_every_kind_but_annotates_and_calls` unterminated. The compiler caught it immediately. The CHANGELOG entry moved under the `### Fixed` heading from your one-heading-per-type reorganisation.

## Three defects in the blanking scanner

You said you were trusting the tests on the parts you did not exercise. That was the right instinct to flag, so I built an independent oracle rather than adding more fixtures of my own: a harness that runs `blank_declaration_macros` over a corpus and asserts that every rewrite preserves length, line count and character boundaries, is idempotent, and that every changed region reads back from the *original* bytes as a caps-named macro carrying at most its own argument list. The primitives in the checker are reimplemented rather than imported, so it is an oracle and not an echo of the code under test. Corpus: fmt, nlohmann/json, googletest, asio and wil at their current heads, 2,059 C-family files.

It found three shapes where blanking deleted live code. Each is the failure mode this PR claims to avoid, and each is silent: an offset-preserving rewrite that drops call edges without an error anywhere.

**A body opened past a trailing specifier was read as a declaration scope.** A `{` counted as a body only when the byte before it was `)`, `=`, `,` or `{`. So `void TearDown() override {` did not qualify, the brace was treated as a class-like scope, and the `EXPECT_EQ(1, n1_)` inside it was blanked out of existence. Same for `int Size() const {`, `void Run() noexcept {` and `auto Get() -> int {`. In `gtest_pred_impl_unittest.cc` alone that erased 15 call sites and took the file from 0 parse errors to 15. The rule is now the keyword that opened the brace rather than the punctuation before it: `class`, `struct`, `union`, `enum`, `namespace` and `extern` open a declaration scope, and everything else holds code. That is the same test `RecordBodyDepth` already makes in `header_dialect`, which is where I should have taken it from in the first place.

**A `:` ended a declaration, so a member initializer looked like an access label.** `Trace::Trace(Dispatcher& md) : MD(md), n_(0) {}` blanked `MD(md)` straight out of the initializer list, because `MD` is caps-named and the `:` before it read as `public:`. A qualified name was exposed the same way. Only `public`, `private` and `protected` end a declaration now.

**A macro that IS the declaration was recognised only by a following `{`.** That guard catches gtest's `TEST(S, C) {}` and missed two shapes: COM's `STDMETHOD(Read)(void* p) override { ... }`, where the tell is a *second* parameter list, and a lambda passed as a macro argument, where the body sits inside the parens and was blanked along with the call. One file of wil's test suite had 19 `STDMETHOD` declarations blanked head-first, taking the method each one declared with it. A following `(` now refuses, as does a `{` anywhere inside the argument list.

Six regression tests: one per shape above, plus two controls that the shapes this PR does want blanked still are.

Evidence after the fixes: across those 2,059 files no rewritten span contains a `;` or a `{`, and the region oracle passes on every file. Indexing all five trees before and after the three fixes gains 85 nodes and 67 edges and loses none. The only node kind that moves down is a `STDMETHOD` member, now correctly a method rather than a free function: in wil, 442 to 480 methods against 1,067 to 1,060 functions.

One methodological note, since it cost me time and might cost yours: **tree-sitter error-node counts are a bad oracle here.** Error *bytes* fell 28% corpus-wide, but individual files trade many small error nodes for one large error region and back again, in both directions, so the same file reads as improved on one metric and regressed on the other. Node and edge deltas from an actual index are the measurement that held up.

## One thing I added beyond the ask

Your `lang_from_path` change was unreachable as written, and so was the `is_header_path` entry it sat beside. `supported_extensions()` is built from the extractors, and `CppExtractor` claimed none of `.inl`, `.ipp` or `.tcc`, so the walker never opened one of those files and no node ever carried such a path. Meanwhile `hooks.rs` already lists `.ipp` and `.tcc` under "file extensions tokensave indexes". Everything above the extractor was ready; nothing produced a node.

I registered them, since template-heavy C++ is the case you named and half a fix there is worth less than none. They are C++ by construction, so they route by extension with no dialect sniff, and they map to `cpp` in `ts_lang_key_from_path` and `language_from_extension` too, so blame, fingerprints and the language label agree with the extractor that parsed them. Measured on two template-heavy trees at their current heads:

| | files | nodes | edges |
| --- | --- | --- | --- |
| asio | 75 | 0 -> 1,145 | 10,805 -> 13,002 |
| glm | 140 | 0 -> 2,989 | 1,560 -> 3,242 |

If you would rather this landed on its own, say so and I will pull it out into a separate PR against a merged version of this one. It is a self-contained commit.

## What I know is still missed

Two limits I can name rather than leave for the next person to find:

- **A paren-less export macro in front of a caps-named type is not blanked.** `WINBASEAPI HANDLE WINAPI CreateFileW(...)` stays as it is, because `names_a_type` needs either a type keyword or two identifiers past the macro and `HANDLE` supplies neither. This is under-blanking rather than corruption, and the conservative direction is the right one to fail in, but it is common in Win32 headers and it means those declarations still parse poorly.
- **`NLOHMANN_JSON_SERIALIZE_ENUM(Color, { ... })` is now preserved** where the version you reviewed blanked the whole mapping table. That follows from the brace rule and is right by this PR's own stated invariant, but it does mean tree-sitter now sees the invocation rather than whitespace.

And one inconsistency I did not carry here: `hooks.rs` also lists `c++` and `h++`, which no extractor claims either. Same shape as the `.inl` problem, outside what this PR's code names, and yours to schedule.

## Verification

`cargo test --release` is green across all 139 suites. `cargo clippy --all-targets --release` reports the same four warnings as `master` does, all in `upgrade.rs`, `guard.rs` and `agents/fs.rs`, none in a file this PR touches. `cargo fmt` applied.

aovestdipaperino:
Status note while the review is still open: `master` has moved again (#435, #444, #440, #438, #445 landed), and this branch now shows as conflicting.

The conflict is **`CHANGELOG.md` only** — I checked by merging locally rather than trusting the label, and no source or test file conflicts. It is the `### Fixed` list picking up new entries above yours, so the resolution is mechanical: keep both sides, your entry among them.

Nothing here asks you to redo review work — the three blanking-scanner defects you found and the two asks from the earlier round are still the substance in flight. Just flagging that a rebase (or a merge of `master`) will be needed before this can go in, and that it should cost you one file.

cezkid:
Rebased onto `5ac7522`.

Confirmed it cost one file: `CHANGELOG.md` was the only conflict, on three of the five commits, and no source or test file conflicted.

Two shapes in there, worth naming since only the first is the mechanical one you described:

- The first commit's conflict is the `### Fixed` list picking up your new entries above mine. Kept both sides, yours above mine, including the blank line your side carries between the `#446` and `#435` entries.
- The last two conflicts are not new entries at all. Both of those commits *amend* the same C++ header bullet rather than adding one, so keeping both sides would have duplicated it, and taking either side alone would have dropped a revision. Took mine in each, after checking rather than assuming: each later version is a strict superset of the earlier one, insertions only with no replacements, and the last commit's `CHANGELOG.md` diff is `1 insertion(+), 1 deletion(-)` on that single line.

Verified the rebase moved nothing but the changelog: the source diff against the new base is identical to the source diff against the old one, apart from blob hashes and hunk offsets.

Gates on the rebased tree, Windows, release profile: full suite `exit 0`, 0 failures across every target, 2m25s, with your new `literal_search_unscanned_test`, `branch_drift_test` and `hooks_test` all passing alongside it. `cargo fmt --all -- --check` clean, `cargo clippy --workspace --all-targets` clean.

Nothing in the review substance changed. The three blanking defects and the two asks from the earlier round are still what is in flight.

aovestdipaperino (commented):
Second review round. I rebuilt this against current `master` rather than reviewing the diff alone, so first the state, then three findings — graded honestly, because only one of them is worth your time.

## State: this is mergeable again

Your 2026-08-28 rebase still holds even though `master` has moved 13 commits since (#467, #453, #469, #470, #471 among them). I verified rather than trusting the label:

- `git merge-tree` against current `master`: **clean, zero conflicts**
- merged locally and built: `cargo build`, `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings` all clean
- full suite on the merge result: **144 suites, 0 failures**

Your handling of the earlier CHANGELOG rebase was right, including the part that wasn't mechanical. Noticing that the last two conflicts *amended* the same bullet rather than adding one — so keeping both sides would have duplicated it and taking either alone would have dropped a revision — is exactly the case where "keep both sides" as blanket advice would have quietly produced a wrong file. Checking that each later version was a strict superset instead of assuming it was the right instinct.

---

## 1. Blocking: `is_default_member_initializer` misroutes valid plain C

`src/extraction/mod.rs`, `is_default_member_initializer` — the rule is "≥2 words before an `=`, no `(` or `[`", evaluated on any line inside a `struct`/`union` body. An anonymous enum inside a struct is legal C and satisfies it:

```c
struct S {
    enum { KMax = 8 };
    int a;
};
```

`declaration` is `"    enum { KMax "` → two words (`enum`, `KMax`), ends alphanumeric, no paren or bracket → `true` → `header_dialect` returns `Some("C++")`.

I confirmed this by calling `header_dialect` directly rather than reasoning about it:

| fixture | result |
|---|---|
| anon enum in struct | **`Some("C++")`** ← wrong |
| plain struct | `None` ✓ |
| top-level enum | `None` ✓ |
| bitfield struct | `None` ✓ |

The three correct cases are worth noting too: single-identifier enumerators (`RED = 1`) are safe because they're one word, and `enum` doesn't open a tracked body, so only the *nested-in-a-record* case trips. That's a narrow hole rather than a broad one, which is why it survived your adversarial pass.

I'd rate the practical blast radius lower than "every node changes" — C is close enough to a C++ subset that `CppExtractor` will parse most such headers acceptably. But it's a real misclassification on valid, non-exotic C, and it's the one thing here I'd want fixed before merge. Requiring the word immediately before `=` to not be preceded by a `{` on the same line would close it; you'll have a better idea than I do.

## 2. Worth a look, not blocking: stale `.h` mapping in the stats breakdown

`src/db/queries/mod.rs`, `display_language_for_path` still has `"c" | "h" => "C"`, while this PR's whole premise is that a `.h` is now *sniffed* and may be indexed as C++ or Objective-C. So a header this PR correctly routes to `CppExtractor` still reports as "C".

I traced the callers before flagging it: the only non-test one is `get_stats`, so this is the language breakdown in stats and nothing else — no filtering or resolution depends on it. Cosmetic, but it's a number that now contradicts what the indexer actually did, and you're the person with the context to fix it cheaply.

## 3. Retracted — not a bug, flagging so you don't chase it

My first pass suspected that a `.h` sniffed as Objective-C is tagged `"c"` by `lang_from_path`, so refs from `.m`/`.mm` would hit the cross-language `score -= 80` path and lose their edges. **That does not happen**, and I'm recording why rather than deleting it, because the reasoning is a trap worth naming.

`.m` and `.mm` aren't in `lang_from_path` at all, so they resolve to `"unknown"` — and all three penalty sites (resolver ~890, ~970, ~1035) are guarded on *both* languages being known. Which is precisely the thing you flagged in your last comment: an unmapped extension is a scoring advantage, not a missing signal. The guard that makes your `.inl`/`.ipp`/`.tcc` mapping matter is the same guard that makes this a non-issue.

It is a **latent** trap, though: the day someone maps `.m`/`.mm` to an `objc` tag, ObjC headers tagged `"c"` start hitting the penalty, and `same_language_family` only covers c↔cpp. Not yours to fix here — worth a comment at that mapping so the next person sees it coming.

---

Only item 1 needs action from you. Items 2 and 3 are context so the next reader doesn't re-derive them.

aovestdipaperino:
Merged — thank you, @cezkid. This is a substantial piece of work and the writeup made it easy to review: routing a `.h` by its source dialect rather than its extension is the right call, and blanking declaration macros space-for-byte before parsing is a neat way to keep every offset honest. The before/after table over a real UE-style tree is exactly the evidence I wanted.

I ran the merge locally against current master before merging: no conflicts, and the full `cargo test --release` suite is green (150 suites, 0 failures), including the 31 C, 46 C++ and 32 resolution tests.

Two parts I especially appreciated: scoring C and C++ as one language family, which fixes a whole class of silently-dropped edges rather than just the headline bug, and the redundancy candidate fix — `scanned: 0` being indistinguishable from "no duplication found" was a real reporting bug hiding behind the extraction one.

Issue #413, opened 2026-08-18
TypeScript: `abstract class` declarations are never extracted, so the class, its methods, and every edge into them are missing


vianbas:
Measured on `3b294bc`, and present unchanged on `774479a` and on `67ff014`, so
this is not a regression from the recent receiver work. I found it while
checking what `3b294bc` still misses for [#378](https://github.com/aovestdipaperino/tokensave/issues/378);
it turned out to be the larger part of what I was seeing.

## What happens

A class declared `abstract` produces no `class` node and no `method` nodes at
all. A non-abstract class in the same file is extracted normally. Every method
defined on the abstract class disappears with it, so no call into those methods
can ever resolve, and every subclass loses the target of its `extends`.

## Fixture

```bash
mkdir -p /tmp/ts-abstract/src && cd /tmp/ts-abstract
cat > src/kelas.ts <<'TS'
abstract class Lokal {
  satu(): string {
    return "lokal";
  }
}

export abstract class Diekspor {
  dua(): string {
    return "diekspor";
  }
}

export class Biasa extends Diekspor {
  tiga(): string {
    return "biasa";
  }
}
TS
tokensave init
sqlite3 .tokensave/tokensave.db \
  "SELECT kind || ' ' || name FROM nodes WHERE kind IN ('class','method') ORDER BY 1;"
```

Expected, six rows:

```
class Biasa
class Diekspor
class Lokal
method dua
method satu
method tiga
```

Actual on `3b294bc`, two rows:

```
class Biasa
method tiga
```

`Lokal` and `Diekspor` are gone, `satu` and `dua` are gone with them, and
`Biasa` survives while the class it extends does not exist in the graph.

I ran this twice per binary on three binaries. The output was identical each
time, and identical across `67ff014`, `774479a`, and `3b294bc`.

## Cause

`visit_node` in `src/extraction/typescript_extractor.rs:175` dispatches on
`"class_declaration"`. tree-sitter-typescript 0.23.2, the version in
`Cargo.lock`, gives `abstract class X {}` its own node kind,
`abstract_class_declaration`, which no arm matches. The node is walked past and
its whole body with it. The same file handles `interface_declaration`,
`enum_declaration`, and `type_alias_declaration` explicitly, so this reads like
an omission rather than a decision.

## In real projects

Library code only, test and fixture paths excluded, indexed with `3b294bc`:

[...]

PR #428, opened 2026-08-21
fix(typescript): extract abstract method declarations (#424)

### LINKED ISSUE #424: TypeScript: an `abstract` method declaration produces no node, so a call into it loses its edge when the base has more than one subclass
Measured on master `0651c0a`, release build of that commit, macOS, built from
source. That commit is `v7.10.0` plus one documentation-only change, so the
released binary behaves the same here. I found this while checking what
`e9084f1` left open after it closed
[#413](https://github.com/aovestdipaperino/tokensave/issues/413) and
[#414](https://github.com/aovestdipaperino/tokensave/issues/414).

## What happens

A method declared without a body inside an `abstract class` is not extracted.
The class itself is extracted, and so is every sibling method that has a body.
Only the bodiless declaration is dropped, and nothing is recorded about it.

The visible consequence is at the call site. When the abstract base has two or
more subclasses, a call through a receiver typed as the base produces no `calls`
edge at all. There is no declaration node to resolve to, so the resolver sees
only the subclass implementations, cannot separate them, and refuses the tie.
The call is recorded in `ambiguous_calls` and `tokensave tool callees` on the
caller returns an empty list.

With exactly one subclass the call still lands, because the single same-named
implementation resolves by name. That is why this shape is easy to walk past.

## Fixture

```bash
mkdir -p /tmp/ts-absmethod/src && cd /tmp/ts-absmethod
cat > src/kelas.ts <<'TS'
export abstract class Basis {
  abstract dua(): string;
}

export class TurunanA extends Basis {
  dua(): string {
    return "a";
  }
}

export class TurunanB extends Basis {
  dua(): string {
    return "b";
  }
}

export class Pemakai {
  private b: Basis;
  constructor(b: Basis) {
    this.b = b;
  }
  jalan(): string {
    return this.b.dua();
  }
}
TS
tokensave init
sqlite3 .tokensave/tokensave.db \
  "SELECT kind, name, start_line FROM nodes ORDER BY start_line;"
tokensave tool callees \
  "$(sqlite3 .tokensave/tokensave.db "SELECT id FROM nodes WHERE name='jalan';")"
```

There are two `method` nodes named `dua`, one per subclass, and none for the
declaration on `Basis`. `callees` prints `[]`.

Use `v7.10.0` or later. On `v7.9.0` the same fixture shows something else for a
different reason: `class Basis` is absent altogether, which is #413 before its
fix, and `callees` returns one result pointing at `TurunanA.dua`. Both
`e9084f1` and the resolution change in `e8a8a8a` landed after that tag.

Now give that method a body. Replace `abstract dua(): string;` with
`dua(): string { return "basis"; }` and leave everything else alone. The node
appears and `callees` returns it. The class is still declared `abstract`, so
what decides the outcome is the member, not the keyword on the class.

## The kind it should map to already exists

`AbstractMethod` is already a node kind here, and the resolver already accepts
it as a call target: the `kind_compatible` gate documented in `CHANGELOG.md`
lists abstract-method among the callables a `Calls` reference may resolve to.
`tokensave_doc_coverage` and `tokensave_port_status` both include
`abstract_method` in their kind lists.

Two other extractors already produce it. A Java fixture
(`public abstract String dua();`) gives a node of kind `abstract_method`. So
does Kotlin: indexing moeru-ai/airi at `faf96ef` yields two `abstract_method`
nodes, both from `HostWebSocketBridge.kt`, and none from any `.ts` file in the
same tree.

## Where it comes from

`visit_class_body` in `src/extraction/typescript_extractor.rs` matches three
child kinds: `decorator`, `method_definition`, and `public_field_definition`.
Every other child falls through the catch-all arm. The grammar gives a bodiless
declaration its own kind, `abstract_method_signature` (tree-sitter-typescript
0.23.2, pinned in `Cargo.lock`), so it never reaches an arm that would record
it. I looked for a trace of the dropped declaration and found none: no row in
`nodes`, `edges`, `unresolved_refs`, or `ambiguous_calls` refers to its line.

An interface body takes a different path. `visit_interface_body` handles
`method_signature` and hands it to `visit_interface_method`, which is why the
same bodiless shape written inside an `interface` is extracted and why a call
through an interface-typed receiver resolves and produces dispatch rows for
every implementation.

## How much of this is out there

| project | revision | abstract declarations in source | `abstract_method` nodes in `.ts` |
|---|---|---|---|
| nestjs/nest | `2fe907a` | 51, across 10 abstract classes | 0 |
| typeorm/typeorm | `df07bf1` | 9, across 6 abstract classes | 0 |
| moeru-ai/airi | `faf96ef` | 2, in 1 abstract class | 0 |

The left column comes from the TypeScript compiler's own AST, walking every
`ClassDeclaration` carrying the `abstract` modifier and counting members that
are `MethodDeclaration` with no body. Test directories, `.d.ts` files, and
`declare` contexts are excluded. I ran it under TypeScript 5.9.3 and 5.6.3 and
got the same counts from both. The right column comes from indexing each tree
twice with the release build described above.

## What the tests added with `e9084f1` cover

`tests/ts_abstract_and_ctor_param_test.rs` has two tests whose names sound like
they reach this: `abstract_classes_and_their_methods_are_extracted` and
`a_call_into_an_abstract_base_method_resolves`. Every method in both fixtures
has a body, and the second fixture declares no subclass at all, so neither
exercises a bodiless declaration. I ran the file on `0651c0a` and all five of
its tests pass, while the fixture above fails.

## Offer

I am happy to send a patch that adds `abstract_method_signature` to
`visit_class_body` and maps it to `NodeKind::AbstractMethod`, with a regression
test that uses two subclasses. I have not written it yet, so I cannot tell you
whether the new node is enough on its own to make that call edge land, or
whether the resolver needs more. Say the word and I will find out and open it.

vianbas:
## Summary

- A method declared without a body in an `abstract class` now becomes an `AbstractMethod` node instead of being dropped, so a call through a base-typed receiver has something to resolve to.
- Closes #424.

## Motivation

`abstract render(): string;` parses as `abstract_method_signature`. `visit_class_body` matched `decorator`, `method_definition` and `public_field_definition`, and everything else fell through the catch-all arm, so the declaration reached nothing. The same bodiless shape inside an `interface` is extracted, by `visit_interface_body`, and the Java and Kotlin extractors already emit `AbstractMethod` for the identical construct, so TypeScript was the odd one out rather than the kind being unsupported.

The visible cost is at the call site once the base has more than one subclass. With no declaration to resolve to, the resolver sees only the implementations, cannot separate them, and refuses the tie, so the call produces no edge and `tokensave_callees` on the caller returns nothing. With exactly one subclass the call still lands on that implementation by name, which is why the shape hid behind the #413 fixtures, where every method has a body.

## Changes

- `src/extraction/typescript_extractor.rs`: an `abstract_method_signature` arm in `visit_class_body`, and `visit_abstract_method` modelled on `visit_interface_method`, emitting `NodeKind::AbstractMethod` with the member's accessibility, JSDoc, signature text and `Contains` edge. Decorators attach the same way they do for a `method_definition`.
- `tests/ts_abstract_and_ctor_param_test.rs`: one regression test, two subclasses so the single-implementation name path cannot mask the defect.

The test asserts the **edge**, not the node. You raised the reason on the issue: kind filtering runs ahead of scoring since `ab9ff29`, and your experiment produced a `method` node because the member had a body, so a node of a newly emitted kind reaching the graph does not by itself mean a call can resolve to it. It does. On the fixture the call resolves to the declaration on the base rather than to either implementation, which answers the question you left open.

## Test plan

- [x] `cargo test` passes: 3,373 passed, 0 failed, 0 ignored across 139 binaries, `--no-fail-fast`.
- [x] `cargo clippy --all-targets` has no new warnings, and no warnings at all.
- [x] Tested manually. Indexing the issue's fixture gives `abstract_method` at line 1 alongside the two `method` nodes at 5 and 11, and the `calls` edge from `jalan` lands on the `abstract_method`. Reverting only the extractor change, with the test left in place, fails the new test; restoring it passes. Both checked twice.

## Checklist

- [x] No breaking changes. Existing nodes and edges are unchanged; this only adds nodes that were previously absent, and the three Godot tests, the Ruby and TypeScript extraction suites, and the resolution suite are all unaffected.

One correction to something I said on #424: I offered to take the second follow-up from #423, the `scope_prefix` test for the symbol arms. I would rather hand that back to you and stop at this one, so please do not hold it for me.


aovestdipaperino:
Reviewed against `master` by building both commits and indexing fixtures with each. The extractor change is right and it follows the surrounding conventions closely — merging.

Two things I checked that you may not have, one of which needs a correction to the PR body.

## `parent_id: None` is correct here, and I had to convince myself of that

It reads like a bug next to the `Contains` edge, so for the record: `visit_method` and `visit_interface_method` both do exactly the same thing, and migration v9 (`src/db/migrations.rs:912`) denormalises `parent_id` from `Contains` edges after the fact. So your edge is what makes the abstract member count as a member of its class — 19 nodes in this repository's `.ts` files carry a `parent_id` no extractor ever wrote. Nothing to change; I mention it so the next reader doesn't file it as a defect.

Using `extract_ts_accessibility` rather than hardcoding `Visibility::Pub` the way `visit_interface_method` does is also the right call, since `protected abstract` is a real and common shape.

## The claim "existing nodes and edges are unchanged" is not quite true

Your checklist says no breaking changes and that this only adds nodes that were previously absent. It also retargets an existing edge, in the single-subclass case you correctly identified as the one that already worked:

```
one.ts:  abstract class Base { abstract solo(): string; }
         class Only extends Base { solo() {...} }
         class User { go() { return this.b.solo(); } }

master:  go --calls--> method:solo:4          (the implementation)
this PR: go --calls--> abstract_method:solo:1 (the declaration)
```

Indexed twice on each build. So `tokensave_callees` on `go` returns a different symbol after this change for code that has one subclass — which is the common shape, not the corner case. Your two-subclass fixture goes from *no edge* to *an edge*, an unambiguous gain; the one-subclass path goes from *the implementation* to *the declaration*.

I think the new behaviour is the better one and I am not asking you to change it. At a base-typed call site the declaration is the honest answer — you cannot know the implementation there — and it makes classes behave like interfaces, which already resolve this way. But it is a change in existing output rather than a pure addition, and it should say so.

## The consequence to document alongside it

Nothing links the declaration to its implementations. The only edge between the two classes is `extends` at class level; there is no method-level override edge:

```
class:Only        --extends--> class:Base
method:go         --calls-->   abstract_method:solo
abstract_method -> method:     0 edges
```

So a `callees` walk now terminates at the declaration instead of reaching executable code, and `Only.solo` has no callers at all. I checked whether that produces `dead_code` false positives and it does not — `dead_code_count` stayed 0 on a fixture that also held a deliberately dead exported function, so exports are exempt and the implementations are not flagged. Worth knowing rather than assuming, given #346.

Whether an `overrides` edge from a declaration to its implementations is worth adding is a separate question, and a real one — it is what would make this navigable in both directions. Not for this PR.

## What I'm doing

Merging as-is and amending the CHANGELOG entry myself to say that the single-subclass edge is retargeted and that the walk stops at the declaration — one sentence, not worth a round trip. Everything else in that entry is accurate, including the three-corpus counts, which I did not re-derive but which match the numbers you already corrected twice on the issue.

Noted on the #423 `scope_prefix` test — I'll pick it up, along with the file arms, which are already done in #426. Thanks for this, and for stopping at one PR rather than taking everything I offered.

vianbas:
A correction to the checklist in this PR's body, from my own measurement after
the merge.

I ticked "Existing nodes and edges are unchanged; this only adds nodes that were
previously absent." That is not true. @aovestdipaperino named one way it fails
when he merged, and there is a second way that is not written down anywhere.

Measured on `53627d7`, this PR as it landed, against its parent `28d42b9`, two
runs a side, in both profiles, and repeated on current master `8fa53ce` with the
same result. The fixture is two files:

```ts
// base.ts
export abstract class Base {
    abstract solo(): number;
}
export class Impl extends Base {
    solo(): number {
        return 1;
    }
}

// user.ts
import { Base } from "./base";
export class User {
    private b: Base;
    constructor(b: Base) { this.b = b; }
    run(): number { return this.b.solo(); }
}
```

Rows one to three vary only in how the receiver `b` is declared in `User`. The
fourth adds a second subclass to `base.ts` and keeps the first row's `User`.

| how `b` is declared | before | after |
|---|---|---|
| `private b: Base;` | 1 edge into the implementation's `solo` | 1 edge into the declaration's `solo` |
| `constructor(private b: Base)` | 1 edge into the implementation's `solo` | 1 edge into the declaration's `solo` |
| not declared at all | 1 edge into the implementation's `solo` | **no edge** |
| `private b: Base;`, base has two subclasses | **no edge** | 1 edge into the declaration's `solo` |

The last row is what this PR set out to fix and it works. The first two are the
retarget you described: the edge survives but points somewhere else. The third
is the one I should have caught before I ticked that box, because it is a loss
rather than a move.

One more thing the same runs show, which changes how the first two rows should
be read: after this PR, all three receiver shapes record a row in
`ambiguous_calls` for `this.b.solo`, and before it none of the three did. So in
the first two rows the edge is emitted and the ambiguity is recorded at the same
time, and in the third only the ambiguity is recorded. The information is not
lost, but a consumer counting `calls` edges sees the third case as absent. The
fourth row is the other way round: before this PR it recorded two ambiguous
candidates and no edge, which is what #424 was about, and after it records one
and resolves.

This is a follow-up to #424, whose own case is the last row. Under
`tsc --noEmit --strict`, 5.6.3, rows one and two are clean and row three is
`error TS2339: Property 'b' does not exist on type 'User'`.

I am not asking for a change. `abstract_method` is arguably the better target
for a call through a `Base`-typed receiver, and row three does not typecheck.
What I did not want to leave standing is a checklist line of mine that says
nothing moved, when the runs I did afterwards show two ways it did.

PR #438, opened 2026-08-22
fix(python): emit refs for module-scope calls and self-bound method references

map588:
## Summary

`tokensave_dead_code` reported live Python symbols for two reasons, both in the Python extractor.

**1. Calls at module scope produced no `calls` edge.** Call sites were extracted only from function bodies. A function called from a module-level assignment RHS (`_KEYMAP = _build_keymap()`), a bare statement (`_setup()`), or under `if __name__ == "__main__":` had no incoming edge. Module- and class-scope statements — simple and compound (`if`, `try`, `with`, `for`, `while`) — now run `extract_call_sites`, attributed to the File node or the enclosing class. `extract_call_sites` stops at nested defs, so nothing is attributed twice.

**2. Methods passed by reference through `self` produced no ref.** The value-position scan (#224) stopped at every `attribute` node, so `Thread(target=self._flush_loop)` or `schedule(self._tick, 1.0)` gave the method nothing. A `self.name` / `cls.name` attribute in a value position now emits a `uses` ref qualified with the enclosing class (`Daemon::_flush_loop`), so the resolver matches that class's method and not a same-named method elsewhere. Any other `obj.attr` is still skipped.

No change to the `dead_code` query: `uses` already counts as liveness in `find_dead_code_inner`.

## Test

`tests/python_module_scope_refs_test.rs` runs the real `tokensave_dead_code` tool over a tempdir project, same pattern as `python_bug224_test.rs`. Each case was run against the unchanged extractor first and failed; each keeps a dead control symbol so a fix that suppressed all output would not pass.

- Before: `["_truly_dead", "_build_keymap", "_setup", "_run"]` and `["_tick_timer", "_truly_dead", "_handle_scan", "_flush_loop"]`
- After: only `_truly_dead` in each.

Full suite: 3393 passed, 0 failed. `cargo fmt --check` clean.


aovestdipaperino:
Merged as `2c843f5`.

Reviewed the two mechanisms and the shapes they could have got wrong, and they hold up:

- The new catch-all arm in `visit_node` cannot double-attribute anything, because function-body statements never reach `visit_node` — bodies go through `extract_call_sites`/`visit_nested_defs` — and `extract_call_sites` stops at `function_definition`/`class_definition`, so a def inside a module-level `if` still keeps its own calls.
- `visit_assignment` runs `scan_value_positions` on the RHS, not call extraction, so an assignment statement gets its callee edge exactly once.
- The qualified name built by `enclosing_class_qualified_name` matches what `qualified_prefix()` produces for the method node itself, file-path duplication included, so the `uses` ref lands on the right class rather than relying on suffix matching.
- `python_class_attrs` is pushed and popped by `visit_class` only, so a closure inside a method still sees the enclosing class — which is right, since a closure genuinely captures `self`.

The care in the `self.<name>` gate is what made this mergeable as-is: distinguishing a method reference from a field read via `assigned`, and then exempting `@property`/`.setter` descriptors from that exemption, is the difference between fixing the false positives and trading them for false negatives. Same for putting the unreferenced class in the file that indexes first, so the cross-file case would fail against a first-suffix-match resolver instead of passing by luck.

Thank you for the fix, and for measuring each case against the unchanged extractor first — the before/after symbol lists made the review much shorter.

PR #445, opened 2026-08-23
feat: extract Ruby define_method as first-class method nodes

janstol:
## Summary

- A statically-resolvable `define_method`/`define_singleton_method` site now emits a first-class method node, the same shape `visit_method` produces for an equivalent `def`.
- Covers block, `do...end`, lambda, and `proc`/`Proc.new` second-argument forms; `class << self`; `module_function` interaction; `private`/`protected`/`public`/`private_class_method`/`public_class_method` targeting; and expression-position sites (assignment RHS, call argument, receiver, array/hash element), not just statement position.

## Motivation

Previously these DSL-defined methods produced no node at all, making them invisible to `search`, `module_api`, `callers`, `rename_preview`, `dead_code`, and `doc_coverage`, and leaving every call to them permanently unresolvable — worse than a missing edge, since there was no node to hang anything on at all.

## Changes

- `src/extraction/ruby_extractor.rs`: a `classify_define_method_site` classifier (name resolution, receiver shape, second-argument type, argument count) shared by node emission, a `MethodBody` block-traversal scope so nested definitions inside the site's body get a fresh method-body frame instead of leaking the enclosing class's ambient state, and dispatch from both statement and expression position with guards against double-emission and against second-guessing `private`/`private_class_method` wrapper decisions.
- `tests/ruby_extraction_test.rs`: extensive coverage — positive forms, visibility interactions, `module_function` interaction, body attribution (calls/complexity/nested defs), and negative cases (dynamic names, wrong argument types/counts, unreachable positions).

## Test plan

- [x] `cargo test` passes
- [x] Tested manually: synced this feature branch's binary against a local checkout of the Rails repository (`tokensave sync -f`) and confirmed `Rails::Generators::NamedBase#check_class_collision`, defined via `define_method` inside a `def self.check_class_collision` class-macro body, now gets its own instance-method node with a resolved outgoing call — previously invisible to the graph entirely.

## Checklist

- [x] Breaking changes documented (if any) — none; extraction-layer only, no schema/resolver/MCP change


aovestdipaperino:
Merged.

Verified against current `master` rather than the PR's own base, since #444 landed in `ruby_extractor.rs` in the meantime: merged locally, then `ruby_extraction_test` (301), the full suite, and clippy all clean.

The part of this that earned the merge as-is is the `self_is_instance` hoisting. Moving the realignment above `extract_call_sites` in both `visit_method` and `visit_singleton_method` is not a `define_method` change at all — it fixes an inconsistency that predates this PR, where the same nested site was judged under one value during the call-site walk and another during the real body traversal. Doing that first, and saying so in the comment, is what made the new `MethodBody` scope reviewable instead of load-bearing on an accident.

The visibility mirror-imaging is the other detail worth recording: pairing `define_singleton_method` with `private_class_method`/`public_class_method` rather than with `private`/`protected`/`public`, and emitting nothing for the two combinations Ruby raises on (P11/P19), means the graph matches what actually gets defined instead of what the wrapper appears to say. The `nodes.len()` guard before `last_mut()` is a real bug avoided too — without it a declined emission would have repointed an unrelated node's visibility.

Thank you for the fix, and for the Rails check on top of the fixtures. `check_class_collision` defined through a `define_method` inside a class-macro body is exactly the shape the synthetic cases can't prove, and having a real symbol come back with a resolved outgoing call is worth more than another fixture.

Issue #400, opened 2026-08-17
MCP server never re-resolves the active branch: indexes another branch's files into the started-on branch DB, silently


aovestdipaperino:
Split out of #395, reported by @vianbas. This is the serious half of that issue.

## Summary

A `tokensave serve` process started on `main` keeps writing to **`main`'s** database after the user checks out a tracked branch. It continues reporting `active_branch: main`, and emits no warning on either the tool response or stderr. Files that exist only on the other branch get indexed into `main`'s DB.

The index is not merely stale — it describes a tree that has never existed.

## Reproduction (reporter's, verified across five cooldown windows)

Server started on `main` → `tokensave branch add feature` → `git checkout feature` → commit a new file on `feature` → make a tool call:

```
db 'feature': ['src/base.rs']
db 'main'   : ['src/base.rs', 'src/on_feature.rs']
active_branch reported by tokensave_status: main
warning in tool response: False
warning on server stderr : False
```

`src/on_feature.rs` is committed on `feature` and exists nowhere else.

Polled across five cooldown windows without restarting, to rule out lazy re-resolution on a later call:

```
t= 36.3s  active_branch=main  feature=['src/base.rs']  main=['src/base.rs', 'src/on_feature.rs']
t= 71.1s  active_branch=main  feature=['src/base.rs']  main=['src/base.rs', 'src/on_feature.rs']
t=106.4s  active_branch=main  feature=['src/base.rs']  main=['src/base.rs', 'src/on_feature.rs']
t=141.1s  active_branch=main  feature=['src/base.rs']  main=['src/base.rs', 'src/on_feature.rs']
t=176.4s  active_branch=main  feature=['src/base.rs']  main=['src/base.rs', 'src/on_feature.rs']
stderr over 176s: 0 WARNING
after restart: feature=['src/base.rs', 'src/on_feature.rs']
```

Zero warnings over 176 seconds is what sets the severity: there is no signal at all.

## Note on what a restart actually fixes

Two different repairs, and only the first is about branch resolution:

- Restarted while `feature` is checked out: the new process resolves `feature`, and its index becomes correct.
- Restarted while `main` is checked out: the new process's sync finds `src/on_feature.rs` absent from disk and drops the stray row — the deletion being noticed, not the branch being re-resolved.

## Relationship to #372

[...]

PR #447, opened 2026-08-24
fix(mcp): refuse queries after branch drift

### LINKED ISSUE #446: MCP server must refuse local graph queries after tracked-branch drift
### tokensave version

7.10.0

### Operating system

macOS

### Installation method

Built source from the local fork.

### What happened?

A long-lived MCP server opened one tracked branch database, then the working tree switched to another tracked branch and local graph tools continued answering from the original branch.

## Problem

A long-lived MCP server opens one branch database at startup. If the working tree later switches to another tracked branch, ordinary local graph tools can return plausible results from the wrong branch, including stale symbols from a deleted branch.

## Impact

Agents may trust incorrect search, context, caller, impact, diagnostic, or analysis results. The existing automatic-sync safeguard prevents corrupt writes but does not prevent stale read answers.

## Proposed behavior

Detect serving-branch drift before local graph work and return an actionable `InvalidRequest` naming both branches and recommending a server restart or reopen. Keep `tokensave_status` available for diagnosis. Preserve explicit external `graph_root` and `graph_branch` read-only selections.

## Acceptance

- Local graph-consuming MCP tools fail closed after a tracked-branch checkout.
- The refusal names the serving branch, working-tree branch, and recovery action.
- `tokensave_status` remains callable.
- Single-database projects and explicit external graph selections remain unaffected.
- Regression tests cover selector-capable and non-selector graph tools.

## Reproduction

1. In a project with multi-branch indexing, track two branches with `tokensave branch add` and sync both.
2. Start an MCP server while the working tree is on branch A.
3. Check out branch B without restarting the server.
4. Call a local graph tool such as `tokensave_search`. It answers from branch A's database even though the working tree is on branch B. If branch A is then deleted, its database keeps serving symbols that no longer exist anywhere.

The write side of this drift already refuses (#400, #403): automatic sync will not write branch B's files into branch A's database. This issue is the read side of the same condition.

rNoz:
## Motivation

Prevent plausible MCP graph answers from the wrong tracked branch after a checkout under a long-lived server.

## Summary

Fail closed when a long-lived MCP server serves one tracked branch while the working tree has moved to another tracked branch.

## Changes

- Add a request-time branch-drift gate before freshness work and handler dispatch.
- Cover selector-capable tools and graph-consuming tools without selectors.
- Preserve `tokensave_status`, single-database behavior, and explicit external graph selections.
- Add live MCP regression coverage for refusal details, status availability, and the complete local graph-tool matrix.
- Document recovery behavior in the user guides and changelog.

## Issue

Resolves #446.

## Test plan

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets`
- `cargo test --features test-transport --test strict_tree_mismatch_test -- --nocapture`
- `cargo test --features test-transport --test branch_drift_test -- --nocapture`
- `cargo test graph_scope --lib -- --nocapture`

The full workspace suite was attempted. An unrelated `git_hook_isolation_test` failure detected a developer-level global post-commit hook running inside its test fixture.

## Checklist

- [x] Focused regression tests pass.
- [x] Clippy and formatting pass.
- [x] User documentation and changelog updated.


aovestdipaperino:
Landed on `master` as `d7b2eea`, and #446 is closed by it.

Squashed locally rather than merged through the button: the branch conflicted on `CHANGELOG.md` only (the `### Fixed` list moved under it), so GitHub could not squash it. Verified before landing — merged onto current `master`, then `cargo test --features test-transport --all-targets` and clippy clean; the four `strict_tree_mismatch_test` cases including `branch_drift_refuses_local_graph_tools_by_default` all pass. Note for next time: the test plan lists `--test branch_drift_test`, but the coverage actually lives in `strict_tree_mismatch_test.rs` — there is no `branch_drift_test` file in the diff.

**On the default change**, which is the substance here. Making this refuse for everyone rather than gating it behind `strict_tree` is the right call, and the argument that convinced me is the one your issue makes: the two halves of a single condition disagreed. Sync has refused to *write* across drift since #400/#403, while reads answered from the same wrong database behind a warning — and a warning is not a control an agent instructed to prefer tokensave over reading files can act on. The deleted-branch case settles it: an index answering with symbols that exist nowhere has no reading under which its output is useful.

So I have expanded the changelog entry rather than keeping the four-line version. A default that changes for every user has to be findable later by someone who did not read this thread — including what happened to `strict_tree`, which keeps its remaining job (the separate different-worktree mismatch, still advisory).

**One thing to follow up on, not a merge blocker.** `LOCAL_GRAPH_TOOLS_NOT_SUPPORTING_SELECTORS` is a hard-coded list of six names. A future local graph tool that takes no selector will not be gated, and nothing will fail — the gate will just quietly not cover it, which is the same silent-wrong-answer shape this PR exists to remove. Deriving that set (or asserting it against the tool registry the way `graph_scoped_defs_match_the_canonical_set` pins the selector-capable set) would make it self-maintaining. Happy to take that as a follow-up issue if you want to pick it up.

Thank you for this one, and for keeping `tokensave_status` callable in the design rather than as an afterthought — a fail-closed gate with no diagnostic left open would have traded a wrong answer for a dead end.

aovestdipaperino:
Correction to my note above: I said there is no `branch_drift_test` file. There is — `tests/branch_drift_test.rs`, added by #403 for the write-side refusal. It is not in this PR's diff because you did not need to change it, and running it was the right thing to list in the test plan, since your gate depends on the `branch_drift()` detection those tests pin. My inference from the diff was wrong; the test plan was accurate as written. Nothing about the merge changes.

rNoz:
Implemented + merged your follow-up proposal  https://github.com/aovestdipaperino/tokensave/pull/539

Issue #420, opened 2026-08-19
tokensave can revert agent settings, but there is no option to remove commit hooks


jasonb194:
### tokensave version

7.10.0

### Operating system

Windows

### Installation method

GitHub release binary

### What happened?

I can install and uninstall tokensave form any agent, but there is only an install option for github commit hooks. 
Uninstalling from all agents should include the option to uninstall github hooks, and there should be a way to remove github hooks without uninstalling tokensave from an agent.

### Steps to reproduce

1. install tokensave on one or more agents with commit hooks
2. init in a local repo
3. uninstall tokensave globally
4. remove tokensave files from repo
5. commit any change
6. tokensave files recreated

aovestdipaperino:
Fixed on `master` (`cfacd73`, #429). You were right on both counts, and your repro was the useful part — step 6 is only mysterious until you notice that step 3 never touched the hook.

Two ways to remove them now:

```bash
tokensave githooks              # show which hooks tokensave owns, and where
tokensave githooks off          # remove them
tokensave uninstall             # removes them along with all agent integrations
tokensave uninstall --keep-git-hooks   # ...or keep them
```

`githooks on` installs them without a prompt, for symmetry.

Two cases deliberately do **not** remove them: `uninstall --local`, since the hooks are global rather than project-scoped, and `uninstall --agent X`, since another agent may still be installed and relying on them. Your ask was that uninstalling from *all* agents should include them, and that is what the default now does.

Removal is conservative, because it edits files tokensave does not own — your git hooks and your global gitconfig:

- A hook file holding anything you wrote keeps that content and loses only tokensave's marked section. Tested with a `pre-push` that had its own guard; the guard survived, the tokensave forwarder went.
- A file that is nothing but tokensave's own content is deleted.
- A hook file with no tokensave marker is not rewritten at all, not even reformatted.
- `core.hooksPath` is unset, and the hooks directory removed, only when it points at tokensave's own default, the directory is left empty, **and** the run actually removed a tokensave hook. That last condition matters on your setup: if you had set `core.hooksPath` to `~/.config/git/hooks` yourself, a `githooks off` that found nothing of ours must not edit your config, and the path alone cannot say who wrote it. The cost is that an already-empty leftover directory survives; that is inert, since git just finds no hooks in it.

If the directory still holds hooks tokensave did not write, it stays and the command tells you so rather than leaving you to guess.

[...]

Issue #164, opened 2026-07-02
`install` sets global `core.hooksPath`, silently disabling existing `init.templateDir` / `.git/hooks` hooks


rNoz:
### tokensave version

7.0.3

### Operating system

macOS

### Installation method

Homebrew (brew)

### What happened?

Accepting the "Install global git post-commit + post-checkout hooks?" prompt during `tokensave install` **sets a global `core.hooksPath`** (`~/.config/git/hooks`) whenever no `core.hooksPath` is already configured. Git then uses that directory
*exclusively* and ignores every repository's `.git/hooks/` — which is where hooks delivered via `git config --global init.templateDir` are copied at clone/init time. The user's existing hooks (e.g. a dotfiles-managed `post-checkout` that
configures identity/SSH keys) silently stop running, with no detection, warning, or opt-out.

Nothing is deleted (`write_global_hook` appends when a file exists), so this is not file corruption — but `core.hooksPath` and `init.templateDir`/`.git/hooks` are **mutually exclusive** in git, and tokensave silently switches the user onto the former. Only tokensave's one-line `post-checkout` (`tokensave init`) then runs.

### Steps to reproduce

1. `git config --global init.templateDir ~/.config/git/template`, with a    `hooks/post-checkout` present, and **no** `core.hooksPath` set.
2. `tokensave install`, and accept the git-hooks prompt.
3. `git config --global core.hooksPath` now returns `~/.config/git/hooks`.
4. `git clone <repo>` — the template `post-checkout` no longer runs; only   `tokensave init` does.

### Error output

```shell
$ git config --global init.templateDir
~/.config/git/template
$ git config --global core.hooksPath        # before install: empty
$ tokensave install                         # accept the git-hooks prompt
✔ Set git core.hooksPath to ~/.config/git/hooks
$ git config --global core.hooksPath        # after install:
~/.config/git/hooks
# → git now ignores each repo's .git/hooks/, so the templateDir post-checkout
#   never runs again; only ~/.config/git/hooks/post-checkout (tokensave init) does.
```

### Component

CLI

### Additional context

Root cause — `src/agents/mod.rs`, `offer_git_post_commit_hook`:

```rust
let hooks_dir = read_global_hooks_path(&home);
let (hooks_dir, need_set_hookspath) = match hooks_dir {
    Some(dir) => (dir, false),
    None => (home.join(".config").join("git").join("hooks"), true),
};
// ...
if need_set_hookspath {
[...]

PR #453, opened 2026-08-26
feat(git-hooks): add a post-merge hook so git pull reindexes automatically

devofthings:
## Summary

`tokensave install`'s git-hook offer wires `post-commit` (after your own commits) and `post-checkout` (fresh clones, branch switches), but never `post-merge` — so a plain `git pull` that fast-forwards or produces a merge commit doesn't trigger a reindex. A teammate's changes land on disk with nothing to run `tokensave sync`, and the index stays stale until your next local commit.

Concrete case this bit: check out a PR branch to review it, switch back to your own branch, author pushes fixes and re-requests review, you check out the PR branch again and `git pull` — today nothing reindexes at that point.

This adds `post-merge` as a third owned hook, installed/removed alongside `post-commit`/`post-checkout` under the same `--git-hook` opt-in, `tokensave githooks on`/`off`, and `tokensave uninstall --keep-git-hooks`. It chains to a repo's own `post-merge` hook (e.g. Husky-managed) the same way the other two already do.

## Test plan
- [x] `cargo test --lib` — 819/819 passing, including 2 new tests for the post-merge snippet and its removal
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --lib -- -D warnings` clean


aovestdipaperino:
Landed as `dd8696c` — squashed locally with your authorship preserved, since the branch conflicted on `CHANGELOG.md` only (the `[Unreleased]` Added section moved under it while this was in review). GitHub won't auto-close a PR merged this way, hence closing by hand; the code is all yours and byte-identical to your branch — I diffed `src/agents/hooks.rs`, `src/branch.rs` and `src/cli.rs` against `pr453` to be sure before pushing.

Thanks @devofthings. The gap was real and the diagnosis was exactly right: `post-commit` covers your own commits and `post-checkout` covers clones and switches, so a fast-forward `git pull` fell through both — which is the single most common way a teammate's code arrives on disk. The review round-trip you describe (check out a PR branch, switch back, pull the author's fixes) is the case where a stale index is most likely to mislead, so this is worth more than the one-hook diff suggests.

Two things I particularly appreciated:

- Removing `post-merge` from `FORWARDED_REPO_HOOKS` rather than leaving it in both lists. A pure forwarder and an owned hook in the same slot would have double-written, and it would have been easy to miss.
- The chaining preamble and its own idempotency marker, so an existing install migrates on the next `install` and a Husky-managed `post-merge` keeps firing. Claiming a global `core.hooksPath` silently disabling a repo's own pull hook is exactly the #164 failure, and you didn't reintroduce it.

The removal-path test (`strip_removes_the_post_merge_section_from_a_pure_tokensave_hook`) was the right thing to include too — `githooks off` and `uninstall` had to learn about the new hook, and #420 exists because removal was the half that got forgotten last time.

CI was green on all five checks before landing.

Issue #480, opened 2026-08-30
PreToolUse grep hook still denies the whole command when the extra work is a substitution or a redirect (#475 follow-up)


dangeReis:
### tokensave version

7.10.0 (master @ 35f8b59)

### Operating system

macOS

### Installation method

Built from source

### What happened?

#476 fixed the `&&` / `||` / `;` / `|` shapes of #475: a search chained to other work
now passes through instead of denying the whole command. Two shapes are still outside
`has_chained_command`'s scan, so the guarantee has holes — **command substitution** and
**redirection**:

| command | master | consequence |
|---|---|---|
| `grep -n my_function src/lib.rs && ./deploy.sh` | allow | #476, working |
| ``grep -n my_function src/lib.rs `date +%s` `` | **deny** | the substitution never runs |
| `grep -n my_function src/lib.rs "$(cat patterns.txt)"` | **deny** | same, quoted form |
| `grep -n my_function src/lib.rs > /tmp/hits` | **deny** | `/tmp/hits` is never written |

`has_chained_command` scans for the sequencing operators only, so a line whose extra
work sits inside `$( )`, backticks, or after a `>` still reads to
`extract_grep_invocation` as a bare search and short-circuits to a deny.

The cost is the same one #475 named. A denial returns the whole command unrun, and the
agent takes the redirect and moves on — so the substituted command never runs, and any
later step that reads `/tmp/hits` finds nothing. The redirect cannot fulfil the request
it replaced.

`2>/dev/null` denies too and is the benign end of this: nothing downstream depends on
it, and the tokensave suggestion is a fair substitute. If you want to keep denying that
one, the distinction is stdout redirection versus stderr suppression.

This is pre-existing on master, independent of #451 — I verified it against an
unpatched `aovestdipaperino/master` build, output below.

### Steps to reproduce

```bash
git clone https://github.com/aovestdipaperino/tokensave && cd tokensave
cargo build --bin tokensave

mkdir -p /tmp/demo/src && cd /tmp/demo
echo 'fn my_function() {}' > src/lib.rs
tokensave init

printf '{"tool_name":"Bash","cwd":"/tmp/demo","tool_input":{"command":"grep -n my_function src/lib.rs > /tmp/hits"}}' \
  | /path/to/tokensave/target/debug/tokensave hook-pre-tool-use
```

### Error output

```
--- grep -n my_function src/lib.rs && ./deploy.sh
allow
[...]

Issue #475, opened 2026-08-30
PreToolUse grep hook denies the whole command when the search comes first: the trailing work is discarded


dangeReis:
### tokensave version

7.10.0

### Operating system

macOS

### Installation method

GitHub release binary

### What happened?

#451 is the half of this where the guardrail goes **silent**: a search behind another
command is allowed. This is the other half, and it fails in the expensive direction —
a search *in front of* another command **denies the whole command**, so the trailing
work never runs.

`extract_grep_invocation` only requires the command to *start* with `grep`/`rg`/`ag`
once `strip_command_prefixes` has peeled its known wrappers. It then classifies the
rest of the line as that grep's arguments, so a trailing `&& …` is invisible to it:

| command | decision | consequence |
|---|---|---|
| `git status && grep -n MySymbol src/lib.rs` | allow | #451: guardrail silent |
| `grep -n MySymbol src/lib.rs && git status` | **deny** | the `git status` never runs |
| `grep -n MySymbol src/lib.rs && ./deploy.sh` | **deny** | the deploy never runs |
| `cd src && grep -n MySymbol lib.rs && ./deploy.sh` | **deny** | same, through the `cd` prefix path |

A denial returns the entire command unrun, so the cost isn't a redirected grep — it's
the discarded `./deploy.sh`. That is the failure mode you named when you set the
semantics for #451: *"a whole call must never be rejected for a trailing diagnostic
grep … the guardrail would be destroying work, which is a different and much worse
failure than the one it exists to prevent."* The rule you gave there — judge by side
effects, not by position — is exactly what fixes this too; it just isn't reachable
while the leading-`grep` shape short-circuits before any splitting happens.

This is pre-existing behaviour on 7.10.0, not something #451's patch introduces. I
verified that by running the same inputs against the unpatched tree.

### Steps to reproduce

Run from inside a project with a `.tokensave` index, with `TOKENSAVE_DISABLE_GREP_HOOK`
unset, substituting a symbol that exists in it.

1. Search first, work second — denied, and the trailing command is discarded:

```
echo '{"tool_name":"Bash","tool_input":{"command":"grep -n MySymbol src/lib.rs && git status"}}' \
  | tokensave hook-pre-tool-use
```

2. The same two commands in the other order — allowed (this is #451):

```
echo '{"tool_name":"Bash","tool_input":{"command":"git status && grep -n MySymbol src/lib.rs"}}' \
  | tokensave hook-pre-tool-use
```

[...]

Issue #452, opened 2026-08-26
PreToolUse hook fires on 4.4% of real greps: Grep/Read have no handler, and the code-directory allow-list is 8 hardcoded names


vibhor-allen:
# PreToolUse hook fires on 4.4% of real greps: `Grep`/`Read` have no handler, and the code-directory allow-list is 8 hardcoded names

**tokensave** v7.10.0 · installed via the `easy-headroom` VS Code extension v0.6.2 · macOS 15 (Darwin 25.6.0) arm64 · Claude Code

## Summary

The `hook-pre-tool-use` nudge is working as designed but almost never fires. I replayed **915 unique real `grep` commands** (pulled from my Claude Code transcripts) through the hook binary. It returned `deny` for **40 of them — 4.4%**.

Roughly half the misses are correct: they target `.sql` / `.md` / `.json`, which tokensave does not index. The other half are greps over ordinary source files that tokensave could have answered.

Three separate causes, each reproducible below. All verdicts are from piping a payload straight into `tokensave hook-pre-tool-use` in an indexed project.

---

## 1. `Grep` and `Read` are in the installed matcher but have no handler

The extension installs the matcher `Agent|Grep|Bash|Glob`, so the native `Grep` tool is routed to the hook. Every payload shape returns `allow`:

```jsonc
{"tool_name":"Grep","tool_input":{"pattern":"my_function"}}                   // -> allow
{"tool_name":"Grep","tool_input":{"pattern":"my_function","path":"."}}        // -> allow
{"tool_name":"Grep","tool_input":{"pattern":"my_function","path":"main.py"}}  // -> allow
{"tool_name":"Grep","tool_input":{"pattern":"my_function","glob":"*.py"}}     // -> allow
```

Only `grep` embedded in a `Bash` command is handled. Since Claude Code's native `Grep` is the *preferred* search tool, this is the largest single gap.

Adding `Read` to the matcher behaves the same way — every Read payload returns `allow`, including large source files:

```jsonc
{"tool_name":"Read","tool_input":{"file_path":"/abs/path/big_module.py"}}     // -> allow  (1550-line file)
```

**Expected:** either handle `Grep`/`Read`, or drop `Grep` from the matcher it installs so the routing matches the implementation.

## 2. The code-directory allow-list is 8 hardcoded names

When the grep target is a directory, detection depends on the directory's *name*. Same pattern, same repo, only the target differs:

```bash
grep -rn my_function .            # -> deny  ✅
[...]

PR #485, opened 2026-08-30
fix(hooks): see a search batched behind another command (#451)

### LINKED ISSUE #451: PreToolUse grep hook's Bash branch misses compound commands: any other leading command hides a symbol-shaped search
### tokensave version

7.10.0

### Operating system

macOS

### Installation method

GitHub release binary

### What happened?

The Bash branch of `hook-pre-tool-use` recognizes a code search only when the command *starts with* `grep `/`rg `/`ag ` after `strip_command_prefixes` peels its known wrappers (`rtk`/`sudo`/`time`/`nice`, `NAME=value` assignments, a leading `cd … &&`/`cd … ;`).

Any other leading command in a sequence therefore hides the search:

| command | decision |
|---|---|
| `grep -n MySymbol src/lib.rs` | deny (correct) |
| `echo x && grep -n MySymbol src/lib.rs` | allow |
| `true; grep -n MySymbol src/lib.rs` | allow |
| `ls && grep -n MySymbol src/lib.rs` | allow |
| `git status && grep -n MySymbol src/lib.rs` | allow |

It shows up in practice because agents batch searches for efficiency. The shape that surfaced this wasn't a crafted bypass, it was a routine labelled batch:

```
echo "=A=" && grep -n SymbolOne src/one.rs
echo "=B=" && grep -n SymbolTwo src/two.rs
```

Every segment is a symbol search on an indexed project and none are redirected, so an agent that batches greps gets no steering at all. That was my case: I ran a multi-line labelled batch of greps against an indexed repo and the guardrail stayed silent, where each grep on its own line would have been redirected.

I realise #238's scope note already anticipated this direction — it says going further "trends toward a real shell parser (next would be `;` chains, subshells, `$(...)`, aliases, functions), which the guardrail deliberately avoids." So this may be working as designed, and I'm filing it as a data point rather than a defect claim.

Environment: 7.10.0, arm64, installed via `tokensave upgrade`.

### Steps to reproduce

Reproducible against the binary directly, no agent needed. Run from inside a project that has a `.tokensave` index, substituting a symbol that exists in it, and with `TOKENSAVE_DISABLE_GREP_HOOK` unset.

1. Confirm the plain form is denied:

```
echo '{"tool_name":"Bash","tool_input":{"command":"grep -n MySymbol src/lib.rs"}}' \
  | tokensave hook-pre-tool-use
```

2. Put any command in front of it:

```
echo '{"tool_name":"Bash","tool_input":{"command":"echo x && grep -n MySymbol src/lib.rs"}}' \
  | tokensave hook-pre-tool-use
```

3. Compare the two decisions.

### Error output

```shell
# 1. plain grep -> denied
#    (abridged: the real payload also carries agent_message and a nested
#     hookSpecificOutput block with the same text)
{"permission":"deny","user_message":"STOP: This Bash grep targets a code file in a
tokensave-indexed project and the pattern `MySymbol` looks like a symbol name. Use
tokensave_search (definition) or tokensave_callers_for (usages) instead ..."}

# 2. same search behind any leading command -> allowed (exact output)
{"permission":"allow"}
```

**Expected:** case 2 is redirected exactly like case 1 — it is the same search.
**Actual:** case 2 is allowed and the hook does not fire.

### Component

Agent integration (Claude, Cursor, etc.)

### Additional context

I looked at fixing this and did not find a patch I was confident enough in to send.

Splitting on `&&`/`||`/`;` and denying any segment that is a search means a whole Bash call gets rejected because of a trailing diagnostic grep — `git checkout -b x && rg Sym src/` would never create the branch. That seemed like exactly the false-positive pressure already visible in #449 and #448. A narrower rule matching only `echo <literal> && <search>` handles the minimal repro but misses the interleaved batch above, which is the shape that actually occurs.

So the question I'd rather ask than guess at: given #238's scope note, is fail-open on compound commands still the intended tradeoff now that batching makes it easy to hit unintentionally? If it is, this can be closed — a line in the docs noting the Bash branch is best-effort for single commands might save the next person the investigation. If the batching shape changes the calculus, I'm happy to put up a focused patch with tests once you've said what the desired semantics are.

dangeReis:
Fixes #451.

`extract_grep_invocation` required the command to *begin* with `grep`/`rg`/`ag` once its prefix stripper had run, so anything in front of the search hid it completely:

```
echo "=== A ===" && grep -n SymbolOne src/one.rs
printf ready; rg MySymbol src/
ls && grep -n MySymbol src/lib.rs
```

all passed through untouched. That isn't a crafted bypass — it's how anyone greps several symbols in one call, since a label before each search is the natural way to read the combined output. The guardrail was silently absent from a common real shape while working on the single-command spelling beside it.

## Approach

Per your spec on the issue: the Bash branch now falls back to splitting on top-level `&&`, `||` and `;` and classifying each segment on its own.

The **whole command is still tried first, unchanged**. That's the only path that models a leading `cd` and resolves the grep target against it; a split would read `cd src` as an ordinary command and lose that. `a_leading_cd_is_still_modeled` is the ordering guard.

A denial returns the entire command to the caller unrun, so the search is redirected **only when every other segment is inert**. Unknown ⇒ side effects ⇒ fail open: `git checkout -b x && rg Sym src/` still passes through, because blocking it would throw away the checkout to save a grep (`a_segment_that_does_real_work_is_left_alone`).

Subshells, `$(...)`, pipes, redirects, newline lists and background jobs fall through to allow rather than get partially parsed (`unmodeled_shell_shapes_fall_through`). A segment that is a search but whose pattern isn't symbol-shaped stays the pattern classifier's call, not the splitter's.

## One deviation from your list, and why — worth your call

You gave `echo, true, :, pwd, ls, cat, printf`. I shipped that list, but **`true` and `:` are inert only *ahead of* the search**, not after it.

Implementing your table verbatim fails your own merged test. `hook_475_test.rs:65` pins `grep -n MySymbol <file> || true` as pass-through, and adding `true` outright turns it red — I ran it:

```
---- search_followed_by_other_work_passes_through stdout ----
chained command must not be denied: grep -n MySymbol …/src/lib.rs || true
```

The two positions are genuinely different things. Ahead of the search, `true` carries no work, so `true; grep -n Sym src/lib.rs` is a batched search and your row is right. After it, the controller is consuming the *search's* exit status — `grep … || true` is how a failing search is stopped from failing a script, which is exactly what #475/#476 guarantee is left alone.

The split loop already knows which side of the search it's on (`reason.is_none()`), so both hold instead of trading off. Both directions are pinned, and the leading-position assertion fails without the change. If you'd rather have the simple flat list and accept `true; grep …` passing through, say so and I'll cut it — it's four lines.

## On deriving the list

You asked me to flag it if I saw a way to avoid a second hardcoded command list — this is the same smell @not-antoni raised about `CODE_DIRS` in #452. I don't see one. Inertness isn't derivable without executing the command or shipping a shell semantics model, and both are worse than seven names. What keeps it honest is the failure direction: the list only ever *adds* denials, and anything unrecognised allows, so a missing entry is a miss rather than eaten work.

## Tests

`tests/hook_451_test.rs`, 9 tests: your four table rows, the subshell fall-through, the non-symbol-shaped search, the `cd` ordering guard, the headline multi-symbol batch, and both positions of the exit-status controllers. Full suite green locally (158 binaries, 0 failures), `cargo fmt --check` and `cargo clippy --all-targets` clean.

Unrelated but adjacent: your comment here says substitutions should fall through to allow, which is what #480 reports isn't happening. Happy to take that next if you want it.


aovestdipaperino:
Merged — thanks @dangeReis. Trying the whole command first so a leading `cd` still resolves the target, and pinning that ordering with `a_leading_cd_is_still_modeled`, is the detail that would have quietly regressed otherwise.

Thanks too for not implementing my inert-command list verbatim. You were right that `true` and `:` are inert only *ahead of* the search — after it they're consuming the search's exit status, and my own #475 test says so. Good catch, and good that you showed the red test rather than just asserting it.

PR #487, opened 2026-08-31
fix(hooks): let a substitution or a redirect through the grep guardrail (#480)

### LINKED ISSUE #480: PreToolUse grep hook still denies the whole command when the extra work is a substitution or a redirect (#475 follow-up)

dangeReis:
Fixes #480

## The rule this restores

#475 settled that a search is only ever a *suggestion* to ask the graph instead, so the hook must never veto a line it does not fully model — a denial returns the **entire** command unrun, and destroying a `./deploy.sh` to save a grep is the worse failure by a wide margin.

`has_chained_command` is where that rule lives, but it only looked for `&&`, `||`, `;`, `|` and a backgrounding `&`. Those are not the only ways a line carries work the hook cannot serve:

| command | before | after |
|---|---|---|
| `grep -n my_function src/lib.rs && ./deploy.sh` | allow | allow (#476, unchanged) |
| ``grep -n my_function src/lib.rs `date +%s` `` | **deny** — the substitution never runs | allow |
| `grep -n "$(cat patterns.txt)" src/lib.rs` | **deny** — the substitution never runs | allow |
| `grep -n my_function src/lib.rs > /tmp/hits` | **deny** — no file written | allow |
| `grep -n my_function src/lib.rs 2>/dev/null` | deny | deny (deliberate, see below) |

Substitutions are detected inside double-quoted regions as well as bare, since `$(…)` and backticks expand there. Single quotes still suppress them, so a metacharacter in a quoted pattern is not mistaken for shell syntax.

## Why `2>` stays deniable

- An output redirect (`>`, `>>`, `1>`) is a **file the caller asked for**. Denying it produces no file — the #475 failure exactly.
- `2>` / `2>>` **discards** the search's own stderr. Nothing downstream depends on it, so the line still *is* just the search, the graph is a genuine substitute, and the suggestion should stand.

So `2>`/`2>>` is consumed before the redirect arm sees it, and `stderr_suppression_is_still_deniable` pins that reasoning in a test rather than leaving it incidental behaviour. If you read it the other way, deleting the `'2'` arm is the whole change.

`2>&1` continues to pass through, as it did before — the `&` reaches the operator arm.

## Tests

`tests/hook_480_test.rs`, 5 tests. Verified fail-first: reverting `src/hooks.rs` turns 4 of the 5 red. The fifth (`stderr_suppression_is_still_deniable`) pins behaviour this change does *not* alter, so it correctly stays green.

`quoting_does_not_by_itself_open_the_gate` asserts the quoted pattern is still denied *before* asserting the redirected version falls through — without the first half the second proves nothing.

The `find`/`fd` path gets the same treatment, since it shares the scanner.

`cargo test` clean (no failures across the suite), `cargo fmt --check` clean, `cargo clippy --all-targets` produces no new warnings — the operator arm lists all five characters in one arm rather than splitting them, to avoid a `match_same_arms` hit.

## One thing I could not pin, and why

I first tried to test that `grep -n "a > b" src/lib.rs` is still denied, to prove quoting keeps a metacharacter out of the scanner. That fails — but not because of this change. A pattern containing `>` is not symbol-shaped, so the pattern classifier allows the command before the redirect scanner is ever consulted. `quoting_does_not_by_itself_open_the_gate` replaces it and asserts the layer that this change actually governs.

## Relationship to #485

Independent — this branch is based on `master`, not on #485. The one test I had that exercised the interaction between the two (a redirect surviving #451's segment split) needs `hook_451_test.rs`, which only exists on #485, so I dropped it here. If #485 lands I'll send that assertion as a one-line addition to `hook_451_test.rs`, where it belongs.


aovestdipaperino:
Merged — thanks @dangeReis. The `2>`/`2>>` carve-out is the right line: an output redirect is a file the caller asked for, discarded stderr is not, and I'm glad the reasoning is pinned by `stderr_suppression_is_still_deniable` instead of left incidental.

Also appreciated the section on the assertion you couldn't pin and why the pattern classifier gets there first. That's more useful than a test that passes for the wrong reason.

PR #489, opened 2026-08-31
fix(githooks): exit non-zero when a hook could not be installed (#488)

### LINKED ISSUE #488: `githooks on` prints a failure and still exits 0 when a hook cannot be written
`tokensave githooks on` reports a hook it could not write, then exits `0`. A CI step or a wrapper script that checks the exit code concludes the hooks are installed. Both the `--local` and the global path are affected.

Reproduced on `master` (`af9183e`, a debug build), macOS. The unwritable-hook condition is faked by creating a *directory* where the hook file goes, which is the simplest portable way to make the open fail.

### `--local`

```
T=$(mktemp -d); git -C "$T" init -q; mkdir -p "$T/.git/hooks/post-commit"
(cd "$T" && tokensave githooks on --local); echo "exit=$?"
```

```
  ✘ Failed to open …/.git/hooks/post-commit for writing
  ✔ Installed git post-checkout hook at …
  ✔ Installed git post-merge hook at …
exit=0
```

### global

```
H=$(mktemp -d); mkdir -p "$H/.config/git/hooks/post-commit"
(cd "$T" && HOME=$H XDG_CONFIG_HOME=$H/.config tokensave githooks on); echo "exit=$?"
```

```
  ✘ Failed to open …/.config/git/hooks/post-commit for writing
  ✘ Failed to open …/.config/git/hooks/post-commit for writing
exit=0
```

(The global path prints the same failure twice — a separate cosmetic point, but it suggests the write is attempted twice.)

## Mechanism

`write_global_hook` returns `bool` and prints the reason, but the call site inside `install_local_git_hooks` (`src/agents/hooks.rs`) drops the `false`. The hook is then merely absent from `installed`, which is indistinguishable from one that was never attempted, so nothing downstream can tell success from failure. The global path in `src/main.rs` has the same shape.

## Suggested shape

Carry the failures rather than fail fast: add a `failed: Vec<String>` alongside `installed` in the existing `LocalHookInstall` report struct, push to it in the `else`, and have the explicit `githooks on` command return `Err` when it is non-empty.

Fail-fast on the first failure would be wrong here — `install_local_git_hooks` installs three hooks, and an early return would discard the report of the ones that did succeed. The struct's own doc comment says it exists "so the caller can report it".

`offer_local_git_hooks` (the interactive `init` prompt) should stay best-effort and unchanged; making a prompt fail the command invents a new failure mode for people who just wanted to decline.

Happy to send a PR for this — I have the fix and tests for both paths ready.

dangeReis:
Fixes #488

Both paths printed a red `✘ Failed to open … for writing` and then exited 0, so a setup script gating on the exit code carried on with no hooks installed and no way to tell.

- The **global** path called `write_global_hook`, which returns a `bool`, in an `if` with no `else`.
- The **local** path collected `installed` and `already_present`, but had nowhere to put a hook whose write failed.

## The change

`offer_git_post_commit_hook` returns `Result<(), String>`; `LocalHookInstall` gains a `failed` list; `githooks on` turns either into a non-zero exit naming the hooks.

Failures are **collected**, not raised at the first one — the command installs three hooks, and bailing early would skip the two the user also asked for. The specific reason is still printed where it happens; the returned message only exists so the process can exit non-zero.

What deliberately stays a success:

| case | result |
|---|---|
| user declines the prompt | `Ok` |
| hook already present | `Ok` |
| hook write fails during `install` | `Ok` — best-effort (`let _ =`), a hook must not fail a whole install |
| explicit `githooks on` could not write a hook | **`Err`, non-zero exit** |

## Tests

`tests/githooks_exit_code_test.rs` runs the binary, because the bug was the wiring: the library could know perfectly well that the write failed and the process still exited 0. Both tests are red against unfixed `src/` and green with it.

Library-level: `local_git_hooks_test` (the `failed` list is populated), `global_git_hooks_failure_test` (nothing installable → `Err`), `global_git_hooks_partial_failure_test` (one unwritable hook, the other two still install). Each was verified fail-first by neutering *its own* code path while keeping the signature, so a compile error isn't being mistaken for a red test.

`cargo test` clean across the suite, `cargo fmt --check` clean, `cargo clippy --all-targets` produces no new warnings.

Independent of #487 and #485 in code — different files, based on `master`. The one overlap is the CHANGELOG: both this and #487 add a `### Fixed` section under `[Unreleased]` at the same anchor, so whichever lands first, I'll rebase the other onto it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), n
 
 
 
+### Fixed
+
+- **`githooks on` no longer exits 0 after failing to install a hook (#488).** Both paths printed a red `✘ Failed to open … for writing` and then returned success, so a setup script gating on the exit code carried on with no hooks installed and no way to tell. The global path called `write_global_hook`, which returns a `bool`, in an `if` whose `else` did not exist; the local path collected `installed` and `already_present` but had nowhere to put a hook that failed. `offer_git_post_commit_hook` now returns `Result<(), String>` and `LocalHookInstall` carries a `failed` list, which `githooks on` turns into a non-zero exit naming the hooks. Failures are collected rather than raised at the first one: the command installs three hooks, and bailing early would skip the two the user also asked for. Declining the prompt, an already-present hook, and the best-effort call inside `install` all stay success — only an explicit `githooks on` that could not do what it was asked now fails.
+
 ## [7.11.0] - 2026-08-30
 
 ### Added
--- a/src/agents/hooks.rs
+++ b/src/agents/hooks.rs
@@ -118,17 +118,21 @@ fn install_repo_hook_forwarders(
     hooks_dir: &Path,
     claiming_hookspath: bool,
     hooks_dir_is_default: bool,
-) {
+) -> Vec<&'static str> {
+    let mut failed = Vec::new();
     if !(claiming_hookspath || hooks_dir_is_default) {
-        return;
+        return failed;
     }
     for name in FORWARDED_REPO_HOOKS {
         let path = hooks_dir.join(name);
         if path.exists() {
             continue;
         }
-        write_global_hook(&path, &chain_repo_hook_snippet(name));
+        if !write_global_hook(&path, &chain_repo_hook_snippet(name)) {
+            failed.push(*name);
+        }
     }
+    failed
 }
 
 /// Whether the chaining preamble should be added to a global hook file.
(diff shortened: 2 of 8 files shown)


aovestdipaperino:
Merged — thanks @dangeReis. Collecting the failures rather than bailing on the first is right for a command that installs three hooks, and the split between best-effort inside `install` and non-zero for an explicit `githooks on` is exactly what a gating setup script needs.

The binary-level test earns its place: the bug was the wiring, so the library could know the write failed and the process would still exit 0 — a library-only test would have stayed green.

PR #490, opened 2026-08-31
fix(git): bound diff payloads before serializing, not after (#486)

### LINKED ISSUE #486: `diff_context` truncates large responses into invalid JSON
## Version

TokenSave 7.10.0 on Linux, built from source.

## What happened?

When `diff_context` exceeds the 15,000-character response limit, TokenSave cuts the serialized JSON and appends:

```text
[... truncated at 15000 chars]
```

The command exits successfully, but the output is no longer valid JSON.

## Reproduction

In an indexed repository with sufficiently connected files:

```bash
tokensave tool diff_context --files src/foo.ts --files src/bar.ts --files test/foo.test.ts > /tmp/result.json
```

```bash
jq -e . /tmp/result.json
```

Result:

```text
jq: parse error: Unfinished string at EOF
```

TokenSave exits with status 0.

## Expected behavior

Successful structured output should always be valid JSON. Large results could return a bounded page with pagination metadata:

```json
{
  "data": {
    "items": []
  },
  "pagination": {
    "current_page": 1,
    "page_size": 25,
    "total_items": 143,
    "next_cursor": "dc_01JQ8Z4V7M6K2P9X3T5N8R1C0A"
  }
}
```

On the final page, `next_cursor` would be `null`. The cursor could retain the original query, position, and index snapshot.

At minimum, TokenSave should exit nonzero rather than returning malformed JSON.

## Component

CLI

## Additional context

This appears related to #185 and #186, which fixed `tokensave_diff` composing already-truncated delegate results. That fix still truncates the final public response at the serialization edge.

Issue #124 discusses a similar bounded-output and continuation approach.

I reproduced this with 7.10.0. The v7.11.0 source still uses `truncate_response` to slice the serialized string and append the plain-text marker, so the latest tag appears to retain the same behavior.

aovestdipaperino:
Fixes #486.

## Problem

`diff_context` built its payload, pretty-printed it, then handed the string to `truncate_response`, which slices at 15,000 chars and appends `[... truncated at N chars]`. Past that limit the text callers parse as JSON is a half-written object — `jq -e .` fails and the CLI still exits 0. `changelog`, `commit_context`, and the `diff` aggregator had the same shape.

## Fix

`serialize_bounded_json(value, shedable)` bounds the payload *before* serialization: it halves the longest shedable list each round (O(log n) serializations, not one per dropped element) until the pretty-printed form fits, and records what happened under a top-level `truncated` object:

```json
"truncated": { "impacted_symbols": { "shown": 40, "total": 900 } }
```

`impacted_symbols_count` still carries the true total, so the bound is legible in the data. Shed order is declared per delegate, least useful first — for `diff_context` that is `impacted_symbols`, then `affected_tests`, then `modified_symbols`; `changed_files` is the caller's own input and stays whole. Dotted paths (`changes.impacted_symbols`) let `tokensave_diff` bound its nested payload through the envelope. If nothing is shedable and the remainder still exceeds the limit, the response is a JSON note rather than a sliced object.

Not full cursor pagination as the issue sketches — that needs a query/snapshot cursor store and is a larger change — but the output is now always valid JSON with the truncation described in-band.

## Not covered

`commit_context`'s `symbols_by_role` is a map of arrays, not an array, so it is not shedable; a commit touching enough files to blow the limit on that field alone falls back to the bounded note. Noted in a comment on the constant.

## Tests

Five unit tests in `src/mcp/tools/handlers/mod.rs`: oversized payload parses and reports `shown`/`total`, small payload untouched, shed order respected, dotted-path shedding, and the nothing-shedable fallback. `cargo test --lib`, `cargo test --test mcp_handler_test` (196 passed), and `cargo clippy --all-targets -- -D warnings` all clean.

### DIFF
--- a/src/mcp/tools/handlers/git.rs
+++ b/src/mcp/tools/handlers/git.rs
@@ -7,7 +7,7 @@ use std::collections::{HashMap, HashSet};
 use serde_json::{json, Value};
 
 use super::super::ToolResult;
-use super::{truncate_response, unique_file_paths};
+use super::{serialize_bounded_json, truncate_response, unique_file_paths};
 use crate::errors::{Result, TokenSaveError};
 use crate::tokensave::TokenSave;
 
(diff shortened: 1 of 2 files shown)


aovestdipaperino:
Landed on `master` as f81b189/edb5b55 (rebased onto the #485/#487/#489 merges, so the SHA differs from this branch). Closing this PR — the fix is in.

Thanks @humemm for the report: the pointer to `truncate_response` slicing at the serialization edge, and the note that #185/#186 only moved that edge rather than removing it, is exactly what made this a short fix.

PR #505, opened 2026-09-05
fix(mcp): do not respond to unhandled notifications

okunnig:
## Summary

Per [JSON-RPC 2.0 §4.1 (Notification)](https://www.jsonrpc.org/specification#notification):
> "A Notification is a Request object without an 'id' member. [...] no Response object MUST be returned to the client. The Server MUST NOT reply to a Notification, including those that are within a batch request."

Previously, `McpServer::handle_request` only matched `initialized` and `notifications/initialized` to return `None`. Any other notification—such as `notifications/roots/list_changed` (sent by clients when workspace roots update) or custom notifications without an `id`—fell through to the wildcard `_ =>` arm and returned:
```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32601,"message":"method not found: ..."}}
```

Strict JSON-RPC 2.0 implementations, including the official Go MCP SDK (`github.com/modelcontextprotocol/go-sdk`) used by Antigravity and Gemini CLI, validate that incoming responses without a method must carry a valid non-null ID. When they receive `id: null`, they return `ErrInvalidRequest` (`-32600: invalid request`) and abruptly terminate the connection, dropping all TokenSave tools.

## Changes

1. **`JsonRpcRequest::is_notification`**: Added helper to detect notifications (`id.is_null()`, `initialized`, or methods prefixed with `notifications/`).
2. **`McpServer::handle_request`**:
   - Explicitly listed `notifications/roots/list_changed` alongside `initialized` and `notifications/initialized`.
   - Updated the wildcard match arm to check `request.is_notification()` and return `None` (silently discarding notifications) rather than sending an error response with `id: null`. Requests that carry an ID still receive standard `-32601 MethodNotFound` error responses.
3. **Tests**:
   - Added unit test `test_is_notification` in `src/mcp/transport.rs`.
   - Added integration tests in `tests/mcp_server_test.rs`:
     - `test_notifications_roots_list_changed`
     - `test_unhandled_notification_produces_no_response`
     - `test_unhandled_request_with_id_produces_method_not_found`
4. **`CHANGELOG.md`**: Documented fix under `[Unreleased] -> [Fixed]`.


aovestdipaperino:
Merged — thank you, @okunnig. The JSON-RPC citation and the Go SDK detail made this easy to verify: replying to a notification with a null id was costing those clients the whole tool set, not just the one message. The `is_notification` helper plus the three integration tests are exactly the right shape, and covering `notifications/roots/list_changed` explicitly alongside the wildcard arm is a nice touch.

PR #517, opened 2026-09-08
fix(mcp): deduplicate and suppress index-age warning banners

### LINKED ISSUE #516: MCP tool results repeatedly prepend the one-hour index-age warning and can trigger agent retries
### tokensave version

7.11.0

### Operating system

macOS

### Installation method

Built from source

### What happened?

When an MCP server's `last_sync_at` is more than one hour old, `McpServer::handle_tools_call` inserts `WARNING: Index last synced ... Run `tokensave sync` to update.` as the first text item in every successful local tool result. The outer JSON-RPC response remains valid, but the warning is mixed into the tool's content stream as if it were part of the answer. In agent clients this renders as a banner before the result and can be mistaken for a failed or incomplete query. A query that returns no match is still a successful tool result, so it receives the same banner: `WARNING: ... No symbol named '...' found.` Repeated calls can therefore produce warning spam and agent retry loops.

Expected: the answer payload stays independently parseable, and a diagnostic advisory is not prepended to every successful result. Actual: every successful result on an index older than one hour carries the warning as its first content item, with no deduplication and no structured freshness field.

### Steps to reproduce

1. Initialise and index a small project.
2. Backdate the `last_sync_at` metadata value by more than one hour, or leave an indexed project idle for that long.
3. Start `tokensave serve` and call `tokensave_body` with a symbol that does not exist.

### Error output

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "WARNING: Index last synced 2h 0m ago. Run `tokensave sync` to update."
      },
      {
        "type": "text",
        "text": "No symbol named 'no_such_symbol_anywhere' found."
      }
    ]
  }
}
```

The same ordering puts the warning before a text item containing JSON for tools whose answer is serialized as text. A client that parses the first text item as the answer, rather than handling all MCP content items, can fail or hand warning text to the model as the result.

### Component

MCP server

### Additional context

The age banner was introduced by `acc1bec` ("add staleness detection") and is still present upstream. The warning is attached after the handler returns `Ok(ToolResult)`, at the overall index-age check in `src/mcp/server.rs`. Handler errors take a separate JSON-RPC error path and do not receive this age banner. `tokensave_body` deliberately returns `Ok` with `No symbol named ...` for a zero-hit lookup, so that message is not an MCP error.

Before dispatch, the server already runs bounded automatic freshness work through `maybe_sync_if_stale`. It is gated by a 30-second cooldown and refuses unsafe or expensive cases, including an uninitialised index, too many stale files, and branch drift. Those refusal reasons are currently printed to the server's stderr rather than attached to the tool result. An index can therefore be more than one hour old even when the automatic check found no file changes, was inside its cooldown, or refused the sync. The age banner then repeats on every successful call.

The related upstream history covers timestamp correctness (#86), stale-file response guidance (#428), bounded automatic sync (#401), branch/worktree warnings (#201, #400, #446), and the broader freshness design discussion (#342). None of those changes establishes a response-level policy for repeated age warnings.

This is not a JSON-RPC framing or stdout-pollution bug. The server serializes one valid JSON-RPC response line, and the warning is a legal MCP text content item. It is a response-contract and agent-interaction bug, because a diagnostic advisory is inserted ahead of machine-oriented answer text on every successful call. The behavior is also too coarse for the current automatic-sync model: wall-clock age alone does not mean that the requested graph answer is stale.

Suggested direction: report freshness as structured metadata (for example an MCP `_meta` field with age, check status, and refusal reason) so answer text remains parseable; if human-readable text remains, emit it as a separately identified diagnostic rather than prepending it to the answer content; at minimum, deduplicate or rate-limit the age advisory per server session and do not attach it to zero-hit or otherwise diagnostic-only successful results; and keep an explicit `tokensave sync` recovery path for genuinely stale or refused indexes.

rNoz:
## Summary

- Fixes the response-contract bug where the one-hour index-age advisory (`WARNING: Index last synced ...`) was prepended to the content of every successful MCP tool result, including zero-hit lookups, causing warning spam and agent retry loops.
- Freshness state is now reported as structured `_meta.freshness` metadata on stale results; the human-readable banner is rate-limited to once per hour per server session and suppressed on zero-hit, not-found, and `tokensave_status` results.

## Motivation

Fixes #516.

When an index is more than one hour old, `handle_tools_call` inserted the warning as the first text item of every successful result. The JSON-RPC envelope stayed valid, but the warning sat ahead of the machine-oriented answer text. A zero-hit lookup is a successful result, so it got the banner too: `WARNING: ...` followed by `No symbol named '...' found`. Agent clients that treat the first text item as the answer, or that read the banner as a failure signal, retry the call, and every retry produces the banner again. Wall-clock age alone also does not mean the requested answer is stale: the bounded automatic freshness check may have found no changed files, been inside its cooldown, or refused the sync, and the refusal reasons only went to stderr.

## Changes

- `src/mcp/server.rs`: on stale results, attach `_meta.freshness` with `index_age_secs`, `last_sync_at`, `stale`, and `warning_emitted` so clients can read freshness without parsing prose.
- The text banner is emitted at most once per hour per server session for the local graph, and once per hour per provenance root for selected (`graph_root`) graphs; the timestamp resets when the index is fresh again.
- `is_zero_hit_or_diagnostic` suppresses the banner on not-found answers, empty result sets, and `tokensave_status`, so the answer payload stays independently parseable.
- `should_emit_index_age_warning` is a pure decision function with a truth-table unit test.
- `tests/mcp_server_test.rs`: regression test covering zero-hit suppression, first emission, same-session dedup, and `tokensave_status`, asserting both the content text and the `_meta.freshness` fields.
- `CHANGELOG.md` entry under `[Unreleased]`.

## Test plan

- [x] `cargo test` passes

New regression test `test_index_age_warning_dedup_and_zero_hit_suppression` passes; existing staleness unit tests pass. Manual end-to-end proof against a scratch project with `last_sync_at` backdated by two hours, driving the installed server over stdio JSON-RPC:

| call | banner | `_meta.freshness` |
| --- | --- | --- |
| zero-hit `tokensave_body` | no | `{"stale": true, "warning_emitted": false}`, answer text `No symbol named 'no_such_symbol_anywhere' found.` |
| `tokensave_search` (first) | yes | `{"stale": true, "warning_emitted": true}` |
| `tokensave_search` (repeat) | no | `{"stale": true, "warning_emitted": false}` |
| `tokensave_status` | no | not applicable |

<details>

<summary>Gate output</summary>

```
cargo fmt --all -- --check    # clean
cargo clippy --workspace --all-targets    # no new warnings
cargo test --features test-transport --test mcp_server_test test_index_age_warning_dedup_and_zero_hit_suppression    # 1 passed
cargo test --lib staleness    # 5 passed
```

</details>

## Checklist

No breaking changes: the JSON-RPC envelope is unchanged, and clients that already iterate all content items see the same warning text, only deduplicated. Clients gain the `_meta.freshness` field.


aovestdipaperino:
Merged — thank you @rNoz, both for the report in #516 and for fixing it.

The framing in the issue is what made this straightforward to accept: it was argued as a response-contract problem rather than a formatting complaint, and the point that wall-clock age alone does not mean *the requested answer* is stale is the right one. Structured `_meta.freshness` plus a rate-limited banner is a better shape than either keeping the old behaviour or dropping the advisory entirely, and suppressing it on zero-hit results keeps the answer text independently parseable, which was the actual bug.

The end-to-end table driving the installed server over stdio with a backdated `last_sync_at` was more verification than I'd have asked for. Appreciated.

Issue #513, opened 2026-09-08
upgrade: a release with no asset for the platform is reported as "could not reach GitHub"


xaerogonzo:
## Summary

When the latest release has no asset for the running platform, `tokensave upgrade` fails with:

```
Error: config error: failed to check for updates — could not reach GitHub
```

GitHub is reachable. Nothing is wrong with the user's configuration. The message names the two things that are fine and not the one thing that is wrong, so it sends people to debug a working network.

**This is deliberately not a re-report of #512.** #512 is that v7.11.1's Windows assets are missing, and why — the SignPath signing legs timed out in release run #170, so the signed ZIPs never uploaded. That is a release-workflow problem with a known fix. This issue is that `upgrade` cannot tell a missing asset from an unreachable host, which survives the v7.11.1 re-sign and will misdiagnose the next occurrence. #136 ("v6.4.0 doesn't have Windows assets") shows there is a next occurrence.

## Reproduction

Windows 11, tokensave 7.11.0 (stable), while v7.11.1 is latest and carries no Windows asset:

```
$ tokensave upgrade
Current version: v7.11.0 (stable channel)
Checking for updates...
Error: config error: failed to check for updates — could not reach GitHub
```

Exit code 1, deterministic across repeated invocations.

## Why this is a diagnosis bug and not a network fault

Reachability verified three ways from the same shell, minutes either side of a failing `upgrade`:

- `curl` to `api.github.com/repos/aovestdipaperino/tokensave/releases/latest` — **200**
- PowerShell `Invoke-WebRequest` (schannel, a different TLS stack from the one curl uses here) to both `api.github.com/.../releases/latest` and `github.com/.../releases/latest` — **200** each
- `gh release list` — returns the full release list, including v7.11.1

Explicitly ruled out:

| Suspect | Observation |
|---|---|
| Rate limiting | `api.github.com/rate_limit` reported `core` 55/60 remaining during a failing run |
| Proxy | No `*_PROXY` variables in the environment; `ProxyEnable = 0` in Internet Settings |
| Firewall | No `Get-NetFirewallApplicationFilter` rule matches the binary, in either direction |
| IPv6 | No AAAA record for `api.github.com` here, so not a v6 route timing out before fallback |
| Authentication | Re-running with a valid `GITHUB_TOKEN` and `GH_TOKEN` in the environment changes nothing |

The tool's own state agrees that its checks are failing rather than that it is current. `~/.tokensave/state.toml`:

```toml
[...]

PR #526, opened 2026-09-10
fix(upgrade): trust OS certificate store for HTTPS calls

gioxx:
## Why

Running tokensave behind a corporate proxy that does TLS inspection (e.g. Cato), `tokensave upgrade` fails with:

    Error: config error: failed to check for updates - could not reach GitHub

The network itself is fine: curl, the browser and git all work through the same proxy. Only tokensave's HTTPS calls fail.

The cause: ureq (the HTTP client tokensave uses) validates TLS certificates against a bundled Mozilla root store instead of the operating system's trust store. A corporate proxy that inspects TLS installs its own root CA into the OS trust store, and every OS-native client (browsers, curl, git) trusts it there. ureq's bundled store never sees that CA, so the handshake fails and gets reported as a generic connectivity problem, sending people to debug a network connection that was never broken.

## What changed

Every `ureq::Agent` tokensave builds now uses `RootCerts::PlatformVerifier` instead of the default `WebPki` root set. This makes tokensave validate certificates against the OS trust store (Windows CryptoAPI/schannel, macOS/iOS Keychain, Linux via the OS loaded store), the same approach curl and git for Windows already use by default on Windows. TLS verification itself is unchanged: only the set of trusted root certificates changes, and it only grows to include what the OS already trusts. No new CLI flag or environment variable, this becomes the default for everyone.

Also cleaned up: `src/upgrade.rs` had two places building their own `ureq::Agent` with duplicated config. Both now go through the existing `cloud::agent_with_timeout`, so there is a single place where HTTP client behavior is configured.

## Testing

- CI green on Linux, macOS and Windows (test, clippy, fmt)
- Added a lightweight unit test that constructs the agent, since a provider/feature mismatch would otherwise only surface as a runtime panic on a user's first HTTPS call
- CHANGELOG.md updated under [Unreleased]
- Manual verification behind an actual TLS inspecting proxy still pending, will confirm before merge


aovestdipaperino:
Grazie!

gioxx:
Sempre a disposizione, Enzino! 🫡 😄 ❤️

aovestdipaperino:
Merged — thank you @gioxx. The diagnosis was the valuable part: a TLS trust failure reported as `could not reach GitHub` sends people to debug a working network, which is the same class of misdirection #513 was about, and it would not have been obvious without someone sitting behind an actual intercepting proxy.

I resolved the CHANGELOG conflict on your branch (master had moved) and corrected one claim in the entry before merging, because it matters for what users should expect:

**`RootCerts` is an enum, so this replaces the bundled roots rather than adding to them.** `PlatformVerifier` and `WebPki` are mutually exclusive variants (`ureq-3.4.1/src/tls/mod.rs:294`), so the OS store is now the only source of trust anchors, not an additional one. The practical consequence is on Linux, in `rustls-platform-verifier`:

```rust
if root_store.is_empty() {
    return Err(rustls::Error::General(
        "No CA certificates were loaded from the system".to_owned(),
    ));
}
```

There is no fallback. A host with no system CA store — a distroless or `scratch` container without `ca-certificates` — loses HTTPS where the bundled roots previously carried it. Installing `ca-certificates` restores it, and the CHANGELOG now says so.

Merging anyway: the proxied case has no workaround today and the container case has an easy one. Worth knowing if issues come in from CI images after the next release.

The agent-construction test was a good call — a provider/feature mismatch panicking on a user's first HTTPS call is exactly the kind of thing that would have shipped otherwise. Consolidating both agents onto `cloud::agent_with_timeout` was a welcome cleanup too.

aovestdipaperino:
Follow-up merged in #534: the no-CA-store regression this introduced now has a fallback, so the caveat I left on this PR is closed.

The trust source is chosen rather than assumed. tokensave probes the store first — using `rustls-native-certs`, which is the loader `rustls-platform-verifier` itself uses on Linux/BSD, so the probe answers the exact question the verifier is about to ask and honours `SSL_CERT_FILE`/`SSL_CERT_DIR` identically — and keeps the bundled roots when it comes back empty. Probed once per process, skipped entirely on macOS and Windows where the failure mode does not exist, with the dependency target-gated so those builds do not pull it.

Your fix is unchanged wherever the OS has a store, which is everywhere it was working. A host without `ca-certificates` now stays on the bundled roots instead of losing HTTPS, and warns once so someone behind a proxy can still see why an OS-installed CA is being ignored.

The fallback only ever narrows trust — the bundled set is a fixed audited list, never "no verification" — and there is a test asserting that specifically, since getting it backwards would trade an outage for a silently unverified connection.

Thanks again for the original diagnosis. The TLS-interception case is not something that shows up without someone sitting behind one.

Issue #536, opened 2026-09-11
Phantom edges landing inside production are invisible to the impossible-edge metric: #533 removed 1,216, the metric counted 731


xaerogonzo:
## Summary

On #522 you recorded the limit yourself:

> the impossible-edge count is a lower bound since it only catches crossings into
> `tests/`, never phantoms landing inside production.

Measuring #533 against its own parent puts a number on it: the metric saw **731**
of at least **1,216** phantoms that change removed. The other **485** were the
identical defect landing production → production, where crossings into `tests/`
cannot see them.

This matters because of what comes next. You wrote that tightening the import rule
"needs measuring the same way this did". Measured that way it will undercount its
own benefit by roughly the same factor — and #533 would read as 60% precision when
the evidence says it is close to 100%.

I also have a **negative** result below: the obvious better predicate does not work,
and I think that is worth recording before someone measures with it.

## Measurement

`18452e8` and its parent `97f32fe`, both `cargo build --release`, each running
`init` on its own pristine clone of the same tree at the same commit.
`src/resolution/resolver.rs` is byte-identical between `18452e8` and the merge
commit `d795026`. Python project, 515 files, 16,558 nodes. Node-id sets are
identical across both indexes, so edges diff exactly; 0 edges added.

| | baseline | shipped | delta |
|---|---|---|---|
| total edges | 22,670 | 21,454 | **-1,216** |
| impossible (production → `tests/`) | 782 | 51 | **-731** |

1,216 removed, 731 visible to the metric. **485 unaccounted for**, which by the
metric is collateral — 60.1% precision.

## The 485 are phantoms, not collateral

Every one of them:

* targets the **sole** symbol of that name in the index — 485/485
* **crosses files** — 485/485
* in a tree with **zero** `import *`, the evidence test's only real Python blind
  spot, so a reference that fails it genuinely cannot reach the target

Verified in source on the largest four, 299 of the 485:

| edges | target | what it actually is |
|---|---|---|
| 218 | `src/controllers/help_tab.py:478::p` | a closure `def p(s)` nested inside `_hw()`, a text-writer helper |
| 36 | `src/helpers/instructions_posture.py:741::total` | a `@property` on a dataclass |
| 29 | `src/helpers/instructions_wiring.py:102::files` | a `@property` on a dataclass |
| 16 | `src/helpers/geometry.py:63::right` | a `@property` returning `self.x + self.w` |

[...]

Issue #522, opened 2026-09-08
Python extractor: the bare-name fallback still binds `uses` edges without evidence (the #503 defect, one edge kind over)


xaerogonzo:
## Summary

7.11.1 (#508) taught the **call** resolver that being the only candidate is
not evidence. The **reference** resolver was not taught the same thing.

A production function that declares a local variable named `exe` acquires a
`uses` edge pointing at a **pytest fixture** named `exe` in the test tree —
because that fixture is the only symbol of that name in the project. The
remedy #508 applied to calls (require the candidate to be in the caller's
file, its directory, or something the caller imports) is exactly the remedy
this needs, and none of it is applied here.

The result is the same class of impossible edge #503 described, surviving in
a graph that #508 otherwise cleaned up by 94%.

## Measured, before and after 7.11.1

One Python project, 489 files, 14,422 nodes. Full `sync --force` under
7.11.0, then the identical tree fully re-indexed under 7.11.1. Counting
edges from production files into the test tree — impossible by construction,
since tests import production code and never the reverse:

| edge kind | 7.11.0 | 7.11.1 | examined (7.11.1) |
|---|---|---|---|
| `calls`   | 455 | **29** | 10,656 |
| `uses`    | 351 | **351** | 7,277 |
| `annotates` | 0 | 0 | 531 |
| `extends` | 0 | 0 | 45 |

`calls` fell 94%. `uses` did not move by a single edge.

The per-name counts show the same split, and are their own proof that the
two populations are disjoint. Names that are **method** names moved:
`after` 168 → 20, `askyesno` 77 → 0, `delete` 55 → 0, `grid` 24 → 6,
`winfo_width` 34 → 7. Names that are **variable / fixture** names did not
move at all: `exe` 109 → 109, `node` 55 → 55, `lines` 46 → 46,
`cfg_path` 40 → 40, `proj` 11 → 11, `python_exe` 11 → 11, `npm_exe` 9 → 9.

## Reproduction

The target — a pytest fixture, and the only symbol named `exe` anywhere in
the project:

```python
# tests/test_<something>.py:42
@pytest.fixture
def exe(tmp_path):
    p = tmp_path / "pyscope.exe"
    p.write_text("", encoding="utf-8")
    return str(p)
```

A source — an ordinary local variable in production code, in a different
directory, in a file that imports nothing from the test tree:

```python
# src/cli.py:146
    exe = (cfg.get("tokensave_exe") or "").strip()
    if not exe:
[...]

Issue #503, opened 2026-09-04
Python extractor: unqualified calls on untracked receivers still fall back to bare-name resolution (the #378 defect, in Python)


xaerogonzo:
## Summary

When a Python call's receiver is a type the index does not track — a stdlib
object, a third-party instance, a `logging.Logger` — the resolver falls back
to matching the bare method name against every indexed symbol. If exactly one
symbol in the whole project carries that name, the call is bound to it,
**regardless of which module or directory it lives in**.

This is the defect reported for the TypeScript extractor in #378 and closed in
August. The Python extractor exhibits it in 7.11.0.

The failure is quiet and it is directional: production code binds to helpers
in the test tree, because test doubles are deliberately named after the API
they stand in for. A faithful fake of a UI toolkit defines `after`, `delete`,
`grid` and `winfo_width`; a fake logger defines `info`. Those are exactly the
names production code calls on untracked receivers.

## Reproduction

A whole project, three files. No configuration.

`src/a.py` — production code. Calls a logger. Defines no `info` of its own:

```python
import logging

log = logging.getLogger(__name__)


def do_work(payload):
    log.info("starting %s", payload)
    return payload
```

`tests/helper.py` — a test double. Never imported by `src/`:

```python
def info(msg, *args):
    return (msg, args)
```

Then:

```bash
git init . && git add -A && git commit -m init
tokensave init
```

### Observed — one candidate

The graph contains exactly one `calls` edge, and it is wrong:

```
src/a.py:7   do_work --calls--> info   (tests/helper.py)
ambiguous_calls: 0 rows
```

`do_work` does not call anything in `tests/`. It cannot: nothing in `src/`
imports that module. The edge is the only edge in the graph.

### Observed — two candidates

Add a second file that changes nothing about `src/`:

`tests/other.py`

```python
def info(msg, *args):
    return None
```

Re-index from scratch. The resolver now declines, correctly:

```
calls edges:     (none)
ambiguous_calls: log.info at src/a.py:7  -> 2 candidates
```

**This contrast is the report.** The resolver already has a "not confident
enough" path and already routes this exact call site into it. It just does not
take that path when the ambiguity set happens to have one member. One
candidate is not evidence of a correct match when nothing checked that the
candidate is reachable from the call site.

## Secondary finding: `sync` leaves the superseded edge behind

Found while building the contrast above, and reproducible with the same files.

Running `tokensave sync` after adding `tests/other.py` — rather than indexing
[...]

PR #537, opened 2026-09-12
test(resolution): pin the bare-name gate in the production → production direction (#536)

xaerogonzo:
Test-only. No production code, no behaviour change — #533 already handles this
case correctly. What is missing is coverage.

## What is uncovered

#533 removed **1,216** edges on a 515-file Python project. Only **731** crossed
into `tests/`; the other **485** ran production → production, where a
production → `tests/` metric cannot see them. The largest single target was a
closure `def p(s)` nested inside a method, which 218 sites acquired a `uses` edge
to purely because it was the only symbol named `p` in the project. Measured in
#536.

Both tests written for this defect class build the phantom as `src/` → `tests/`:
`python_phantom_call_edges_test` (#503) uses a `logging.Logger` and a test double
defining `info`; `bare_name_reachability_test` (#522) uses a pytest fixture named
`exe`. Neither puts both ends in production.

The collision that generates these is not specific to test trees — `p`, `total`,
`files`, `right` are what a nested helper gets called anywhere.

## That it is uncovered is measured, not asserted

`97f32fe` (the parent of #533) is green across **all 171 test targets**, and the
phantom test below **fails** on that same commit. If anything existing pinned this
direction, that baseline could not have been green.

```
97f32fe  a_python_bare_name_does_not_bind_to_an_unreachable_production_candidate ... FAILED
         a_python_bare_name_still_binds_across_production_modules_when_imported ... ok
master   6 passed; 0 failed
```

## The two cases

`a_python_bare_name_does_not_bind_to_an_unreachable_production_candidate` — the
measured shape with neither end in a test tree: a closure named `p` nested inside a
method is the sole symbol of that name, and an ordinary local `p` sits in a
different production package importing nothing from it.

`a_python_bare_name_still_binds_across_production_modules_when_imported` — the
recall control. Without it the first test would also be satisfied by a rule
refusing every cross-package bare name, which would be a worse defect than the one
being guarded. This is the same before/after property the existing three controls
have, and the reason they are there.

Appended to `bare_name_reachability_test.rs` rather than added as a new target, so
they reuse its `has_edge` helper verbatim. 61 insertions, 0 deletions.

## Gates

* `cargo fmt --all -- --check` — clean
* `cargo test --test bare_name_reachability_test` — 6 passed, 0 failed
* `cargo clippy --workspace --all-targets` — exit 0, **no new warnings**. Master
  currently carries 12, all in `src/` (`upgrade.rs`, `tokensave/guard.rs`,
  `resolution/variants.rs`, `mcp/transport.rs`, `db/queries/edges.rs`,
  `agents/fs.rs`); none reference the file this touches, which is the only file
  changed.

Built and run on Windows 11, `rustc 1.98.1` MSVC — so this is also a second
platform for #533's suite, which came back 172 targets / 3,747 tests / 0 failures.

Closes nothing on its own; it is the coverage half of #536.

### DIFF
--- a/tests/bare_name_reachability_test.rs
+++ b/tests/bare_name_reachability_test.rs
@@ -150,3 +150,64 @@ async fn a_rust_bare_name_across_modules_is_not_gated() {
         "Rust bare-name resolution must be unaffected by the gate"
     );
 }
+
+/// The measured shape, with neither end in a test tree: a closure named `p`
+/// nested inside a method is the only symbol of that name in the project, and
+/// an ordinary local named `p` sits in a different production package that
+/// imports nothing from it.
+#[tokio::test]
+async fn a_python_bare_name_does_not_bind_to_an_unreachable_production_candidate() {
+    let tmp = tempdir().unwrap();
+    let root = tmp.path();
+    std::fs::create_dir_all(root.join("ui")).unwrap();
+    std::fs::create_dir_all(root.join("util")).unwrap();
+
+    std::fs::write(
+        root.join("ui/panel.py"),
+        "import tkinter as tk\n\n\nclass Panel:\n    def writers(self):\n        t = self._txt\n\n        def p(s):\n            t.insert(tk.END, s + \"\n\")\n\n        return p\n",
+    )
+    .unwrap();
+    std::fs::write(
+        root.join("util/paths.py"),
+        "import os\n\n\ndef cache_dir(root):\n    p = os.path.join(root, \".cache\")\n    os.makedirs(p, exist_ok=True)\n    return p\n",
+    )
+    .unwrap();
+
+    let cg = TokenSave::init(root).await.unwrap();
+    cg.sync().await.unwrap();
+
+    assert!(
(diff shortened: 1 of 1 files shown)


aovestdipaperino:
Thanks @xaerogonzo — merging. The before/after baseline against `97f32fe` is what makes this worth taking: a test that fails on the parent and passes on master is coverage, not restatement, and the recall control is the right companion.

One follow-up I'll push myself rather than bounce back to you: in `a_python_bare_name_does_not_bind_to_an_unreachable_production_candidate`, the Rust string writes `s + \"\n\"` with an unescaped newline, so `ui/panel.py` is not valid Python (`SyntaxError: unterminated string literal`). I checked and the test is *not* vacuous today — the closure `p` is still extracted from the recovered tree and is still the sole candidate, so the gate is genuinely what the assertion is pinning. But it's load-bearing on error recovery by accident, and a future parser change could quietly turn it into a pass-for-the-wrong-reason. Escaping it to `\\n` on master.

PR #548, opened 2026-09-12
feat(config): explain environment override precedence

### LINKED ISSUE #546: Explain TOKENSAVE environment overrides in generated config
## Context

Tokensave generates `.tokensave/config.json`, while several operational settings can also be overridden by `TOKENSAVE_*` environment variables. The current JSON contains no local explanation, so a user can see `"auto_track": false` while `TOKENSAVE_AUTO_TRACK=1` is the effective setting.

## Gap

Real JSON does not support comments, so adding `//` text would make the generated file invalid. The precedence is implemented and tested, but users discover it only through documentation or source inspection.

## Proposal

Add a harmless `_comment` metadata field to generated JSON explaining that `TOKENSAVE_*` environment variables override matching config values when set. Document the same rule in the user guide and preserve forward-compatible loading of existing and newly annotated files.

## Acceptance

Generated config remains valid JSON and loads unchanged, the note names the override convention and the important variables, and existing configs without the metadata field continue to load.

rNoz:
## Summary

- Resolves #546.
- Explains environment-variable precedence directly in generated `.tokensave/config.json`.
- Documents the same rule in the user guide without pretending JSON supports comments.

## Motivation

`TOKENSAVE_AUTO_TRACK=1` can make automatic branch tracking effective even when a project file says `"auto_track": false`. The behavior is correct and already tested, but the generated file does not tell the operator why the visible value is not authoritative.

## Changes

- `src/config.rs`: add a `_comment` metadata field when serializing generated config; unknown metadata remains ignored when loading.
- `tests/config_test.rs`: verify the note is emitted and the annotated JSON still deserializes.
- `docs/USER-GUIDE.md`: document the precedence and name the main override variables.

## Test plan

- [x] `cargo test --test config_test` (42 passed)

## Checklist

- [x] Breaking changes documented (not applicable)
- [x] Existing config files remain readable
- [x] No new dependency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), not SemVer.
 
+## [Unreleased]
+
+### Changed
+- **Generated project configuration now explains environment-variable precedence.** A JSON-safe `_comment` field tells users that matching `TOKENSAVE_*` variables override values in `.tokensave/config.json`.
+
 ## [7.12.1] - 2026-09-12
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


rNoz:
Reviewed the whitespace and syntax concerns in this change. The section relocation now uses a single blank line before and after the heading, the Rust serialization change is rustfmt-formatted, and the generated JSON remains valid. The earlier apparent separator/spacing issue came from the moved Markdown section in the comparison view, not from an extra publication separator.

aovestdipaperino:

PR #550, opened 2026-09-12
fix(branch): coordinate automatic and explicit tracking

### LINKED ISSUE #549: Explicit branch add can race automatic post-checkout tracking
### tokensave version

7.11.1

### Operating system

All supported platforms

### Installation method

Built from source or installed from a release binary

### What happened?

With `TOKENSAVE_AUTO_TRACK=1`, the post-checkout hook starts `tokensave branch add --if-enabled` asynchronously. Immediately running an explicit `tokensave branch add` for the same newly checked-out branch can race that background operation and fail with `sync lock: another sync is already in progress`, even though automatic tracking is already completing successfully.

An explicit branch add can overlap the asynchronous automatic operation for the same branch and surface a transient lock error. This is a duplicate invocation of an expected idempotent operation, not evidence that automatic tracking is disabled.

### Steps to reproduce

1. Enable automatic tracking with `TOKENSAVE_AUTO_TRACK=1`.
2. Check out a new branch in an initialized project.
3. Immediately run `tokensave branch add --if-enabled` or `tokensave branch add` manually.
4. Observe the explicit command while the post-checkout background sync is still copying or indexing the branch database.

### Error output

```text
Error: branch operation lock: another branch add is already in progress
```

### Component

Branch tracking and post-checkout hook coordination

### Additional context

The normal workflow should be either automatic tracking or explicit tracking, not both back-to-back. However, an idempotent explicit command should detect that the same branch is already being tracked or wait for the in-progress operation rather than surface a transient lock error. Candidate fixes include a short bounded wait/recheck for the branch metadata, a dedicated branch-add operation lock, or making the hook hand off serialized work without blocking checkout.

The prepared fix adds a bounded, stale-PID-aware `branch-add.lock` around copy and metadata updates, used by both the CLI and transparent auto-tracking. Repeated branch adds become idempotent, actual sync continues to use the existing `sync.lock`, and focused branch/lock tests cover the coordination behavior.

rNoz:
## Summary

- Resolves #549.
- Resolves the race between asynchronous post-checkout auto-tracking and explicit `branch add`.
- Serializes branch database copy and metadata updates without weakening the existing sync lock.

## Motivation

The post-checkout hook deliberately runs `tokensave branch add --if-enabled` in the background so checkout remains non-blocking. An agent or script can immediately issue the same operation and reach the copy/sync path concurrently. The existing `sync.lock` protects indexing, but not the earlier branch-copy and metadata phase, so the duplicate operation can surface a transient lock error.

## Changes

- Add a bounded, stale-PID-aware `branch-add.lock` for branch copy and metadata updates.
- Use the lock from both the CLI branch-add path and transparent `TokenSave::open` auto-tracking.
- Recheck metadata after acquiring the lock so a duplicate operation becomes an idempotent no-op.
- Report branch-operation lock failures separately from sync-lock failures.
- Keep actual sync protected by the existing `sync.lock`; unrelated sync contention is not hidden.
- Add tests for waiting, timeout, and repeated branch add while keeping the existing auto-track tests green.

## Test plan

- [x] `cargo test --lib branch::tests` (7 passed)
- [x] `cargo test --lib guard::tests` (2 passed)
- [x] `TOKENSAVE_AUTO_TRACK` unset, `GIT_CONFIG_GLOBAL=/dev/null cargo test --test branch_add_if_enabled_test` (5 passed)
- [x] `cargo clippy --workspace --all-targets`

## Checklist

- [x] Breaking changes documented (not applicable)
- [x] No new dependency
- [x] Existing sync-lock behavior preserved
- [x] Lock timeout is bounded and explicit
- [x] Repeated branch tracking remains idempotent


aovestdipaperino:

Issue #524, opened 2026-09-09
fix(db): get_inheritance_depth interpolates path prefix into SQL


aovestdipaperino:
## Summary

`Database::get_inheritance_depth` (src/db/queries/edges.rs, around line 992) builds its filter with string interpolation:

```rust
Some(prefix) => format!("WHERE n.file_path LIKE '{prefix}%'"),
```

`prefix` is the `path` argument of `tokensave_inheritance_depth` (via `effective_path`), passed through unescaped. Every sibling query (`get_nodes_filtered`, `get_ambiguous_calls`, `NodeFilter`) uses either a bound `?N` parameter or `push_quoted` with `LIKE` escaping, so this is an inconsistency rather than a design choice.

## Impact

Low. The MCP caller is the user's own agent and the target is the user's own index DB, so this is not a trust-boundary crossing. But a path containing a single quote, `%` or `_` yields a failed query or wrong matches, and the write connection could be driven into unintended statements.

## Fix

Bind the prefix as a parameter (`file_path LIKE ?1` with `format!("{prefix}%")` as the value), or reuse the `push_quoted` + `LIKE` escape pattern from `get_ambiguous_calls`. Add a regression test with a prefix containing `'`, `%` and `_`.

Found during a security pass on the edit/query tool surface.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

vianbas:
I reproduced this on `9ebe7ec` (7.11.1) and tried both fixes the issue suggests. Two findings. I am not claiming the issue.

**1. The same interpolation is in three functions, not one.** All in `src/db/queries/edges.rs`:

| line | function | tool |
|---|---|---|
| 935 | `get_file_coupling` | `coupling` |
| 992 | `get_inheritance_depth` | `inheritance_depth` |
| 1391 | `get_god_classes` | `god_class` |

All three give the same results on small fixtures with directories `src/my_mod/`, `src/myXmod/` and `src/it's/`, two runs each:

| `--path` | result |
|---|---|
| `src/myXmod` (control) | correct |
| `src/my_mod` | also returns `src/myXmod` |
| `src/%` | returns every directory |
| `src/it's` | SQLite syntax error |

**2. Only the second fix handles all three cases.**

[...]

PR #555, opened 2026-09-13
fix(db): escape the path filter in the remaining LIKE queries

vianbas:
## Summary

- Escape the path filter in the ten queries #524 left out, by routing them through the same `push_path_prefix_filter` the #524 fix introduced.
- Before this, nine MCP tools returned a sibling directory for a path containing `_`, and every directory for `src/%`. That also happened with no `path` argument, through the directory the server was launched from.

## Motivation

Follow-up to [your note on #524](https://github.com/aovestdipaperino/tokensave/issues/524#issuecomment-5638372903) that the remaining `LIKE ?N` sites taking a user path are "a real and separate gap" that "deserves its own issue". This is that gap, with the fix attached.

On master (`68086de`), in a tree with `src/my_mod`, `src/myXmod` and `src/it's`, two runs each through `tokensave tool` (`annotations` takes the path as `file`, `port_status` and `port_order` as `source_dir` and `target_dir`):

| path | `rank`, `largest`, `distribution`, `recursion`, `complexity`, `doc_coverage`, `annotations`, `port_status`, `port_order` |
|---|---|
| `src/myXmod` (control) | only `src/myXmod` |
| `src/my_mod` | also `src/myXmod` |
| `src/%` | every directory |
| `src/my` | `src/my_mod` and `src/myXmod` |
| `src/it's` | correct, since these queries bind the value |

With no `path`, these tools fall back to the directory the server was launched from. A `serve` started in `svc/user_api` answered `largest` with `svc/userXapi` too.

## Changes

- `src/db/queries/edges.rs` and `metadata.rs`: `get_annotation_histogram`, `get_annotation_sites`, `get_ranked_nodes_by_edge_kind`, `get_largest_nodes`, `get_node_distribution`, `get_call_edges`, `get_call_edges_with_lines`, `get_complexity_ranked`, `get_undocumented_public_symbols` and `get_nodes_by_dir` now use `push_path_prefix_filter`. Bound parameters after the removed one are renumbered. `get_call_edges` has no caller outside the tests; I changed it anyway so no unescaped copy is left.
- **Behaviour change**, the same one #524 made for its three queries: the filter matches a file or a directory, not a string prefix, so `src/my` no longer matches `src/my_mod`. One existing test relied on the old behaviour. `test_get_undocumented_public_symbols_with_prefix` passed `src/utils` to mean the file `src/utils.rs`; after this change that returns nothing, and the test would still have passed because it only checked each returned row. It now names the file and asserts `helper` is returned.
- `src/mcp/tools/definitions.rs`: the six parameter descriptions that said "prefix" now say file or directory.
- `tests/path_prefix_like_escape_test.rs`: the four existing cases now run against twelve of the thirteen path-filtered queries. The thirteenth, the annotation histogram, returns counts rather than paths, so it gets its own count-based test. Also added a partial-name test and a launch-directory test through `handle_tool_call` with a scope prefix.
- `CHANGELOG.md` under `[Unreleased]`.

Left out on purpose: `get_nodes_by_qualified_name` (`%::{qname}`) and the search fallback (`%{query}%`) also leave `_` live, but they match names rather than paths, so they belong to a different change.

## Test plan

All on macOS arm64, rustc 1.95.0 (the repository's `rust-toolchain.toml`), default features.

- [x] `cargo test --workspace`: 3,868 passed, 0 failed, same result on two runs (176 test binaries). Master: 3,865 passed, 0 failed, the same under `cargo test` and `cargo test --workspace`.
- [x] `cargo clippy --workspace --all-targets`: no new warnings. Master and this branch both report the same 5 warnings (two unknown-lint warnings in `src/mcp/transport.rs`, three missing-backtick doc warnings); two of them sit 5 lines higher in `edges.rs` because the file got shorter.
- [x] The new tests against unmodified master: 5 of the 7 tests in `path_prefix_like_escape_test.rs` fail (`_`, `%`, the histogram count, the partial name, the launch directory), and the control and apostrophe cases pass.
- [x] Tested manually: the table above re-run with a build of this branch, and the `serve` launch-directory case, two runs each: for `src/myXmod`, `src/my_mod` and `src/it's` every tool returned only that directory, and for `src/%` and `src/my` it returned nothing (histogram counts 4, 4, 4, 0, 0). The server launched in `svc/user_api` answered only with `svc/user_api`. Both runs were identical.

## Checklist

- [x] Breaking changes documented: the file-or-directory matching described above

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project uses [maintenance-based versioning](TOKENSAVE-VERSIONING.md), not SemVer.
 
+## [Unreleased]
+
+### Fixed
+- **A path filter no longer treats `_` and `%` as wildcards in the ten queries #524 did not reach.** #524 fixed three queries that interpolated the caller's path into `LIKE '{prefix}%'`. Ten more bound it as a parameter, `format!("{prefix}%")`, which is why an apostrophe never failed there, but binding does not change what `_` and `%` mean inside `LIKE`: `--path src/my_mod` also returned `src/myXmod`, and `--path 'src/%'` returned every directory. The tools that reach them are `rank`, `largest`, `distribution`, `recursion`, `complexity`, `doc_coverage`, `annotations` (its `file` filter, in both modes), `port_status` and `port_order`; the tenth query, `get_call_edges`, has no caller outside the tests. It needed no path from the caller either: with no `path` argument these tools scope to the directory the server was launched from, so a server started inside `svc/user_api` also answered with `svc/userXapi`. All ten now share `push_path_prefix_filter` with the three #524 fixed, so they also match the directory itself plus everything under it rather than any path merely starting with those characters: `--path src/my` no longer matches `src/my_mod`, and a single file is named in full. The tool descriptions that said "prefix" now say file or directory.
+
 ## [7.12.1] - 2026-09-12
 
 ### Fixed
--- a/src/db/queries/edges.rs
+++ b/src/db/queries/edges.rs
@@ -459,8 +459,8 @@ impl Database {
 
     /// Histogram of annotation / attribute / decorator usages across the
     /// project. Each row is `(annotation_name, count)` sorted descending by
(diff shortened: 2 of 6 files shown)


aovestdipaperino:

aovestdipaperino:

aovestdipaperino:

PR #574, opened 2026-09-20
fix(hooks): resolve relative grep targets against the event cwd

cezkid:
## Summary

- A relative grep/find target in a `PreToolUse` payload was resolved against the **hook process's** working directory instead of the event's `cwd`, so the same payload was graded two different ways depending on where the harness happened to spawn the hook.
- `HookEnv` now carries the event `cwd`, and every relative spelling - the target, the leading `cd` operand, and the config-exclude check - resolves against it.
- 3 regression tests added; the payload in them uses the hook process's own cwd as a decoy so a reverted fix fails.

## Motivation

`target_looks_like_code` handed the raw target straight to `dir_holds_code_files`, which resolves a relative path against the process cwd. The harness spawns the hook wherever it sits, which is not where the session is, so this payload:

```json
{"cwd": "<indexed project>", "tool_input": {"command": "grep -rn handle_request mypkg/"}}
```

was denied when the hook happened to run inside the project and allowed when it ran anywhere else. Whether the grep hook fired at all depended on a directory the payload does not mention.

Absolute targets were unaffected. Every relative operand was in the blast radius, including the `cd X && grep ...` form, whose target belongs to `X` rather than to either working directory - there `cd_path` was classified against the project root and `cd_base` was joined onto it, so a session in a subdirectory re-rooted its own `cd` onto the wrong base.

## Changes

`src/hooks.rs`:

- `HookEnv::cwd: Option<PathBuf>` plus `HookEnv::base()`, which is `cwd` falling back to `project_root`. Both constructors already received the event cwd; they now keep it.
- New `resolve_against_base(raw, base, home)` - `~` expanded, absolute taken as it stands, relative joined onto `base`. It replaces the two open-coded `root.join(target)` sites in `classify_path_containment_with_home` and `path_is_config_excluded_with`; both now take `base` explicitly so the tests stay hermetic.
- `dir_holds_code_files` takes `&Path` instead of `&str`, and `target_looks_like_code` carries the resolved `Option<PathBuf>` (`inside_dir`) rather than a bare `known_inside` flag - the resolution is done once, where the base is known, instead of being re-derived from a string later.
- The two `cd`-prefix arms (`evaluate_bash_segment`, `evaluate_find_command`) classify and join against `env.base()`.

Behaviour deliberately unchanged: a target that resolves outside the indexed tree is still not this guardrail's concern, and an unresolvable one still falls back to the extension/basename rules (`Containment::Unknown`, fail-open).

## Test plan

- [x] `cargo test --workspace --locked` - 3805 tests across 176 suites, 0 failures
- [x] `cargo clippy --all-targets` - no new warnings (the ones it prints are pre-existing and in other files)
- [x] 3 new tests in `tests/hooks_test.rs`: relative target against the event cwd, against a subdirectory event cwd, and a `../` target climbing out of the tree left alone
- [x] The 8 existing hook test files that build a `HookEnv` literal set `cwd: None`, so they keep exercising the `project_root` fallback

## Checklist

- [x] No breaking changes - `HookEnv` gains a field, and every in-tree constructor fills it


aovestdipaperino:
Merged — thank you, @cezkid.

The diagnosis is the valuable part here. A guardrail whose verdict depends on a directory the payload never mentions is worse than one that is simply too strict, because it is not reproducible from the input: the same call was denied or allowed depending on where the harness happened to spawn the hook. Collapsing that to one notion of where the call ran is the right fix, and carrying the resolved `PathBuf` through as `inside_dir` rather than re-deriving it from a string later is a genuine tidy-up on top.

Using the hook process's own cwd as a decoy in the regression tests is a nice touch — a reverted fix fails rather than passing by coincidence.

One small residue I noticed and did not hold the merge for: in both `cd` arms, the `canonicalize()` failure fallback still reads `root.join(&cd_path)` rather than the new base, so a `cd` to a directory that does not exist yet falls back to the old behaviour. Only reachable for an unresolvable path, where the result is `Unknown` and fail-open anyway, so nothing misbehaves — but if you are back in this file, that is the last spot where the two bases disagree.

I rebased onto master to resolve a CHANGELOG collision with #581, which landed first and added its entry in the same place. Both entries are kept. Full CI green after the merge.

PR #578, opened 2026-09-20
fix(opencode): support OpenCode v2 mcp.servers configuration format

jondmarien:
## Summary
OpenCode 2.0 has been released and reorganizes MCP server configurations from the top-level `mcp.<name>` (v1) under `mcp.servers.<name>` (v2).

Currently, `tokensave doctor`, `install`, and `uninstall` only inspect `/mcp/tokensave`. On systems using OpenCode v2 configuration, `doctor` falsely reports:
```
✘ MCP server NOT registered in ~/.config/opencode/opencode.json — run tokensave install --agent opencode
```
even when `mcp.servers.tokensave` is correctly configured.

## Changes
- **has_tokensave**: Check `/mcp/servers/tokensave` in addition to legacy `/mcp/tokensave`.
- **install_mcp_server**: Preserve existing command from either `/mcp/servers/tokensave/command` or `/mcp/tokensave/command`, and write under `mcp.servers` when that table exists.
- **uninstall_mcp_server**: Support removing tokensave from both `mcp.servers` and top-level `mcp`.
- **doctor_check_config**: Accept tokensave registered under either `/mcp/servers/tokensave` or `/mcp/tokensave`.
- **tests**: Add test `test_opencode_v2_servers_install_and_healthcheck` in `tests/opencode_agent_test.rs`.

Co-Authored-By: Warp <agent@warp.dev>


aovestdipaperino:
Merged — thank you, @jondmarien. Full CI green.

Good catch on the false negative in particular. `doctor` telling someone to run `install` when their config is already correct is the worst shape of wrong answer: it sends them to fix something that is not broken. Handling all four call sites (`has_tokensave`, `install`, `uninstall`, `doctor`) in one pass rather than just the one that surfaced the symptom is what made this straightforward to take.

Two things I noticed while reading, neither worth holding the merge for — both are follow-ups if you are still in this file, since you are running v2 and I am not:

1. **Fresh installs still write the v1 shape.** `install_mcp_server` writes under `mcp.servers` only when that table already exists, so a v2 user with no `opencode.json` (or one with no `mcp` table yet) gets `mcp.tokensave`. That still works today given the reader accepts both, but it means we keep seeding the legacy shape on new machines indefinitely. If OpenCode exposes a version we can cheaply detect, preferring `mcp.servers` on a fresh write would stop that.

2. **Uninstall can leave `"mcp": {"servers": {}}` behind.** The prune check asks whether `mcp` is empty, and after a v2 removal it still holds an empty `servers` table, so the cleanup does not fire. Cosmetic, but it leaves a fragment in the user's config that nothing owns.

Thanks also for the test — exercising install → healthcheck → uninstall against a pre-populated v2 config, and asserting the unrelated `other` server survives both, is exactly the shape that catches a regression here.

Issue #569, opened 2026-09-17
post-checkout hook fires inside ephemeral temp-dir clones made by unrelated tools (e.g. CocoaPods), racing rsync and breaking their install


gioxx:
### tokensave version

Installed via Homebrew (`/opt/homebrew/bin/tokensave`), current as of 2026-09.

### What happened

`tokensave install` sets a global `core.hooksPath` and writes a `post-checkout` hook that runs unconditionally on any fresh clone/checkout, anywhere on disk — not just inside real projects:

```sh
# ~/.config/git/hooks/post-checkout
if [ "$1" = "0000000000000000000000000000000000000000" ]; then
	/opt/homebrew/bin/tokensave init >/dev/null 2>&1 &
fi
```

Because `core.hooksPath` is global, this also fires for `git clone` invocations made by completely unrelated tools that use throwaway temp directories, e.g. CocoaPods fetching a pod from a git source. Concretely, `pod install` does:

```
$ /usr/bin/git clone https://github.com/google/boringssl.git /var/folders/.../T/d.../
$ /usr/bin/git -C /var/folders/.../T/d.../ checkout --quiet <sha>
$ /usr/bin/rsync -a --exclude=.git --delete /var/folders/.../T/d.../ ~/Library/Caches/CocoaPods/Pods/Release/BoringSSL-GRPC/...
```

The checkout triggers `post-checkout`, which backgrounds `tokensave init`, writing `.tokensave/tokensave.db-wal` (and `-shm`) into that temp clone directory. `rsync -a --delete` runs immediately after and races the backgrounded `tokensave init` process, which is still creating/checkpointing those files. Result, reproducible on every single `pod install` run that needs to git-clone a pod:

```
rsync: /var/folders/.../T/d.../.tokensave/tokensave.db-wal: open (2): No such file or directory
```

This aborts the pod install (deletes `Podfile.lock` in the process) with no indication tokensave is involved — the error only mentions rsync and a `.tokensave` path that looks like project state, not a stray hook side effect.

Confirmed the mechanism directly, independent of CocoaPods:

```
$ rm -rf /tmp/x && git clone --depth 1 https://github.com/octocat/Hello-World.git /tmp/x
$ ls -la /tmp/x
.git/  .tokensave/  README
```

Any tool that clones into a temp dir and expects to own/clean up that directory exclusively (rsync with `--delete`, mv, rm right after, etc.) is exposed to the same race.

### Relation to other issues

[...]

PR #581, opened 2026-09-21
fix(hooks): replace the legacy unfenced post-checkout block in place (#580)

### LINKED ISSUE #580: install keeps the legacy unfenced post-checkout block and adds a second block
### tokensave version

7.12.1

### Operating system

macOS

### Installation method

cargo install

### Summary

Versions 6.4.3 and 7.3.0 wrote a post-checkout block that has no `# tokensave: end auto-init` line.
`install_or_migrate_block` cannot replace that block. It adds the v2 block and keeps the legacy block.
The hook file then contains two tokensave blocks, and the two blocks run on each checkout.

#### Legacy block

```sh
# tokensave: auto-init
if [ "$1" = "0000000000000000000000000000000000000000" ]; then
      /path/to/tokensave init >/dev/null 2>&1 &
fi
```

## Cause

replace_fenced_block returns None when it does not find the end marker.
install_or_migrate_block then calls write_global_hook, which appends the v2 block (src/agents/hooks.rs, the None arm).
The test a_file_with_no_fence_is_declined records this decision.
The comment says that the code cannot know the extent of the legacy block.

## Effects

1. A fresh clone starts tokensave init and tokensave hook post-checkout at the same time. The two processes write to the same .tokensave/ directory.
2. The legacy block does not call hook_post_checkout. Thus the temp directory skip from #569 does not apply to it. A clone into the system temp directory continues to start tokensave init.
3. The file never gets to one clean v2 block. Each later version bump replaces only the fenced block. The legacy block stays.

## Proposed fix

The extent of the legacy block is known. The block starts at the marker line and stops at the first line that is only fi.
migrate_chain_repo_hook already uses this rule for the chain block.
strip_tokensave_sections already removes the legacy block. Refer to the test strip_handles_the_unfenced_post_checkout_bodies_from_6_4_3_and_7_3_0.

1. If the start marker has no version stamp and the file has no end marker, find the first fi line after the marker.
2. Replace the lines from the marker through that fi line with the v2 block.
3. If the file has no fi line after the marker, append the v2 block as the code does now.

This change keeps all lines outside the block byte-for-byte.

### Steps to reproduce

1. Write the legacy block above into ~/.config/git/hooks/post-checkout.
3. Run tokensave install with the global git hook option.
4. Read the hook file.
Expected result: the file contains one v2 block.
Result from the code path: the file contains the legacy block and the v2 block.

My hook file had the legacy block with tokensave 7.12.1 installed. I replaced the block manually.

### Error output

```shell
  ! git post-checkout hook at /Users/<username>/.config/git/hooks/post-checkout carries an outdated tokensave section — run `tokensave reinstall` to update it (content outside tokensave's markers is preserved)
```

### Component

Indexing / sync

#### Environment

- tokensave 7.12.1 (stable)
- macOS, global install, core.hooksPath = ~/.config/git/hooks

map588:
## Summary

- Install now replaces the unfenced `post-checkout` block from 6.4.3 and 7.3.0 in place. It no longer appends a second block after it.
- A hook file that already holds the legacy block and an appended v2 block now becomes a file with one block. The content between the two blocks stays.
- All content outside the tokensave block stays byte-for-byte.

## Motivation

Fixes #580.

Versions 6.4.3 and 7.3.0 wrote the `post-checkout` block with no `# tokensave: end auto-init` line. `replace_fenced_block` declines that block, and `install_or_migrate_block` appends the v2 block after it. This has two effects.

1. The two blocks run on each checkout. The legacy block calls `tokensave init` directly. Thus the temp directory skip from #569 does not apply to it.
2. On the subsequent install, the fenced search pairs the legacy start marker with the end marker of the appended block. The replacement removes all content between the two blocks. This includes the `chain-repo-hook` section and the hook code of the user.

I reproduced the two effects with the stable 7.12.1 binary and a temporary `HOME`.

## Changes

- `src/agents/hooks.rs`: new function `replace_legacy_checkout_block`. It identifies the two shipped shapes by their exact form: the bare marker line, an `if` on the fresh-checkout sentinel, and one `fi`. It declines a blank line, a second tokensave marker, or a nested `if` before the `fi`. It also declines a block that has the end marker after the `fi`, because `replace_fenced_block` owns that block. When it declines, the caller appends, as before.
- `src/agents/hooks.rs`: `install_or_migrate_block` runs the legacy search before the fenced search. This order is necessary. The fenced search gives a span that is too wide when the file holds the two blocks.
- The legacy replacement also removes one appended fenced block and the blank separator line that `write_global_hook` put before it.
- `stale_hook_blocks` needs no change. It already reports the two file shapes as stale.

## Test plan

- [x] `cargo test` passes

`cargo test --workspace --locked`: exit 0. 176 test binaries, 3896 tests passed, 0 faile

`cargo clippy --workspace --all-targets --locked` shows 8 warnings. All 8 are on lines te. The two warnings in `hooks.rs` are `doc_markdown` on the `CocoaPods` doc comments from #570. `cargo fmt --all -- --check` is clean.

New unit tests in `fence_tests`:

- `a_legacy_unfenced_block_is_replaced_in_place` (the 6.4.3 shape and the 7.3.0 shape)
- `a_legacy_block_with_an_appended_duplicate_collapses_to_one_block`
- `a_fenced_or_unrecognised_block_is_not_treated_as_legacy`
- `a_legacy_install_is_migrated_not_appended_to` (through `install_or_migrate_block`, anp)

Manual test: I ran `tokensave githooks on` with a temporary `HOME`. The hook file was a kout` file from before the fix.

| Hook file before | 7.12.1 stable | This branch |
| --- | --- | --- |
| Legacy block, then chain section | Keeps the legacy block and appends the v2 block | Oposition. The chain section stays. |
| Legacy block, user line, chain section, appended v2 block, user line | One v2 block and the last user line. The first user line and the chain section are gone. | One v2 block. The two user lines and
the chain section stay. |

On this branch, a second `githooks on` run does not change the file. The result passes `

## Checklist

- [x] Breaking changes documented (if any): none


aovestdipaperino:
Merged — thank you, @map588. Careful work: reproducing both effects against the stable 7.12.1 binary with a temporary `HOME`, and recognising the two shipped shapes by their exact form rather than guessing an extent, is exactly the right level of caution for a function that rewrites a file the user also owns. The ordering note (legacy search before the fenced one, because the fenced span is too wide when both blocks are present) was the subtle part and you caught it.

Closes #580.

